PR #28, opened 2022-10-14
Feature/reserve management refactor

jkrivine:
Quite a large PR following discussion on asymmetric handling of `reserve` in `Direct` and `Forwarder` strats. 
* `Direct` offers are now treated as a Forwarder that can have only one user: its admin. 
* As a consequence, default reserve for admin of Direct strat is admin's address
* If one wants to use maker `contract` as reserve one must now set explicitely `contract.setReserve(admin, address(contract))`
* `reserve(address)` is now expecting an argument to avoid having a returned value that is caller dependent
* `setReserve(maker, reserver)` is no longer restricted to contract's admin but may only be called by `maker` (which is `admin()` for `Direct` strats) 

Consequences of the above changes are the following simplications:
* `checkList` can now be entirely done in `MangroveOffer`
* `setReserve(address, address)` and `reserve(address)` are now in `MangroveOffer`

Additional change:
* `_newOffer` for `Forwarder` strat needs to be given a caller in order to set owner of the new offer properly: if called in a hook the previous implementation would assign Mangrove as the owner of the offer! 
* Bug corrected in `Direct` strat where `withdrawToken` was assuming reserve to be `makerContract` when no router is used (this is no longer true) since `makerContract` can do the routing itself.
* critical Bug corrected in `Forwarder` strat where `setReserve` was only checking `msg.sender` to be maker. As a consequence `maker` could freely use *any* approved reserve for its funds.


src/strategies/offer_maker/abstract/Direct.sol
```diff
@@ -77,27 +65,22 @@ abstract contract Direct is MangroveOffer {
...
-  function tokenBalance(IERC20 token) external view override returns (uint) {
-    AbstractRouter router_ = router();
-    return router_ == NO_ROUTER ? token.balanceOf(reserve()) : router_.reserveBalance(token, reserve());
-  }
-
   function flush(IERC20[] memory tokens) internal {
     AbstractRouter router_ = MOS.getStorage().router;
     if (router_ == NO_ROUTER) {
       for (uint i = 0; i < tokens.length; i++) {
         require(
```
peterMangrove, line 84: Why not call `reserve(admin())` outside for loop. Doing it inside for loop, could result in multiple calls

peterMangrove (changes requested):

peterMangrove (approved):

PR #62, opened 2022-11-04
fix(package): Fix imports

lnist:
Due to
https://github.com/foundry-rs/foundry/issues/1486
https://github.com/foundry-rs/foundry/issues/1933
we use a semi-unique path for imports.
Change is simply running the following two through vscode search-replace with regex:
```
import (\{[^}]*\}) from "src\/
import $1 from "mgv_src/
```
and
```
import "src/
import "mgv_src/
```

### DIFF
--- a/script/ActivateMarket.s.sol
+++ b/script/ActivateMarket.s.sol
@@ -2,8 +2,8 @@
 pragma solidity ^0.8.13;
 
 import {Deployer} from "mgv_script/lib/Deployer.sol";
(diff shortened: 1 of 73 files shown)


adhusson (approved):
checked, changes are as described + some newlines eg [here](https://github.com/mangrovedao/mangrove-core/pull/62/files#diff-3aaadb266a8a4749e80fe89195914c28c27418b2856ed9f664394a31d8767961L22)

jkrivine (approved):

PR #63, opened 2022-11-04
Increase modularity

adhusson:
* move general lib files (used by script/test) to `lib/`
* MangroveOrder is now in `strategies` -- as a result, not part of regular mangrove deploys
* improved & added test for the single MangroveOrder deploy


lnist:
> MangroveOrder is now in strategies -- as a result, not part of regular mangrove deploys
but we still want to deploy it with the other stuff, just using its own deploy script, right?

lnist (approved):
This is one step closer to splitting out mangrove-strats, super!

PR #68, opened 2022-11-07
feat: mangrove receives all fees, gov can withdraw

adhusson:
After discussing with Jean we evaluated as follows:
* big gas savings (removes a storage by removing `vault` and a write by removing one token transfer)
* low difficulty
* no security issues... that we can think of

--> worth it


src/MgvRoot.sol
```diff
@@ -134,4 +132,26 @@ contract MgvRoot is HasMgvEvents {
...
     require(_local.active(), "mgv/inactive");
   }
+
+  /* # Token transfer functions */
+  /* `transferTokenFrom` is adapted from [existing code](https://soliditydeveloper.com/safe-erc20) and in particular avoids the
+  "no return value" bug. It never throws and returns true iff the transfer was successful according to `tokenAddress`.
+
+    Note that any spurious exception due to an error in Mangrove code will be falsely blamed on `from`.
+  */
+  function transferTokenFrom(address tokenAddress, address from, address to, uint value) internal returns (bool) {
```
lnist, line 140: Could we use the ones in transferlib instead?

lnist (commented):
I have not been part of the discussion of whether to have a vault or not - if the decision is to remove it, then alright, and the changes lgtm.

src/MgvRoot.sol
adhusson: I think transferlib was derived from mangrove's transfer functions. Unless it's the exact same code I'd feel better keeping those.

PR #70, opened 2022-11-08
Audit/fixes

jkrivine:
Fixing issues raised by security audit (detailed in the commits)


src/strategies/interfaces/IOrderLogic.sol
```diff
@@ -86,6 +86,14 @@ interface IOrderLogic {
   ///@return res The timestamp beyond which `offerId` on the `(outbound_tkn, inbound_tkn)` offer list should renege on trade. 0 means no expiry.
   function expiring(IERC20 outbound_tkn, IERC20 inbound_tkn, uint offerId) external returns (uint);
 
+  ///@notice Updates the time to live for a specific offer.
+  ///@param outbound_tkn The outbound token of the order.
+  ///@param inbound_tkn The inbound token of the order.
+  ///@param offerId The offer id whose expiry date is to be set. 
+  ///@param date in seconds since unix epoch
+  ///@dev If new ttl is in the past of the current block, offer will renege on trade.
```
lnist: Maybe add timestamp: "current block timestamp"

lnist: Replace "ttl" with expiry?

src/strategies/MangroveOrder.sol
```diff
@@ -59,6 +52,14 @@ contract MangroveOrder is Forwarder, IOrderLogic {
     }
   }
 
+  ///@inheritdoc IOrderLogic
+  function setExpiry(IERC20 outbound_tkn, IERC20 inbound_tkn, uint offerId, uint date)
+    public
+    mgvOrOwner(outbound_tkn, inbound_tkn, offerId)
```
lnist, line 59: // We also allow Mangrove to call this so that it can be used from, e.g., post-hooks.

src/strategies/MangroveOrder.sol
```diff
@@ -202,46 +209,61 @@ contract MangroveOrder is Forwarder, IOrderLogic {
...
-      // `fund` is no longer needed so sending it back to `msg.sender`
-      refund = fund;
-    } else {
-      // offer was successfully posted
-      // `fund` was used and we leave `refund` at 0.
+    unchecked {
+      uint residualWants;
+      uint residualGives;
+      if (tko.fillWants) {
+        // partialFill => tko.makerWants < res.takerGot + res.fee
```
lnist: (sorry if you already answered this in another thread): makerWants input by the owner, right? It is not computed. So tko.makerWants could be 0 in which case we get an underflow in this unchecked calculation.

Should we stress that the invariant is _supposed_ to hold since tko.makerWants is _supposed_ to be before slippage? Should we have a "require"?

It is worse for "makerGives" because "residualGives" can become very high with underflow - making a really good offer for someone to grab.

src/strategies/MangroveOrder.sol
```diff
@@ -88,6 +89,9 @@ contract MangroveOrder is Forwarder, IOrderLogic {
 
   ///@inheritdoc IOrderLogic
   function take(TakerOrder calldata tko) external payable returns (TakerOrderResult memory res) {
+    // Checking whether order is expired
+    require(tko.expiryDate == 0 || tko.expiryDate >= block.timestamp, "mgvOrder/expired");
```
lnist: In __lastLook__ we use:
require(exp == 0 || block.timestamp <= exp, "mgvOrder/expired");

so just for readability I think we should use the same comparison.

lnist (changes requested):
Some small comments

src/strategies/MangroveOrder.sol
```diff
@@ -209,64 +209,62 @@ contract MangroveOrder is Forwarder, IOrderLogic {
...
-        refund = fund;
-      } else {
-        // offer was successfully posted
-        // `fund` was used and we leave `refund` at 0.
+  ) internal returns (uint refund) {
+    uint residualWants;
+    uint residualGives;
+    if (tko.fillWants) {
+      uint makerGives = tko.takerGives - tko.slippageAmount;
+      // partialFill => tko.makerWants < res.takerGot + res.fee
```
lnist: remove "tko." from "tko.makerwants" in comment

src/strategies/interfaces/IOrderLogic.sol
```diff
@@ -23,11 +23,10 @@ interface IOrderLogic {
...
   ///@param inbound_tkn the inbound token used to identify the order book
   ///@param fillOrKill true to revert if market order cannot be filled and resting order failed or is not enabled; otherwise, false
-  ///@param takerWants desired total amount of `outbound_tkn`
-  ///@param makerWants taker wants before slippage (`makerWants == takerWants` when `fillWants`)
-  ///@param takerGives available total amount of `inbound_tkn`
-  ///@param makerGives taker gives before slippage (`makerGives == takerGives` when `!fillWants`)
-  ///@param fillWants if true, the market order stops when `takerWants` units of `outbound_tkn` have been obtained; otherwise, the market order stops when `takerGives` units of `inbound_tkn` have been sold.
+  ///@param takerWants desired total amount of `outbound_tkn` (slippage included)
+  ///@param takerGives available total amount of `inbound_tkn` (slippage included)
+  ///@param slippageAmount if `fillWants` (buying) this is the amount of inbound tokens that must be substracted from `takerGives` to obtain the taker price before slippage. Else this is the amount of outbound tokens that must be added to `takerWants`
```
lnist, line 28: Should we give an example of how to calculate slippageAmount given a wanted price?

lnist (changes requested):

lnist (approved):

adhusson (approved):
lgtm

PR #83, opened 2022-11-16
feat: Add collectByImpersonation to MgvCleaner

espendk:
NB: Tests still TBD


lnist (approved):
I'll add a task on adding a test for this

PR #92, opened 2022-11-17
fix(deploy): Deployment script for local tests missed MangroveOrderEnriched

lnist:
Consequential change to b41d0d420bd87487501a35cfb7654cbca3b44c12


adhusson (approved):
ok, but should clean things up later -- both mangrove order and enriched? not sure that's necessary

PR #103, opened 2022-11-22
Feat/simplify reserve

jkrivine:
This PR implements features suggested by the implementation effort of the `pooledForwarder` and takes into account auditors' remarks about the absence of use case for offer owners to update gasreq.

Refactoring is:
- Offer owner's `reserve` is no longer settable by offer owners. This is now an admin move (code simplification)
- `IMakerLogic` is renamed in to `ILiquidityProvider_SDK` to emphasize that its utility is to provide a ABI compatible interface with the SDK
- `IOfferLogic` no longer expects an `updateOffer` function, this is now required by `ILiquidityProvider`
- `MangroveOrder` implements its custom `udpateOffer` which does not take `gasreq` nor `gasprice` as argument. This function is offer owner callable.


src/strategies/MangroveOrder.sol
```diff
@@ -239,7 +252,7 @@ contract MangroveOrder is Forwarder, IOrderLogic {
         inbound_tkn: inbound_tkn,
         wants: residualWants,
         gives: residualGives,
-        gasreq: offerGasreq() + additionalGasreq, // using default gasreq of the strat + potential admin defined increase
+        gasreq: offerGasreq(), // using default gasreq of the strat + potential admin defined increase
```
lnist: update comment

src/strategies/offer_forwarder/abstract/Forwarder.sol
```diff
@@ -55,6 +50,22 @@ abstract contract Forwarder is IForwarder, MangroveOffer {
...
+  ///@notice modifier to enforce function caller to be either admin or offer owner
+  modifier adminOrOwner(IERC20 outbound_tkn, IERC20 inbound_tkn, uint offerId) {
+    if (msg.sender != admin()) {
+      require(ownerData[outbound_tkn][inbound_tkn][offerId].owner == msg.sender, "Forwarder/unauthorized");
+    }
+    _;
+  }
+
+  ///@notice modifier to enforce function caller to be offer owner
+  modifier onlyOwner(IERC20 outbound_tkn, IERC20 inbound_tkn, uint offerId) {
```
lnist: we already have mgvOrOwner which is identical and better named - use that? Or change this to not allow mgv.

src/strategies/MangroveOrder.sol
```diff
@@ -61,9 +57,26 @@ contract MangroveOrder is Forwarder, IOrderLogic {
     expiring[outbound_tkn][inbound_tkn][offerId] = date;
   }
 
-  ///@inheritdoc IOrderLogic
-  function setAdditionalGasreq(uint additionalGasreq_) external onlyAdmin {
-    additionalGasreq = additionalGasreq_;
+  ///@notice updates an offer on Mangrove
+  ///@dev if caller is admin, this is used to update offer gasreq if need be
+  ///@dev otherwise this can be used to update price of the resting order
```
lnist: I don't understand this comment.
How does the admin update gasreq via this? The only possibility I see is that she changes router and then invokes this to let offerGasreq return a new value - but she would have to provide the original values for wants/gives - is that intentional? Should it be separate methods?

src/strategies/interfaces/ILiquidityProvider_SDK.sol
```diff
@@ -0,0 +1,58 @@
...
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+pragma solidity >=0.8.0;
+
+pragma abicoder v2;
+
+import "./IOfferLogic.sol";
+
+// Completes IOfferLogic to provide an ABI for LiquidityProvider class of Mangrove's SDK
+interface ILiquidityProvider_SDK is IOfferLogic {
```
lnist: I don't like that we reference the SDK in the name :) . At least remove the underscore. But I could also go with removing SDK altogether. LiquidityProvider is rather generic though, but it is OK by me.
Other suggestions, some worse than others, pick whichever you like ILiquidityProviderSDK, ILiquidityProvider, IOfferChange, IChangeOffers, IOfferRepository, IOfferUpdater, IOfferCRUD, IOrderBook

Also, retractOffer - should that be part of this or still reside on IOfferLogic? It seems to be in the same family of functions which change the order book.

src/strategies/interfaces/ILiquidityProvider_SDK.sol
```diff
@@ -0,0 +1,58 @@
...
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribution.
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+pragma solidity >=0.8.0;
+
+pragma abicoder v2;
+
+import "./IOfferLogic.sol";
+
+// Completes IOfferLogic to provide an ABI for LiquidityProvider class of Mangrove's SDK
```
lnist: Natspec this top comment

src/strategies/interfaces/ILiquidityProvider_SDK.sol
```diff
@@ -0,0 +1,58 @@
...
+interface ILiquidityProvider_SDK is IOfferLogic {
+  ///@notice creates a new offer on Mangrove.
+  ///@param outbound_tkn the outbound token of the offer list of the offer
+  ///@param inbound_tkn the outbound token of the offer list of the offer
+  ///@param wants the new amount of outbound tokens the offer maker requires for a complete fill
```
lnist: drop the "new" in "new amount"

src/strategies/interfaces/ILiquidityProvider_SDK.sol
```diff
@@ -0,0 +1,58 @@
...
+  ///@param gives the new amount of inbound tokens the offer maker gives for a complete fill
+  ///@param gasreq the new amount of gas units that are required to execute the trade (use type(uint).max for using `this.offerGasreq()`)
```
lnist: Having gasprice/gasreq here means the SDK will always be able to set them, but we will possibly ignore them?

src/strategies/interfaces/IOfferLogic.sol
```diff
@@ -85,7 +78,7 @@ interface IOfferLogic is IMaker {
   ///@dev throws with a reason if something (e.g. an approval) is missing.
   function checkList(IERC20[] calldata tokens) external view;
 
-  ///@notice View of the reserve's balance of a particular asset.
+  ///@notice View of the maker's balance of a particular asset.
```
lnist: the only implementation right now is on MangroveOffer where it does look at the maker's reserve's balance.  So for PooledForwarder this tokenBalance would be the router's balance which would be the entire pool?

src/strategies/offer_forwarder/OfferForwarder.sol
```diff
@@ -14,10 +14,10 @@ pragma solidity ^0.8.10;
 pragma abicoder v2;
 
 import {Forwarder, IMangrove, IERC20} from "mgv_src/strategies/offer_forwarder/abstract/Forwarder.sol";
-import {IMakerLogic} from "mgv_src/strategies/interfaces/IMakerLogic.sol";
+import {ILiquidityProvider_SDK as LP} from "mgv_src/strategies/interfaces/ILiquidityProvider_SDK.sol";
```
lnist: Consider spelling it out instead abbreviating to LP

src/strategies/offer_forwarder/abstract/Forwarder.sol
```diff
@@ -55,6 +50,22 @@ abstract contract Forwarder is IForwarder, MangroveOffer {
     _;
   }
 
+  ///@notice modifier to enforce function caller to be either admin or offer owner
+  modifier adminOrOwner(IERC20 outbound_tkn, IERC20 inbound_tkn, uint offerId) {
```
lnist: This seems not to be used? Was it intended to be used for updateOffer?

src/strategies/interfaces/ILiquidityProvider_SDK.sol
lnist: Consider: Should we have a ForwarderLiquidityProvider or maybe an overload without the gasprice gasreq - possibly with a default implementation somewhere?

src/strategies/MangroveOffer.sol
```diff
@@ -139,18 +139,11 @@ abstract contract MangroveOffer is AccessControlled, IOfferLogic {
   }
 
   /// @inheritdoc IOfferLogic
-  function setReserve(address maker, address reserve_) public override onlyCaller(maker) {
-    require(__checkReserveApproval__(reserve_, maker), "mgvOffer/makerNotApproved");
+  function setReserve(address maker, address reserve_) public override onlyAdmin {
```
lnist: Previously admin could not fiddle with reserves - Is it safe to allow? Could admin misuse this? For instance set reserve to 0xcoffee and make all orders fail, collecting bounty?
Is reserve always either maker OR router, then maybe we could codify that somehow? For instance by inheriting from a building block that enforces desired behavior ?

src/strategies/offer_maker/OfferMaker.sol
```diff
@@ -57,4 +57,36 @@ contract OfferMaker is IMakerLogic, Direct {
       })
     );
   }
+
+  // Updates offer `offerId` on the (`outbound_tkn,inbound_tkn`) Offer List of Mangrove.
+  // NB #1: Offer maker MUST:
+  // * Make sure that offer maker has enough WEI provision on Mangrove to cover for the new offer bounty in case Mangrove gasprice has increased (function is payable so that caller can increase provision prior to updating the offer)
+  // * Make sure that `gasreq` and `gives` yield a sufficient offer density
+  // NB #2: This function will revert when the above points are not met
+  function updateOffer(
```
lnist, line 57: use something like IOfferLogic's natspec for this method instead?

src/strategies/offer_maker/OfferMaker.sol
```diff
@@ -57,4 +57,36 @@ contract OfferMaker is IMakerLogic, Direct {
...
+    IERC20 outbound_tkn,
+    IERC20 inbound_tkn,
+    uint wants,
+    uint gives,
+    uint gasreq, // give `type(uint).max` to use previous value
+    uint gasprice,
+    uint pivotId,
+    uint offerId
+  ) public payable override mgvOrAdmin {
```
lnist: onlyadmin+external?

src/strategies/routers/AavePoolManager.sol
```diff
@@ -0,0 +1,57 @@
...
+contract AavePoolManager is AaveRouter {
+  constructor(address _addressesProvider, uint _referralCode, uint _interestRateMode, uint overhead)
+    AaveRouter(_addressesProvider, _referralCode, _interestRateMode, overhead)
+  {}
+
+  // Admin function to manage position on AAVE
+  function redeem(IERC20 token, address reserve, uint amount, address to) external onlyAdmin {
+    // NB if `reserve` != this, it must approve this router for increasing overlying debt token
+    require(
+      TransferLib.transferTokenFrom(overlying(token), reserve, address(this), amount), "AaveRouter/borrow/transferFail"
```
lnist: use AavePoolManager in messages?

src/toy_strategies/offer_maker/tutorial/OfferMakerTutorial.sol
```diff
@@ -3,13 +3,13 @@ pragma solidity ^0.8.10;
 
 // Import the types we will be using below
 import {Direct} from "mgv_src/strategies/offer_maker/abstract/Direct.sol";
-import {IMakerLogic} from "mgv_src/strategies/interfaces/IMakerLogic.sol";
+import {ILiquidityProvider_SDK as ISDK} from "mgv_src/strategies/interfaces/ILiquidityProvider_SDK.sol";
```
lnist: uuh new name. ISDK.. hm - again consider spelling it out :p

src/toy_strategies/offer_maker/tutorial/OfferMakerTutorial.sol
```diff
@@ -3,13 +3,13 @@ pragma solidity ^0.8.10;
...
 import {IMangrove} from "mgv_src/IMangrove.sol";
 import {IERC20, MgvLib} from "mgv_src/MgvLib.sol";
 
 //----------------
 
-contract OfferMakerTutorial is Direct {
+contract OfferMakerTutorial is Direct, ISDK {
```
lnist: Hm, for the tutorial the intend was to keep it simple so I left out a number of parameters that I did not want to explain.
Wouldn't that be O.K.?

Otherwise we can have this change and then we just have to state in the tutorial that they should not care about these just yet.

Also, new name... ISDK... hm - again consider spelling it out :p

src/toy_strategies/offer_maker/tutorial/OfferMakerTutorialResidual.sol
```diff
@@ -3,15 +3,15 @@ pragma solidity ^0.8.10;
...
+import {ILiquidityProvider_SDK as ISDK} from "mgv_src/strategies/interfaces/ILiquidityProvider_SDK.sol";
 import {IMangrove} from "mgv_src/IMangrove.sol";
 import {IERC20, MgvLib} from "mgv_src/MgvLib.sol";
 
 /* Note this is a copy of OfferMakerTutorial.sol with a changed __posthookSuccess__ */
 
 //----------------
 
-contract OfferMakerTutorialResidual is Direct {
+contract OfferMakerTutorialResidual is Direct, ISDK {
```
lnist: Same as comment for other tutorial

lnist (changes requested):

src/strategies/MangroveOrder.sol
jkrivine: done

src/strategies/MangroveOffer.sol
jkrivine: Yes this gives rugpull power to admin. I am wondering what is the best solution here. Maybe to set reserve at deploy time only?

src/strategies/MangroveOrder.sol
jkrivine: deprecated comment

src/strategies/interfaces/ILiquidityProvider_SDK.sol
jkrivine: I'll go for `ILiquidityProvider` then.
Yes I was wondering also about retract, but newOffer and udpateOffer involve placing an offer in the book, which may have very different implementation depending on the strat. For Instance ghost places offer two by two, mango never places offers after initialization (only updates). Forwarder never updates gasreq or gasprice etc.
where retractOffer is always the same irrespectively of the strat.

src/strategies/interfaces/ILiquidityProvider_SDK.sol
jkrivine: Let's park this for now, but ignoring the parameters sounds not as bas as having two classes

src/strategies/interfaces/IOfferLogic.sol
jkrivine: Good catch. This function should go away its return value is strat implementation dependent

src/strategies/offer_maker/OfferMaker.sol
jkrivine: yes

src/strategies/routers/AavePoolManager.sol
jkrivine: wdym?

src/toy_strategies/offer_maker/tutorial/OfferMakerTutorial.sol
jkrivine: OK but it would be the occasion to mention we add these interface for a reason

adhusson:
I'm not sure I can review that until next week, so maybe we can just go over it together!

src/strategies/offer_forwarder/ForwarderTester.sol
```diff
@@ -0,0 +1,60 @@
...
+  function internal__get__(uint amount, MgvLib.SingleOrder calldata order) external returns (uint) {
+    return __get__(amount, order);
+  }
+
+  function internal__posthookFallback__(MgvLib.SingleOrder calldata order, MgvLib.OrderResult calldata result)
+    external
+    returns (bytes32)
+  {
+    return __posthookFallback__(order, result);
+  }
```
lnist, line 59: these functions seem to be unused?

src/strategies/interfaces/ILiquidityProvider_SDK.sol
lnist: Alright, seems reasonable :)

src/strategies/interfaces/IOfferLogic.sol
lnist: Super - and then we agree we simply remove the implementation in mangrove.js that calls this, right? I assume it is not used anywhere in the UI.

src/strategies/MangroveOffer.sol
lnist: You fixed this by making the __reserve__ hook.

src/strategies/routers/AavePoolManager.sol
lnist: The contract is called AavePoolManager now, but the require-strings say AaveRouter.

src/toy_strategies/offer_maker/tutorial/OfferMakerTutorial.sol
lnist: You decide - both can work IMHO.

lnist (approved):
Some suggestions to consider, but lgtm.

src/strategies/offer_forwarder/ForwarderTester.sol
jkrivine: Yes I think we can use them to do more unit testing. But wip

PR #132, opened 2022-12-05
missing exports

jkrivine:


lnist (changes requested):
I have already added SimmpleTestMaker back to the abi_exports - does it have to be in the full exports?

jkrivine:
> I have already added SimmpleTestMaker back to the abi_exports - does it have to be in the full exports?

Yes only abi export is sufficient indeed

lnist (commented):
I assume OfferForwarder export is not used by downstream, so giving a thumbs up.

lnist (approved):

PR #138, opened 2022-12-06
Removing gasreq and gasprice from public interface

jkrivine:


jkrivine:
The idea of this PR is to require `ILiquidityProvider` to provide simple `updateOffer` and `newOffer` functions, using default values for `gasprice` and `gasreq`. This unifies interface between `Forwarder` and `Direct` strats.
Strat builders can still implement a public offer management that lets offer owners set `gasreq` and `gasprice`

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Remove slippage from MangroveOrder - resting order now posted at same price
 - Introduce **reserve** hook on MangroveOffer which replaces other reserve logic
 - Add collectByImpersonation to MgvCleaner
+- `ILiquidityProvider` provide simple `updateOffer` and `newOffer` public functions, using default values for `gasprice` and `gasreq`. This unifies interface between Forwarder and Direct strats. Strat builders can still implement a public offer management that lets offer owners set gasreq and gasprice
```
lnist, line 9: I just release a new package... so I guess this will be part of the next package. Do you want it earlier? Then I can release another package. This doesn't change deployed contracts, right?

src/strategies/offer_forwarder/OfferForwarder.sol
```diff
@@ -26,23 +26,18 @@ contract OfferForwarder is ILiquidityProvider, Forwarder {
...
+  {
     offerId = _newOffer(
       OfferArgs({
         outbound_tkn: outbound_tkn,
         inbound_tkn: inbound_tkn,
         wants: wants,
         gives: gives,
-        gasreq: gasreq,
+        gasreq: offerGasreq(),
         gasprice: 0,
```
lnist, line 41: You use gasprice:0 to use mangrove's gasprice. You specify gasreq, but could use `type(uint24).max` to use Forwarder's/Direct's default (which is offerGasreq(). You decide :)

lnist (approved):
Minor comment - otherwise lgtm.

PR #141, opened 2022-12-07
fix(Forwarder): Reuse config already available and change approxBounty

lnist:
Calculate approxBounty at the latest possibly point to reduce risk of compiler changing the result too much.

### DIFF
--- a/src/strategies/offer_forwarder/abstract/Forwarder.sol
+++ b/src/strategies/offer_forwarder/abstract/Forwarder.sol
@@ -308,13 +308,13 @@ abstract contract Forwarder is IForwarder, MangroveOffer {
     // NB if several offers of `this` contract have failed during the market order, the balance of this contract on Mangrove will contain cumulated free provision
 
     // computing an under approximation of returned provision because of this offer's failure
-    (MgvStructs.GlobalPacked global, MgvStructs.LocalPacked local) = MGV.config(order.outbound_tkn, order.inbound_tkn);
-    uint provision =
-      10 ** 9 * order.offerDetail.gasprice() * (order.offerDetail.gasreq() + order.offerDetail.offer_gasbase());
+    uint gasreq = order.offerDetail.gasreq();
+    uint provision = 10 ** 9 * order.offerDetail.gasprice() * (gasreq + order.offerDetail.offer_gasbase());
 
     // gasUsed estimate to complete posthook and penalize this offer is ~1750 (empirical estimate)
-    uint approxBounty =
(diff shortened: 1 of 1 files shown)


src/strategies/offer_forwarder/abstract/Forwarder.sol
```diff
@@ -308,13 +308,13 @@ abstract contract Forwarder is IForwarder, MangroveOffer {
...
-      10 ** 9 * order.offerDetail.gasprice() * (order.offerDetail.gasreq() + order.offerDetail.offer_gasbase());
+    uint gasreq = order.offerDetail.gasreq();
+    uint provision = 10 ** 9 * order.offerDetail.gasprice() * (gasreq + order.offerDetail.offer_gasbase());
 
     // gasUsed estimate to complete posthook and penalize this offer is ~1750 (empirical estimate)
-    uint approxBounty =
-      (order.offerDetail.gasreq() - (gasleft() - GAS_APPROX) + local.offer_gasbase()) * global.gasprice() * 10 ** 9;
+    uint gasprice = order.global.gasprice() * 10 ** 9;
+    uint approxGasConsumption = gasreq + GAS_APPROX + order.local.offer_gasbase();
+    uint approxBounty = (approxGasConsumption - gasleft()) * gasprice;
```
jkrivine, line 317: if this reverts for underflow are we good? I think so because everything works as if offer logic had used the whole gasreq (and therefore there is nothing to claim from provisions).

jkrivine (approved):
Yes of course, I don't know why I forgot that Mangrove was also sending its current config in the order result.

src/strategies/offer_forwarder/abstract/Forwarder.sol
espendk: x-posting Lasse's response from Slack:

1. the underflow risk was already there before, right?
2. I think it warrants a comment on why this should not underflow - is gasleft() < gasreq?
3. Could we just change approxBounty to int ? Or will that enable other over/underflow problems in the comparison? I am not sure how (int)approxBounty >= (uint)provision works in solidity :smile:

jkrivine: 1. yes
2. if `approxGasConsumption` > `gasleft` this implies that `GAS_APPROX` is underapproximated. This may happen if this hook is not called last in the hiearchy of inheritance. 
3. an underflow will make the hook revert and the whole released provision is lost to maker.

espendk: @jkrivine Are you ok with merging this or is further discussion needed?

jkrivine: the only issue is that this touches audit

PR #144, opened 2022-12-07
Discussion of principles for events in Mangrove and strats

peterMangrove:
This PR serves two purposes:

1. Propose a solution to the issue of not having sufficient information to index limit orders sent from the web app, such that we can show _Order history_ and _Open orders_ in the web app.
1. Discussion of the principles for event emission in Mangrove and strats. This an area where we need to align and improve our development process going forward, so the current issue seems a good opportunity to have the discussion.

# Event principles #

When emitting events, it seems there are two principled approaches (and likely a middle ground to be found between them):

## Emit minimal events and data ##

One approach is emitting as few events and as little data as possible. This means no redundancy among events and event watchers will likely need to parse most `Mangrove` and strat events in order to reconstruct what happened.

For example, in the current `MangroveOrder` we emit an `OrderSummary` containing a lot of the same information as the `OrderComplete` event emitted by Mangrove in the same tx. If we want to be minimal, `OrderSummary` should only contain data that can not already be found in other events, like `OrderComplete`.

Here's what that would mean for the `OrderSummary` event:

```solidity
  // Current OrderSummary event:
  event OrderSummary(
    IMangrove mangrove,
    IERC20 indexed outbound_tkn,
    IERC20 indexed inbound_tkn,
    address indexed taker,
    bool fillWants,
    uint takerGot,
    uint takerGave,
    uint penalty
  );

  // Minimal OrderSummary event:
  event OrderSummary(
    bool fillWants,
  );
```

Event watchers would then need to also watch Mangrove's `OrderComplete` event and match `OrderSummary` with the appropriate instance of the `OrderComplete` event (nested orders and posthooks may make this non-trivial).

The `OrderComplete` itself also has redundant information: It holds the `takerGot` and `takerGave`, but these can be deduced from the preceding `OfferSuccess` events.

## Emit self-contained events ##

At the other end of the spectrum, events should be entirely self-contained and understandable in themselves, without correlating with other events.

Continuing the example above with `OrderSummary`, we would keep the redundancy wrt `OrderComplete` and even add some missing information, to allow event watcher to just watch this event on `MangroveOrder`:

```solidity
  // Self-contained OrderSummary event:
  event OrderSummary(
    IMangrove mangrove,
    IERC20 indexed outbound_tkn,
    IERC20 indexed inbound_tkn,
    address indexed taker,
    uint takerWants,
    uint takerGives,
    bool fillWants,
    bool fillOrKill,
    bool restingOrder,
    uint expiryDate,
    uint takerGot,
    uint takerGave,
    uint bounty,
    uint fee,
    uint offerId
  );
```

This approach is easier on consumers while consuming more gas. In most case, indexers and other event watchers can read a single event and will not have to correlate it with other event nor deduce info like `restingOrder` by looking at events that were not emitted. (e.g. logic like "if `OfferWrite` was emitted,  then `restingOrder` must have been `true`").

# Current situation in Mangrove and strats #

Most of the events being emitted by Mangrove and strats currently seem to lean toward self-contained events. For example, most events include information that allows event watcher to identify the offer list on a particular Mangrove. Similarly, (and as mentioned above), strats events include information that is also included in Mangrove events.

# This PR #

As mentioned above, the code changes in this PR address the issue at hand: Providing sufficient data in events to allow indexing of the data needed to show _Order history_ and _Open orders_ in the web app.

I have followed the second principled approach of emitting self-contained events since this seems to most closely follow the existing principles in Mangrove and strats. Therefore this PR adds all relevant data to the events we emit when an order is complete, both in Mangrove itself and in `MangroveOrder`.


src/strategies/MangroveOrder.sol
```diff
@@ -46,13 +46,18 @@ contract MangroveOrder is Forwarder, IOrderLogic {
     }
   }
 
+  //  added this event to keep track of what the expiry date is. Can't call the event "setExpriy" because of the method of the same name
+  event expiryWrite(address indexed outbound_tkn, address indexed inbound_tkn, uint offerId, uint date);
```
jkrivine: I'd use `SetExpriry` to be consistent with Mangrove's event naming scheme

src/IMangrove.sol
```diff
@@ -40,16 +39,37 @@ interface IMangrove {
...
+  event OrderComplete( // this could be deduced from the OfferSuccess events emtitted from "execute" in Mangrove
     address indexed outbound_tkn,
     address indexed inbound_tkn,
     address indexed taker,
+    uint takerWants,
+    uint takerGives,
     uint takerGot,
     uint takerGave,
+    bool fillWants,
+    uint penalty,
```
jkrivine: we may want to call this `bounty` or `penaltyReceived`

src/IMangrove.sol
```diff
@@ -40,16 +39,37 @@ interface IMangrove {
...
+    uint feePaid
```
jkrivine, line 59: `feePaid` is derivable from last Mangrove's `SetFee` event.

src/IMangrove.sol
```diff
@@ -40,16 +39,37 @@ interface IMangrove {
...
+  );
+  // added this event, because we used the OrderComplete event before, but this did not makes sense for Snipes, when we now also have takerWants and takerGives in OrderComplete
+  // snipes does not have 1 wants and gives, but an array of them.
+  // this could be deduced from the OfferSuccess events emtitted from "execute" in Mangrove
+  // same
+  event SnipesComplete( // this might be to much to add in the event
```
jkrivine: I am not a fan of having a separate log. What do we gain besides the `targets`? 
We actually know the targets from the `OfferSuccess` or `OfferFail` logs. We could have later a frontend that cooks its own market order in the form of snipes, and you would not want to distinguish between the two logs

src/MgvLib.sol
```diff
@@ -77,26 +77,47 @@ contract HasMgvEvents {
...
+    uint takerGave,
+    bool fillWants,
+    uint penalty,
+    uint feePaid
+  );
+  // added this event, because we used the OrderComplete event before, but this did not makes sense for Snipes, when we now also have takerWants and takerGives in OrderComplete
+  // snipes does not have 1 wants and gives, but an array of them.
+  // this could be deduced from the OfferSuccess events emtitted form "execute" in Mangrove
+  // same
+  event SnipesComplete( // this might be to much to add in the event
```
jkrivine: I'd remove this

src/strategies/MangroveOrder.sol
```diff
@@ -106,7 +111,6 @@ contract MangroveOrder is Forwarder, IOrderLogic {
 
   ///@inheritdoc IOrderLogic
   function take(TakerOrder calldata tko) external payable returns (TakerOrderResult memory res) {
-    // Checking whether order is expired
```
jkrivine: why do you remove this comment?

src/strategies/MangroveOrder.sol
```diff
@@ -204,16 +208,12 @@ contract MangroveOrder is Forwarder, IOrderLogic {
...
-      mangrove: MGV,
-      outbound_tkn: tko.outbound_tkn,
-      inbound_tkn: tko.inbound_tkn,
-      fillWants: tko.fillWants,
-      taker: msg.sender,
-      takerGot: res.takerGot,
-      takerGave: res.takerGave,
-      penalty: res.bounty
-    });
+    emit OrderSummary(
```
jkrivine: emitting a data structure is going to make log parser's life complicated.
To avoid stack to deep just define:
```solidity
function logOrderData(TakerOrder memory tko, TakerOrderResult memory res) internal {
  emit OrderSummary(...);
}
```

jkrivine (changes requested):

src/strategies/MangroveOrder.sol
peterMangrove: Agreed. I had tried with a small s, but that clashed with the method. Hadn't tried with a capital "S". That works

src/IMangrove.sol
peterMangrove: Had already changed it to bounty in MgvLib, missed it in IMangrove.

src/strategies/MangroveOrder.sol
peterMangrove: This was a mistake, have added it again

src/IMangrove.sol
espendk: From the perspective on self-contained events (which this PR is going for) it makes sense to include the actual fee.

Also, one will need to look at all the taken offers to calculate this precisely due to rounding.

src/IMangrove.sol
espendk: Market orders and snipes are different types of orders following different rules, so IMO it makes sense to emit different events for them, so that consumers can distinguish between them.

src/strategies/MangroveOrder.sol
peterMangrove: Okay, makes sense. Changed it

src/MgvLib.sol
```diff
@@ -77,26 +77,47 @@ contract HasMgvEvents {
   event SetGasprice(uint value);
 
   /* Market order execution */
-  event OrderStart();
-  event OrderComplete(
+  event OrderStart(); // OrderStart does not need any info, we emit all relevant data in OrderComplete or SnipesComplete
+    // tried to add all relevant data for an order.
+  // same
```
lnist: I assume you clean up the comments before completing the PR :)

src/MgvLib.sol
```diff
@@ -77,26 +77,47 @@ contract HasMgvEvents {
...
   );
 
   /* * Offer execution */
-  event OfferSuccess(
+  event OfferSuccess( // This should be takerGot and TakerGave, not TakerWants and TakerGives, since this is when the offer has been successfully taken.
     address indexed outbound_tkn,
     address indexed inbound_tkn,
     uint id,
     // `maker` is not logged because it can be retrieved from the state using `(outbound_tkn,inbound_tkn,id)`.
     address taker,
```
lnist: index on taker? or maybe id? - that way you could find order history for the taker or for the maker via the id.

src/MgvOfferTaking.sol
```diff
@@ -114,7 +114,18 @@ abstract contract MgvOfferTaking is MgvHasOffers {
       /* Over the course of the market order, a penalty reserved for `msg.sender` has accumulated in `mor.totalPenalty`. No actual transfers have occured yet -- all the ethers given by the makers as provision are owned by Mangrove. `sendPenalty` finally gives the accumulated penalty to `msg.sender`. */
       sendPenalty(mor.totalPenalty);
 
-      emit OrderComplete(outbound_tkn, inbound_tkn, taker, mor.totalGot, mor.totalGave, mor.totalPenalty, mor.feePaid);
+      emit OrderComplete(
+        outbound_tkn,
```
lnist: I would suggest we use the {} syntax to make it clear what we set to what.

src/strategies/MangroveOrder.sol
```diff
@@ -46,13 +46,18 @@ contract MangroveOrder is Forwarder, IOrderLogic {
     }
   }
 
+  //  added this event to keep track of what the expiry date is.
+  event SetExpiry(address indexed outbound_tkn, address indexed inbound_tkn, uint offerId, uint date);
```
lnist, line 49: hm is it because we need a history? I am a bit torn about emitting events for every single change - we could just read this value from the chain, I would not expect us to use the indexer for every read.

src/strategies/MangroveOrder.sol
```diff
@@ -204,19 +209,30 @@ contract MangroveOrder is Forwarder, IOrderLogic {
...
+      MGV,
+      tko.outbound_tkn,
+      tko.inbound_tkn,
+      msg.sender,
+      tko.fillOrKill,
+      tko.takerWants,
+      tko.takerGives,
+      tko.fillWants,
+      tko.restingOrder,
+      tko.expiryDate,
```
lnist: the expiry date is emitted in expiry events, right? So it seems redundant.

src/strategies/interfaces/IOrderLogic.sol
```diff
@@ -55,22 +55,32 @@ interface IOrderLogic {
 
   ///@notice Information about the order.
   ///@param mangrove The Mangrove contract on which the offer was posted
-  ///@param outbound_tkn The outbound token of the order.
-  ///@param inbound_tkn The inbound token of the order.
+  //@param outbound_tkn The outbound token of the order.
+  //@param inbound_tkn The inbound token of the order.
```
lnist: you need three /// for it to be natspec.

lnist (commented):
I've added some comments

src/strategies/MangroveOrder.sol
lnist: For instance in mangrove.js we have a getOfferInfo method to retrieve some data - I would expect we could add something similar to expiry.

src/strategies/MangroveOrder.sol
espendk: This PR is following the self-contained approach as discussed in the PR description.

src/strategies/MangroveOrder.sol
espendk: This is the same principle that is followed in Mangrove: All interesting state changes are signaled via events.

src/MgvLib.sol
jkrivine: index on taker makes sense. On `id` not much

src/strategies/MangroveOrder.sol
jkrivine: If you run a cleaning bot you want to maintain expiry dates in your cache.

src/IMangrove.sol
jkrivine: Then maybe just adding a field `snipe:bool` to `OrderComplete`?

src/MgvLib.sol
peterMangrove: Yes

espendk:
Conclusion after discussion today:
- It's not clear what principles we should follow and we're therefore not ready to make broader changes at this time.
- For now, we'll focus on addressing the issue at hand. We'll do this by adding the needed information to `MangroveOrder` as suggested in this PR.
  - To reduce the impact on gas (due to logging many fields), we'll make the types of the event as small as possible. Eg for `takerWants` we don't need `uint` (=256 bits) but can use `uint96` instead.
- We'll keep this PR for later reference, but make it a draft.
- The changes needed now will be implemented in another PR. https://github.com/mangrovedao/mangrove-core/pull/148

peterMangrove:
### Missing data found:

- We do not emit the penalty for a specific offer.
- We do not emit fee paid for a specific offer

peterMangrove:
The SetExpiry Event should be defined in the IOrderLogic, this way Proxima will be able to catch that event for all future contracts

espendk:
@peterMangrove This is deprecated and should be closed, right?

I'll close it and then you can reopen it if needed :-)

PR #148, opened 2022-12-08
feat: events sufficient for indexer in MangroveOrder

peterMangrove:


peterMangrove:
Did not use specific uint's like `uni96`, because `TakerOrder` and `TakerOrderResult` did not specify the types. We can make this change later, when we have a large review of our Events.

src/strategies/interfaces/IOrderLogic.sol
```diff
@@ -58,19 +58,34 @@ interface IOrderLogic {
   ///@param outbound_tkn The outbound token of the order.
   ///@param inbound_tkn The inbound token of the order.
   ///@param taker The address of the taker
+  ///@param fillOrKill The fillOrKill that take was call with
```
lnist:
```suggestion
  ///@param fillOrKill The fillOrKill that take was called with
```

lnist: Same for expiry, restingOrder

lnist (changes requested):

src/strategies/interfaces/IOrderLogic.sol
peterMangrove: Done

lnist (approved):

src/strategies/interfaces/IOrderLogic.sol
```diff
@@ -58,19 +58,34 @@ interface IOrderLogic {
...
   ///@param fillWants If true, the market order stopped when `takerWants` units of `outbound_tkn` had been obtained; otherwise, the market order stopped when `takerGives` units of `inbound_tkn` had been sold.
+  ///@param restingOrder The restingOrder boolean take was called with
+  ///@param expiryDate The expiry date take was called with
   ///@param takerGot How much the taker got
   ///@param takerGave How much the taker gave
-  ///@param penalty How much penalty was given
+  ///@param bounty How much bounty was given
+  ///@param fee How much fee was paid for the order
+  ///@param restingOrderId If a restingOrder was posted, then this holds the offerId for the restingOrder
+
```
espendk: Blank line should be removed

```suggestion
```

espendk (changes requested):

espendk (approved):

espendk:
> Did not use specific uint's like `uni96`, because `TakerOrder` and `TakerOrderResult` did not specify the types. We can make this change later, when we have a large review of our Events.

It might even be that it's cheaper gas wise to keep `uint` and not make the types smaller: https://ethereum.stackexchange.com/questions/3067/why-does-uint8-cost-more-gas-than-uint256

I agree with Peter that we should push this to the later discussion and get this PR merged.

jkrivine (approved):

PR #151, opened 2022-12-08
fix: correctly return decimals of MintableERC20BlWithDecimals

adhusson:


lnist (approved):
I wonder why it is a public uint - it can be read via decimals. It is only used for test, so it doesn't matter.

PR #155, opened 2022-12-09
Final modification for audit

jkrivine:
This PR addresses remaining issues raised by the audit:
* [FIXES] 5.1 Forwarder.provisionOf Calculation Is Wrong
  * `MangroveOffer` provides a corrected internal calculation for locked provision
  * `Forwarder` exposes the function and adds  offer's `weiBalance` to the returned value
* [FIXES] 5.2 Users Can Steal Funds From MangroveOrder
  * a difference between Mangrove's `offer_gasbase` of update offer's `gasbase` now also triggers a recomputation of `gasprice`
  * `__posthookSuccess__` of MangroveOffer now uses internal `_updateOffer`'s implementation provided by Forwarder (and Direct). This avoids successfully reposting an offer on someone else's provision.
  * as a consequence `_updateOffer` no longer takes the spurious argument `owner` (but takes extra argument `offerId` as compared to `_newOffer` which now requires an `owner` argument for Forwarder strats.
* [GAS OPTIM] using local and global configuration sent by Mangrove during `makerExecute` instead of calling `MGV.config`
* [FIXES NEW ISSUE] We found an issue w.r.t the computation of residual provision initially defined on 
`__posthookFallback__` in Forwarder strats. The issue is that a strat builder can potentially override this hook the following way: 
```solidity 
function __posthookFallback__(…) … {   
  super.__posthookFallback(…);     
  additional_behavior 
}  
```
and all gas cost incurred by the `additional_behavior` is not taken into account by the computation occurring in `super`. This would force strat developers to always call super at the end of their override which is a rather strong constraint. Instead we decided to enforce this by delaying the computation of the residual provision *after* `__posthookFallback__` in a dedicated hook called `__handleResidualProvision__` which is a noop by default.  Note the computation is unchanged simply done later.
* [MINOR] the emitted log `OrderSummary` in `MangroveOrder` has more data.

### DIFF
--- a/src/strategies/MangroveOffer.sol
+++ b/src/strategies/MangroveOffer.sol
@@ -31,9 +31,8 @@ abstract contract MangroveOffer is AccessControlled, IOfferLogic {
   AbstractRouter public constant NO_ROUTER = AbstractRouter(address(0));
   bytes32 constant OUT_OF_FUNDS = keccak256("mgv/insufficientProvision");
   bytes32 constant BELOW_DENSITY = keccak256("mgv/writeOffer/density/tooLow");
-  bytes32 constant REPOST_FAILED_DUST = "posthook/dustRemainder";
-  bytes32 constant REPOST_FAILED = "posthook/repostFailed";
   bytes32 constant REPOST_SUCCESS = "posthook/reposted";
+  bytes32 constant REPOST_FAILED = "posthook/repostFailed";
 
   ///@notice guards for restricting a function call to either `MGV` or `admin()`.
   ///@dev When `msg.sender` is `MGV`, the function is being called either via `makerExecute` or `makerPosthook`.
@@ -260,13 +259,28 @@ abstract contract MangroveOffer is AccessControlled, IOfferLogic {
     return order.offer.gives() - order.wants;
   }
 
+  ///@notice converts Mangrove's error message string back into its original bytes32 type
+  ///@param reason the revert reason
+  function _repostStatus(string memory reason) internal pure returns (bytes32) {
+    bytes32 reason_hsh = keccak256(bytes(reason));
+    if (reason_hsh == BELOW_DENSITY) {
+      return "mgv/writeOffer/density/tooLow"; // offer not reposted because residual is below density
+    } else {
+      if (reason_hsh == OUT_OF_FUNDS) {
+        return "mgv/insufficientProvision"; // offer not reposted for other reasons (i.e lack of provision)
+      } else {
+        return REPOST_FAILED;
+      }
+    }
+  }
+
   ///@notice Post-hook that implements default behavior when Taker Order's execution succeeded.
   ///@param order is a recall of the taker order that is at the origin of the current trade.
   ///@param maker_data is the returned value of the `__lastLook__` hook, triggered during trade execution. The special value `"lastLook/retract"` should be treated as an instruction not to repost the offer on the book.
   ///@return data can be:
   /// * `"posthook/filled"` when offer was completely filled
   /// * `"posthook/reposted"` when offer was partially filled and successfully reposted
-  /// * `"posthook/dustRemainder"` when offer was partially filled but residual was below density (and thus not reposted)
+  /// * Mangrove's revert reason (cast to a bytes32) when residual is below density or `this` balance on Mangrove is too low (and thus not reposted)
   /// @custom:hook overrides of this hook should be conservative and call `super.__posthookSuccess__(order, maker_data)`
   function __posthookSuccess__(MgvLib.SingleOrder calldata order, bytes32 maker_data)
     internal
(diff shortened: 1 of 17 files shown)


src/strategies/MangroveOrder.sol
```diff
@@ -227,7 +228,8 @@ contract MangroveOrder is Forwarder, IOrderLogic {
       takerGave: res.takerGave,
       bounty: res.bounty,
       fee: res.fee,
-      restingOrderId: res.offerId
+      restingOrderId: res.offerId,
+      taker: msg.sender
```
lnist: why this change?

test/lib/tokens/MintableERC20BLWithDecimals.sol
```diff
@@ -11,9 +11,9 @@ contract MintableERC20BLWithDecimals is
   mapping(address => bool) admins;
   uint public _decimals; // full uint to help forge-std's stdstore
 
-  constructor(address admin, string memory name, string memory symbol, uint8 decimals) ERC20BL(name, symbol) {
+  constructor(address admin, string memory name, string memory symbol, uint8 __decimals) ERC20BL(name, symbol) {
```
lnist: why this change?

lnist (commented):
Are there some spurious changes in this PR or are they intended?

src/strategies/MangroveOrder.sol
jkrivine: by error

test/lib/tokens/MintableERC20BLWithDecimals.sol
jkrivine: to remove a warning

src/strategies/MangroveOrder.sol
lnist: should you revert the change then? :)

src/strategies/offer_forwarder/abstract/Forwarder.sol
```diff
@@ -211,10 +216,15 @@ abstract contract Forwarder is IForwarder, MangroveOffer {
...
-        return offerId;
+        return REPOST_SUCCESS;
       } catch Error(string memory reason) {
         require(args.noRevert, reason);
-        return 0;
+        bytes32 reason_hsh = keccak256(bytes(reason));
+        if (reason_hsh == BELOW_DENSITY) {
+          return REPOST_FAILED_DUST; // offer not reposted because residual is below density
+        } else {
+          return REPOST_FAILED; // offer not reposted for other reasons (i.e lack of provision)
```
lnist: Consider: do we want to return the mangrove reason instead?

lnist: or in other words: why do we translate to repost_failed?

src/strategies/MangroveOffer.sol
```diff
@@ -280,26 +283,34 @@ abstract contract MangroveOffer is AccessControlled, IOfferLogic {
...
     uint new_wants = __residualWants__(order);
-    try MGV.updateOffer(
-      order.outbound_tkn,
-      order.inbound_tkn,
-      new_wants,
-      new_gives,
-      order.offerDetail.gasreq(),
-      order.offerDetail.gasprice(),
-      order.offer.next(), // using next as pivot since this offer is off the book
+    return _updateOffer(
```
lnist: Would it make sense to pass in order.offerDetail and order.global, order.local? This would get rid of the
```
(vars.global, vars.local) = MGV.config(address(args.outbound_tkn), address(args.inbound_tkn));
 vars.offerDetail = MGV.offerDetails(address(args.outbound_tkn), address(args.inbound_tkn), offerId);
```

in the _updateOffer function. This would save some gas for the happy path of reposting residual. For a call to updateOffer->_updateOffer from the outside we can then look these up in updateOffer.

The only drawback I see is that Direct has the same method signature and it does not need all that data.

src/strategies/MangroveOrder.sol
jkrivine: If you know how to revert this specific one yes. Otherwise I don't think this matters

src/strategies/offer_forwarder/abstract/Forwarder.sol
jkrivine: Good point, it requires a bit of work because the reason is a string. So we need to compare its keccak, but yes makes sense.

src/strategies/MangroveOffer.sol
jkrivine: Let's keep it aside for future gas optim work (I think there are many spots where gas can be saved)

src/strategies/MangroveOrder.sol
lnist: Simply move  taker: msg.sender to just below inbound then you have in practice reverted it.

src/strategies/MangroveOffer.sol
```diff
@@ -260,13 +259,28 @@ abstract contract MangroveOffer is AccessControlled, IOfferLogic {
     return order.offer.gives() - order.wants;
   }
 
+  ///@notice converts Mangrove's error message string back into its original bytes32 type
+  ///@param reason the revert reason
+  function _repostStatus(string memory reason) internal pure returns (bytes32) {
```
lnist: If I understand correctly then this method could be reduced to return bytes32(reason) except for the REPOST_FAILED case. Am I missing something?

jkrivine: yes expect you can't cast a string to a bytes32...

lnist (approved):

lnist (approved):

PR #157, opened 2022-12-09
refactor: move IERC20 lib to its own file & remove warnings

adhusson:


src/MgvLib.sol
```diff
@@ -187,25 +188,3 @@ interface IMgvMonitor {
 
   function read(address outbound_tkn, address inbound_tkn) external view returns (uint gasprice, uint density);
 }
-
-interface IERC20 {
```
lnist, line 191: So this is changing audited code, but it is just moving stuff around. Is that ok?

src/IERC20.sol
```diff
@@ -0,0 +1,25 @@
...
+  function transferFrom(address sender, address recipient, uint amount) external returns (bool);
+
+  function symbol() external view returns (string memory);
+
+  event Transfer(address indexed from, address indexed to, uint value);
+  event Approval(address indexed owner, address indexed spender, uint value);
+
+  function decimals() external view returns (uint8);
+
+  function name() external view returns (string memory);
```
lnist, line 24: Its just moving stuff around, and adding this method? :)

lnist (approved):

src/MgvLib.sol
adhusson: Damn I misremembered that MgvLib was not audited. Yes I guess :/

src/IERC20.sol
adhusson: Yes I noticed that it was missing, even though it is part of the [ERC20](https://eips.ethereum.org/EIPS/eip-20) standard

PR #161, opened 2022-12-12
fix: misc Deployer small improvements

adhusson:


lib/Script2.sol
```diff
@@ -13,7 +13,12 @@ contract Script2 is Script {
     Shared global refs for multiple contracts. Better than `vm.etch(hash(name),address(new Contract()).code)`, which does *not* carry over state modification caused by the constructor.
   */
 
-  ToyENS _singletons;
+  ToyENS _singletons = ToyENS(hashToAddress("Mangrove:Singletons"));
+
+  // Computes address from last 20 bytes of hash
+  function hashToAddress(string memory str) internal pure returns (address) {
```
lnist, line 19: Consider naming it stringToAddress?

lnist (approved):

lib/Script2.sol
adhusson: after chat, will leave as is :)

PR #165, opened 2022-12-14
refactor: cleaner fork.get construction

adhusson:


test/lib/forks/Generic.sol
```diff
@@ -44,13 +44,24 @@ contract GenericFork is Script {
...
     if (addr == address(0)) {
-      addr = deployed.get(name);
+      return payable(deployed._addrs(name));
+    }
+  }
+
+  function get(string memory name) public view returns (address payable addr) {
+    addr = getNoRevert(name);
+    if (addr == address(0)) {
+      revert("Fork::get(string name): no contract found for name argument");
```
lnist: Consider adding a reason - could not be found in either context or deployed addresses files?

adhusson: updated

lnist (approved):

PR #166, opened 2022-12-14
feat: allow create2-based deploys

adhusson:


lnist (commented):
1) what about mangroveOrder?
2) we use 0x0 salt as default, right? - is the salt captured as part of the resulting tx so we can find it again for future deployments?

adhusson:
Doing it gradually, will add MangroveOrder later.

The salt doesn't matter much, but yes it is part of the calldata sent to the contract-creation contract.

lnist (approved):

PR #167, opened 2022-12-16
feat: add market tracker

adhusson:
Add permisionless "open markets tracking" to MgvReader.

### DIFF
--- a/src/periphery/MgvReader.sol
+++ b/src/periphery/MgvReader.sol
@@ -49,17 +49,18 @@ contract MgvReader {
     bool accumulate;
   }
 
-  /* Open markets tracking */
-  // Information about a market: tokens, and open state.
-  // array of ordered [tknA,tknB] pairs lists the open markets
+  /**
+   * @notice Open markets tracking (below) provides information about which markets on Mangrove are open. Anyone can update a market status by calling `updateMarket`. 
+   * @notice The array of pairs `_openMarkets` is the array of all currently open markets (up to a delay in calling `updateMarkets`). A market is a pair of tokens `[tkn0,tkn1]`. The orientation is non-meaningful but canonical (see `order`).
+   * @notice In this contract, 'markets' are defined by non-oriented pairs. Usually markets come with a base/quote orientation. Please keep that in mind.
+   * @notice A market {tkn0,tkn1} is open if either the tkn0/tkn1 offer list is active or the tkn1/tkn0 offer list is active.
+   */
   address[2][] internal _openMarkets;
-  //  market array position (= index+1); 0 means not in array
-  mapping(address => mapping(address => uint)) internal marketPositions;
-  // Information about a market:
-  // * tkn0, tkn1 (no particular meaning to the orientation but canonical)
-  // * config01 is the local config for the outbound=tkn0, inbound=tkn1 offer list
-  // * config10 is the local config for the outbound=tkn1, inbound=tkn0 offer list
 
+  /// @notice Markets can be added or removed from `_openMarkets` array. To remove a market, we must remember its position in the array. The `marketPositions` mapping does that.
+  mapping(address => mapping(address => uint)) internal marketPositions;
+
+  /// @notice Config of a market. Assumes a context where `tkn0` and `tkn1` are defined. `config01` is the local config of the `tkn0/tkn1` offer list. `config10` is the config of the `tkn1/tkn0` offer list.
   struct MarketConfig {
     MgvStructs.LocalUnpacked config01;
     MgvStructs.LocalUnpacked config10;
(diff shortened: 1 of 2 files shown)


src/periphery/MgvReader.sol
```diff
@@ -316,4 +322,74 @@ contract MgvReader {
       /* end if success branch **/
     }
   }
+
+  /* Open markets tracking */
+
+  // MgvMarketTracker lets anyone query for a market state (open or not), and
```
espendk: `MgvMarketTracker`? Looks like a remnant from an earlier version?

src/periphery/MgvReader.sol
```diff
@@ -316,4 +322,74 @@ contract MgvReader {
...
+  // gets the list of open markets.
+
+  // Anyone can permisionlessly refresh the state of a market with updateMarket(tkn0,tkn1)
```
espendk: neat that updates are permissionless!

src/periphery/MgvReader.sol
```diff
@@ -316,4 +322,74 @@ contract MgvReader {
...
+    return uint160(tkn0) < uint160(tkn1) ? (tkn0, tkn1) : (tkn1, tkn0);
+  }
+
+  function isMarketOpen(address tkn0, address tkn1) external view returns (bool) {
+    (tkn0, tkn1) = order(tkn0, tkn1);
+    return marketPositions[tkn0][tkn1] > 0;
+  }
+
+  // Will consider a market open iff both the offer lists tkn0/tkn1 and
+  // tkn1/tkn0 are open on Mangrove.
```
espendk: Maybe a market should be considered open if one of the offer lists is open? 

Once a reader is aware of a market, they can read the config of that market and figure out if it's only partially open.

With the current solution, partially open markets will not be discoverable.

espendk (changes requested):
neat that it's permissionless!

I propose one change, see the comments.

src/periphery/MgvReader.sol
```diff
@@ -316,4 +322,74 @@ contract MgvReader {
...
+    }
+    return openMarketsSlice;
+  }
+
+  // canonicalize ordering
+  function order(address tkn0, address tkn1) public pure returns (address, address) {
+    return uint160(tkn0) < uint160(tkn1) ? (tkn0, tkn1) : (tkn1, tkn0);
+  }
+
+  function isMarketOpen(address tkn0, address tkn1) external view returns (bool) {
```
lnist, line 423: Consider using natspec so that we can easily show it in our docs.mangrove.exchange.

src/periphery/MgvReader.sol
adhusson: yes!
```suggestion
  // Lets anyone query for a market state (open or not), and
```

src/periphery/MgvReader.sol
adhusson: Do you think there may ever be partially open markets? I'm imagining that mostly being some kind of error state. So with the current logic, users don't have to brainsweat too much, given the list they know what they can do.

espendk:
we're planning on opening an asymmetric market for a partner, so yes this
will happen 🙂

adhusson:
Pushed changes:
* A market is now open iff one of the corresponding offer lists are active
* Functions  of the form `openMarket(...)` returns each market's configs (in almost all cases where you deal with markets you will need their config anyway).
* natspecced comments

src/periphery/MgvReader.sol
```diff
@@ -48,6 +49,23 @@ contract MgvReader {
     bool accumulate;
   }
 
+  /**
+   * @notice Open markets tracking (below) provides information about which markets on Mangrove are open. Anyone can update a market status by calling `updateMarket`. 
+   * @notice The array of pairs `_openMarkets` is the array of all currently open markets (up to a delay in calling `updateMarkets`). A market is a pair of tokens `[tkn0,tkn1]`. The orientation is non-meaningful but canonical (see `order`).
+   * @notice In this contract, 'markets' are defined by non-oriented pairs. Usually markets come with a base/quote orientation. Please keep that in mind.
```
espendk, line 54: I wonder how a UI will determine which token should be considered the base/quote respectively?

src/periphery/MgvReader.sol
```diff
@@ -20,6 +20,7 @@ pragma solidity ^0.8.10;
 
 import {MgvLib, MgvStructs} from "mgv_src/MgvLib.sol";
 import {IMangrove} from "mgv_src/IMangrove.sol";
+import "forge-std/console.sol";
```
espendk: This should be remove, no?

```suggestion
```

espendk (changes requested):
lgtm, except for a left-over import of forge console.

We'll have to figure out to determine the base/quote ordering in the UI, but I'm not sure there's a good way to do this on-chain. Maybe governance could specify a (partial) ordering of tokens where t_1 < t_2 implies that t_1 is the base of the (t_1, t_2) market?

src/periphery/MgvReader.sol
adhusson: as noted by espen in another conversation, we are already planning to do partially open markets so the recent pushes address that

src/periphery/MgvReader.sol
adhusson: It should be hardcoded by each UI I think. There are traditions, such as the more cash-like token being the quote. But it's really a matter of convention.

espendk (approved):

PR #168, opened 2022-12-16
chore: new addresses

adhusson:


addresses/deployed/matic.json
```diff
@@ -1,18 +1,18 @@
...
     "name": "Mangrove"
   },
   {
-    "address": "0xb43B23e87e045e196e839880F8C2806fCccE4221",
+    "address": "0x09f8b6E32DB426aa477ad815544d0573FeFdBeEF",
     "name": "MgvOracle"
   },
   {
-    "address": "0x106b805d76Ea65F8aF0Bb4E9499a2CB69162C75A",
+    "address": "0xe71f73C7C0cfb02Cfa3F9810659CD23cE88FFB28",
```
lnist, line 11: 1) Deploy with the new reader?
2) Also deploy mangroveorderenriched
3) Can you post the multisig address and how to get from the multisig to the deployed addresses? This way the approver can verify the addressees are correct.

adhusson: We will deploy later, and we may not make a release with those addresses, but it's good to share those on git (they are deployed).

lnist (approved):

PR #171, opened 2022-12-19
Deploy MgvReader to Mumbai

lnist:


espendk (approved):
lgtm - checked the address on Mumbai and that the contract is verified with the open markets feature 👍

PR #173, opened 2022-12-19
fix: correct mgvgov address

adhusson:


espendk (approved):
lgtm - this is the ADDMA MS address

PR #177, opened 2022-12-20
Strat/kandel

jkrivine:
This PR implements Kandel strat featuring:
 * Simple Geometric Kandel strat
 * Possible aave liquidity sourcing via a pooled aave router


src/strategies/offer_maker/market_making/Kandel/Kandel.sol
```diff
@@ -0,0 +1,220 @@
...
+  IERC20 immutable BASE;
+  ///@notice quote of the market Kandel is making
+  IERC20 immutable QUOTE;
+  ///@notice quote distribution: `quoteOfIndex[i]` is the amount of quote tokens Kandel must give or want at index i
+  uint[] quoteOfIndex;
+  ///@notice quote distribution: `baseOfIndex[i]` is the amount of base tokens Kandel must give or want at index i
+  uint[] baseOfIndex;
+  ///@notice `pendingBase/Quote` that failed to be published and must be recycled when possible
+  uint pendingBase;
+  uint pendingQuote;
```
lnist: Consider having a struct with erc20, gives, wants, pending and then have two fields - one for base and one for quote with a method to get the struct for each. Then consider fetching the struct early and passing to function calls.
The idea would be to make the structure clearer and avoid some function calls here and there.

src/strategies/offer_maker/market_making/Kandel/Kandel.sol
```diff
@@ -0,0 +1,220 @@
...
+  event AllBids(IMangrove indexed mgv, IERC20 indexed base, IERC20 indexed quote);
+
+  ///@notice a bid or an ask
+  enum OrderType {
+    Bid,
+    Ask
+  }
+
+  ///@notice maps index to offer id on Mangrove
+  mapping(OrderType => mapping(uint => uint)) offerIdOfIndex;
```
lnist: Would two fields like for the other data be more efficient than a map on ordertype?

src/strategies/offer_maker/market_making/Kandel/Kandel.sol
```diff
@@ -0,0 +1,220 @@
...
+  ///@dev args.wants/gives must match the distribution at index
+  function _populateIndex(OrderType ba, uint index, OfferArgs memory args) internal returns (bytes32) {
+    uint offerId = offerIdOfIndex[ba][index];
+    if (offerId == 0) {
+      offerId = _newOffer(args);
+      if (offerId == 0) {
+        return "newOffer/Failed";
+      } else {
+        offerIdOfIndex[ba][index] = offerId;
+        return REPOST_SUCCESS;
```
lnist: This may not be a repost? It could be a new offer.

src/strategies/offer_maker/market_making/Kandel/Kandel.sol
```diff
@@ -0,0 +1,220 @@
...
+      }
+    } else {
+      return _updateOffer(args, offerId);
+    }
+  }
+
+  function populate(uint from, uint to, uint lastBidIndex, uint gasprice) external payable onlyAdmin {
```
lnist: natspec - fund should be enough for all offers - i.e. gasprice*offerGasreq()*(to-from)?

src/strategies/offer_maker/market_making/Kandel/Kandel.sol
```diff
@@ -0,0 +1,220 @@
...
+  ///@param order a recap of the taker order (order.offer is the executed offer)
+  ///@return dualBa the type of order implementing the transport
+  ///@return dualIndex the distribution index where liquidity is transported
+  ///@return args the argument for `populateIndex` specifying volume and price
+  function __transportLogic__(OrderType ba, MgvLib.SingleOrder calldata order)
+    internal
+    virtual
+    returns (OrderType dualBa, uint dualIndex, OfferArgs memory args)
+  {
+    uint index = offerIdOfIndex[ba][order.offerId];
```
lnist: Seems to use the offerIdOfIndex the wrong way round? Maybe we need an indexForOfferId?

src/strategies/offer_maker/market_making/Kandel/Kandel.sol
```diff
@@ -0,0 +1,220 @@
...
+      emit AllBids(MGV, BASE, QUOTE);
+    }
+    (MgvStructs.OfferPacked dualOffer, MgvStructs.OfferDetailPacked dualOfferDetails) = getOffer(dualBa, dualIndex);
+
+    (dualIndex, dualBa) = ba == OrderType.Ask ? (index - 1, OrderType.Bid) : (index + 1, OrderType.Ask);
+
+    // can repost what the current taker order gives + what the dual offer already gives.
+    uint maxDualGives = dualBa == OrderType.Ask
+      ? dualOffer.gives() + order.gives + pendingBase
+      : dualOffer.gives() + order.gives + pendingQuote;
```
lnist: Make a getPending method instead of this if?

src/strategies/offer_maker/market_making/Kandel/Kandel.sol
```diff
@@ -0,0 +1,220 @@
...
+
+    // what the distribution says the dual order should ask/bid
+    uint shouldGive = _givesOfIndex(dualBa, dualIndex);
+    uint shouldWant = _wantsOfIndex(dualBa, dualIndex);
+
+    args.outbound_tkn = IERC20(order.inbound_tkn);
+    args.inbound_tkn = IERC20(order.outbound_tkn);
+    args.gives = shouldGive > maxDualGives ? maxDualGives : shouldGive;
+    // if giving less volume than distribution, one must adapts wants to match distribution price
+    args.wants = args.gives == shouldGive ? shouldWant : (maxDualGives * shouldWant) / shouldGive;
```
lnist: Remember to reduce pendingBase / pendingQuote.

src/strategies/offer_maker/market_making/Kandel/Kandel.sol
```diff
@@ -0,0 +1,220 @@
...
+    args.gasreq = dualOfferDetails.gasreq();
+    args.gasprice = dualOfferDetails.gasprice();
+  }
+
+  ///@notice takes care of reposting residual offer in case of a partial fill and logging potential issues.
+  ///@param order a recap of the taker order
+  ///@param ba whether the executer offer (order.offer) is a bid or an ask
+  function _handleResidual(MgvLib.SingleOrder calldata order, bytes32 makerData, OrderType ba) internal {
+    bytes32 repostStatus = super.__posthookSuccess__(order, makerData);
+    // Offer failed to repost for bad reason, logging the incident
```
lnist: Outdated comment.

src/strategies/offer_maker/market_making/Kandel/Kandel.sol
```diff
@@ -0,0 +1,265 @@
...
+
+  ///@notice publishes (by either creating or updating) a bid/ask at a given price index
+  ///@param ba whether the offer is a bid or an ask
+  ///@param index the index of the distribution
+  ///@param args the argument of the offer.
+  ///@dev args.wants/gives must match the distribution at index
+  function _populateIndex(OrderType ba, uint index, OfferArgs memory args) internal returns (bytes32) {
+    uint offerId = offerIdOfIndex[ba][index];
+    if (offerId == 0) {
+      offerId = _newOffer(args);
```
lnist: whatabout pivot?

src/strategies/offer_maker/market_making/Kandel/Kandel.sol
```diff
@@ -0,0 +1,265 @@
...
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+pragma solidity ^0.8.10;
+
+import {Direct, IMangrove, AbstractRouter, IERC20} from "mgv_src/strategies/offer_maker/abstract/Direct.sol";
+import {MgvLib, MgvStructs} from "mgv_src/MgvLib.sol";
+
+contract Kandel is Direct {
+  ///@notice number of offers managed by this strat
+  uint16 immutable NSLOTS;
```
lnist: Consider NUMBEROFOFFERS over NSLOTS

src/strategies/offer_maker/market_making/Kandel/Kandel.sol
```diff
@@ -0,0 +1,265 @@
...
+  ///@notice base of the market Kandel is making
+  IERC20 immutable BASE;
+  ///@notice quote of the market Kandel is making
+  IERC20 immutable QUOTE;
+  ///@notice quote distribution: `baseOfIndex[i]` is the amount of base tokens Kandel must give or want at index i
+  uint[] baseOfIndex;
```
lnist: consider: IMHO At works better han Of for most of these names, e.g., baseAtIndex

src/strategies/offer_maker/market_making/Kandel/Kandel.sol
```diff
@@ -0,0 +1,265 @@
...
+
+  ///@notice turns an order type into an (outbound, inbound) pair identifying an offer list
+  ///@param ba whether one wishes to access the offer lists where asks or bids are posted
+  function _tokenPairOfOrderType(OrderType ba) internal view returns (IERC20, IERC20) {
+    return ba == OrderType.Bid ? (QUOTE, BASE) : (BASE, QUOTE);
+  }
+
+  ///@notice returns the Kandel order type of the offer list whose outbound token is given in argument
+  ///@param outbound_tkn the outbound token of the offer list
+  function _orderTypeOfOutbound(IERC20 outbound_tkn) internal view returns (OrderType) {
```
lnist: Are you intending to override these methods in specializations?
From our natspec of MangroveOffer:
```
/// `_f() internal`: descendant of this contract should provide a public wrapper for this function, with necessary guards.
```

Normally we do not prefix with _ do we?

src/strategies/offer_maker/abstract/Direct.sol
```diff
@@ -49,13 +49,13 @@ abstract contract Direct is MangroveOffer {
...
-  function push(IERC20 token, uint amount) internal {
+  function push(IERC20 token, uint amount) internal returns (uint) {
     AbstractRouter router_ = router();
     if (router_ == NO_ROUTER) {
-      return; // nothing to do
+      return amount; // nothing to do
     } else {
       // noop if reserve == address(this)
-      require(router_.push(token, reserve(admin()), amount) == amount, "Direct/pushFailed");
+      return router_.push(token, reserve(admin()), amount);
```
lnist: If you want to remove this require (do we want to remove it? It seems you are only changing the failure message) - then you should add a require to AdvancedAaveRetail. I could not find other contracts using the push.

src/strategies/offer_maker/market_making/Kandel/ExplicitKandel.sol
```diff
@@ -0,0 +1,153 @@
...
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+pragma solidity ^0.8.10;
+
+import {CoreKandel, IMangrove, IERC20, AbstractKandel, MgvLib, MgvStructs} from "./abstract/CoreKandel.sol";
+import "mgv_src/strategies/utils/TransferLib.sol";
+import {console} from "forge-std/console.sol";
```
lnist: It's ok to have console imports during development, but it will break at least doc building, so remember to remove them before completing PR.

script/strategies/Kandel/KdlAnyDist.sol
```diff
@@ -0,0 +1,37 @@
...
+  IMangrove
+} from "mgv_src/strategies/offer_maker/market_making/kandel/ExplicitKandel.sol";
+import {MgvReader} from "mgv_src/periphery/MgvReader.sol";
+import {Deployer} from "mgv_script/lib/Deployer.sol";
+
+/**
+ * @notice Populate Kandel's distribution on Mangrove
+ */
+
+contract KdlAnyDist is Deployer {
```
lnist: I prefer non-lazy naming - so Kandel over Kdl. No need for abbreviation here :) Same goes for file name. Also, I would call this KandelSetDistribution

script/strategies/Kandel/KdlDeploy.s.sol
```diff
@@ -0,0 +1,63 @@
...
+    current = new Kandel(
+      mgv,
+      IERC20(base),
+      IERC20(quote),
+      gasreq,
+      uint16(nslots)
+    );
+    string memory kandelName = getName(IERC20(base), IERC20(quote));
+    fork.set(kandelName, address(current));
+    outputDeployment();
```
lnist: outputDeployment will write the Kandel address to our address.json files. Do we want that? It is not to be shared, right? If it was to be shared, then "KdlAnyDist" should read its address from fork.get("Kandel").

script/strategies/Kandel/KdlAnyDist.sol
```diff
@@ -0,0 +1,37 @@
...
+    });
+  }
+
+  function innerRun(Kandel kdl, uint[] memory baseDist, uint[] memory quoteDist, uint from, uint to) public {
+    require(baseDist.length == quoteDist.length, "Distribution must have same length");
+    require(
+      from < to && to < kdl.NSLOTS() && baseDist.length == (to - from), "interval must be of the form [from,...,to["
+    );
+    prettyLog("Setting distribution on Kandel...");
+    vm.broadcast();
```
lnist: use broadcast(); instead of vm.broadcast(); everywhere.

script/strategies/Kandel/KdlDeploy.s.sol
```diff
@@ -0,0 +1,63 @@
...
+      quote: envAddressOrName("QUOTE"),
+      nslots: vm.envUint("NSLOTS"),
+      gasreq: 160_000
+    });
+  }
+
+  /**
+   * @param base Address of the base currency of the market Mango will act upon
+   * @param quote Addres of the quote of Mango's market
+   * @param nslots the number of price slots of the Mango strat
```
lnist:
```suggestion
   * @param base Address of the base token of the market Kandel will act upon
   * @param quote Address of the quote token of the market Kandel will act upon
   * @param nslots the number of price slots of the Kandel strat
```

lnist: Also, add gasreq description

script/strategies/Kandel/KdlDeploy.s.sol
```diff
@@ -0,0 +1,63 @@
...
+      uint16(nslots)
+    );
+    string memory kandelName = getName(IERC20(base), IERC20(quote));
+    fork.set(kandelName, address(current));
+    outputDeployment();
+    smokeTest();
+  }
+
+  function getName(IERC20 base, IERC20 quote) public view returns (string memory) {
+    try vm.envString("NAME") returns (string memory mangoName) {
```
lnist: search for mango :)

script/strategies/Kandel/KdlGeometricDist.sol
```diff
@@ -0,0 +1,65 @@
...
+    public
+  {
+    require(from < to && to < kdl.NSLOTS(), "interval must be of the form [from,...,to[");
+    require(uint96(baseFrom) == baseFrom, "BASE0 is too high");
+    require(uint96(quoteFrom) == quoteFrom, "QUOTE0 is too high");
+
+    prettyLog("Generating distributions...");
+    uint baseDecimals = kdl.BASE().decimals();
+    fillGeometricDist(baseFrom, /*0.1 ether*/ baseRatio, /*110*/ from, to, baseDist);
+    fillGeometricDist(quoteFrom, /*cash(usdc, 800)*/ quoteRatio, from, to, quoteDist);
```
lnist:
```suggestion
    fillGeometricDist(baseFrom, baseRatio, from, to, baseDist);
    fillGeometricDist(quoteFrom, quoteRatio, from, to, quoteDist);
```

script/strategies/Kandel/KdlGeometricDist.sol
```diff
@@ -0,0 +1,65 @@
...
+    //turning price distribution into quote volumes
+    for (uint i = from; i < to; i++) {
+      quoteDist[i] = (quoteDist[i] * baseDist[i]) / (10 ** baseDecimals);
```
lnist: I would like a comment on the parameters telling that quote is a ... price?

script/strategies/Kandel/KdlPopulates.s.sol
```diff
@@ -0,0 +1,101 @@
...
+import {
+  ExplicitKandel as Kandel,
+  IERC20,
+  IMangrove
+} from "mgv_src/strategies/offer_maker/market_making/kandel/ExplicitKandel.sol";
+import {MgvReader} from "mgv_src/periphery/MgvReader.sol";
+import {Deployer} from "mgv_script/lib/Deployer.sol";
+
+/**
+ * @notice Populate Kandel's distribution on Mangrove
```
lnist:
```suggestion
 * @notice Populate Kandel's offers on Mangrove
```

script/strategies/Kandel/KdlDeploy.s.sol
```diff
@@ -0,0 +1,63 @@
...
+import {
+  ExplicitKandel as Kandel,
+  IERC20,
+  IMangrove
+} from "mgv_src/strategies/offer_maker/market_making/kandel/ExplicitKandel.sol";
+import {Deployer} from "mgv_script/lib/Deployer.sol";
+import {MangroveTest, Test} from "mgv_test/lib/MangroveTest.sol";
+
+/**
+ * @notice deploys a Kandel instance on a given market
```
lnist: Do we expect users to utilize these deploy scripts, or will it be through the SDK that they deploy an instance? (It still makes sense to have the foundry scripts)

script/strategies/Kandel/KdlPopulates.s.sol
```diff
@@ -0,0 +1,101 @@
...
+      lastBidIndex: vm.envUint("LASTBID"),
+      gasprice: vm.envUint("GASPRICE")
+    });
+  }
+
+  function innerRun(
+    uint from, // start index for the first element of the distribution
+    uint to,
+    uint lastBidIndex,
+    uint gasprice
```
lnist: this gasprice includes a scaling to avoid underprovision?

script/strategies/Kandel/KdlPopulates.s.sol
```diff
@@ -0,0 +1,101 @@
...
+        lastBidIndex: lastBidIndex,
+        provBid: provBid,
+        provAsk: provAsk
+      })
+    );
+    vm.stopPrank();
+
+    prettyLog("Populating Mangrove...");
+    vm.broadcast();
+    kdl.populate{value: (provAsk + provBid) * (to - from)}(from, to, lastBidIndex, gasprice, pivotIds);
```
lnist: So we are provisioning both the bid and the ask even though only one is active. Should it be max(provAsk,provBid) instead?

script/strategies/Kandel/KdlPopulates.s.sol
```diff
@@ -0,0 +1,101 @@
...
+
+  struct HeapArgs {
+    uint96[] baseDist;
+    uint96[] quoteDist;
+    uint lastBidIndex;
+    uint provBid;
+    uint provAsk;
+  }
+
+  function evaluatePivots(HeapArgs memory args) internal returns (uint[] memory pivotIds) {
```
lnist: Hm either we need to make a lot of calls from the SDK, use multicall, or call this function which either means calling it in foundry fork, or deploying it if we use the SDK/UI to populate.

src/strategies/offer_maker/market_making/Kandel/abstract/AbstractKandel.sol
```diff
@@ -0,0 +1,58 @@
...
+
+  ///@notice a bid or an ask
+  enum OrderType {
+    Bid,
+    Ask
+  }
+
+  ///@notice base distribution at given index
+  ///@param index of the distribution of base tokens
+  ///@return amount of base tokens that Kandel should give/want at `index`
```
lnist:
```suggestion
  ///@return amount of base tokens that Kandel should give(ask)/want(bid) at `index`
```

src/strategies/offer_maker/market_making/Kandel/abstract/AbstractKandel.sol
```diff
@@ -0,0 +1,58 @@
...
+  function baseOfIndex(uint index) public view virtual returns (uint96 amount);
+
+  ///@notice quote distribution at given index
+  ///@param index of the distribution of quote tokens
+  ///@return amount of quote tokens that Kandel should give/want at `index`
```
lnist:
```suggestion
  ///@return amount of quote tokens that Kandel should give(bid)/want(ask) at `index`
```

src/strategies/offer_maker/market_making/Kandel/abstract/AbstractKandel.sol
```diff
@@ -0,0 +1,58 @@
...
+  function quoteOfIndex(uint index) public view virtual returns (uint96 amount);
+
+  ///@notice transport logic followed by Kandel
+  ///@param ba whether the offer that was executed is a bid or an ask
+  ///@param order a recap of the taker order (order.offer is the executed offer)
+  ///@return dualBa the type of order implementing the transport
+  ///@return dualIndex the distribution index where liquidity is transported
+  ///@return args the argument for `populateIndex` specifying volume and price
```
lnist:
```suggestion
  ///@return args the argument for `populateIndex` specifying gives and wants
```

src/strategies/offer_maker/market_making/Kandel/abstract/AbstractKandel.sol
```diff
@@ -0,0 +1,58 @@
...
+  ///@param index of the distribution of base tokens
+  ///@return amount of base tokens that Kandel should give/want at `index`
+  function baseOfIndex(uint index) public view virtual returns (uint96 amount);
+
+  ///@notice quote distribution at given index
+  ///@param index of the distribution of quote tokens
+  ///@return amount of quote tokens that Kandel should give/want at `index`
+  function quoteOfIndex(uint index) public view virtual returns (uint96 amount);
+
+  ///@notice transport logic followed by Kandel
```
lnist: is "transport logic" a well-known concept or our own term? If the latter, then I think a bit longer explanation would be in order.

src/strategies/offer_maker/market_making/Kandel/abstract/CoreKandel.sol
```diff
@@ -0,0 +1,277 @@
...
+abstract contract CoreKandel is Direct, AbstractKandel {
+  ///@notice number of offers managed by this strat
+  uint16 public immutable NSLOTS;
+  ///@notice base of the market Kandel is making
+  IERC20 public immutable BASE;
+  ///@notice quote of the market Kandel is making
+  IERC20 public immutable QUOTE;
+  ///@notice `pendingBase` is the amount of free (not promised) base tokens in reserve
+  uint128 public pendingBase;
+  ///@notice `pendingQuote` is the amount of free quote tokens in reserve
```
lnist:
```suggestion
  ///@notice `pendingQuote` is the amount of free (not promised) quote tokens in reserve
```

src/strategies/offer_maker/market_making/Kandel/abstract/CoreKandel.sol
```diff
@@ -0,0 +1,277 @@
...
+      pendingQuote = uint128(amount);
+    }
+  }
+
+  function offerIdOfIndex(OrderType ba, uint index) public view returns (uint) {
+    return _offerIdOfIndex[uint(ba)][index];
+  }
+
+  function indexOfOfferId(OrderType ba, uint offerId) public view returns (uint) {
+    return _indexOfOfferId[ba][offerId];
```
lnist: I would still suggest two maps instead :)

src/strategies/offer_maker/market_making/Kandel/abstract/CoreKandel.sol
```diff
@@ -0,0 +1,277 @@
...
+    return OrderType((uint(ba) + 1) % 2);
+  }
+
+  ///@notice returns next index if order is an ask, and previous price index if order is a bid
+  ///@param ba whether the order is an ask or a bid
+  ///@param index the price index of the order
+  ///@return dualIndex the index of the dual order
+  ///@dev `ba == Ask => index > 0` and `ba == Bid => index < NSLOT-1`
+  function dual(OrderType ba, uint index) public pure returns (uint dualIndex) {
+    return ba == OrderType.Ask ? index - 1 : index + 1;
```
lnist: hm, this is only the dual if we are near lastbid? otherwise I would call this method "next", "farther", "further" or "successor".

src/strategies/offer_maker/market_making/Kandel/abstract/CoreKandel.sol
```diff
@@ -0,0 +1,277 @@
...
+  ///@inheritdoc MangroveOffer
+  function __posthookSuccess__(MgvLib.SingleOrder calldata order, bytes32 makerData)
+    internal
+    override
+    returns (bytes32)
+  {
+    OrderType ba = _orderTypeOfOutbound(IERC20(order.outbound_tkn));
+    // adds any unpublished liquidity to pending[Base/Quote]
+    setPending(ba, _handleResidual(order, makerData));
+    // preparing arguments for the dual maker order
```
lnist:
```suggestion
    // preparing arguments for the dual maker offer
```

src/strategies/offer_maker/market_making/Kandel/abstract/CoreKandel.sol
```diff
@@ -0,0 +1,277 @@
...
+  }
+
+  ///@notice publishes (by either creating or updating) a bid/ask at a given price index
+  ///@param ba whether the offer is a bid or an ask
+  ///@param index the index of the distribution
+  ///@param args the argument of the offer.
+  ///@dev args.wants/gives must match the distribution at index
+  function _populateIndex(OrderType ba, uint index, OfferArgs memory args) internal returns (bytes32) {
+    uint offerId = offerIdOfIndex(ba, index);
+    if (isLive(dual(ba), index)) {
```
lnist: when does this happen?

src/strategies/offer_maker/market_making/Kandel/abstract/CoreKandel.sol
```diff
@@ -0,0 +1,277 @@
...
+      _populateIndex(ba, index, args);
+    }
+  }
+
+  ///@notice retracts and deprovisions offers of the distribution interval `[from, to[`
+  ///@param from the start index
+  ///@param to the end index
+  ///@dev this simply provisions this contract's balance on Mangrove.
+  ///@dev use in conjunction of `withdrawFromMangrove` if the user wishes to redeem the available WEIs
+  function retractOffers(uint from, uint to) external onlyAdmin {
```
lnist: Will this revert for invalid ids? i.e. 0? I would expect so.

src/strategies/offer_maker/market_making/Kandel/abstract/CoreKandel.sol
```diff
@@ -0,0 +1,277 @@
...
+  ///@notice repost residual offer and dual offer according to transport logic
+  ///@inheritdoc MangroveOffer
+  function __posthookSuccess__(MgvLib.SingleOrder calldata order, bytes32 makerData)
+    internal
+    override
+    returns (bytes32)
+  {
+    OrderType ba = _orderTypeOfOutbound(IERC20(order.outbound_tkn));
+    // adds any unpublished liquidity to pending[Base/Quote]
+    setPending(ba, _handleResidual(order, makerData));
```
lnist: setPending should either increase/decrease or we should add the current pending to the residual? Currently the existing pending is overwritten.

src/strategies/offer_maker/market_making/Kandel/abstract/CoreKandel.sol
```diff
@@ -0,0 +1,277 @@
...
+
+  ///@notice In case an offer failed to deliver, promised liquidity becomes pending, but offer is not reposted.
+  ///@inheritdoc MangroveOffer
+  function __posthookFallback__(MgvLib.SingleOrder calldata order, MgvLib.OrderResult calldata)
+    internal
+    override
+    returns (bytes32)
+  {
+    OrderType ba = _orderTypeOfOutbound(IERC20(order.outbound_tkn));
+    setPending(ba, order.offer.gives());
```
lnist: I wonder what is best to do here - because why did it fail? if we failed to deliver we will most likely also fail in the future? But yes this solution is probably as good as any.

src/strategies/offer_maker/market_making/Kandel/ExplicitKandel.sol
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  ///@notice sets the base/quote distribution that Kandel should follow using continuous slices
+  ///@param from start index (included). Must be less than `to`.
+  ///@param to end index (excluded). Must be less than `NSLOT`.
+  ///@param slice slice[i][0/1] is the distribution of base/quote at index i.
+  function setDistribution(uint from, uint to, uint[][2] calldata slice) external onlyAdmin {
+    for (uint i = from; i < to; i++) {
+      uint sliceIndex = i - from;
+      require(uint96(slice[0][sliceIndex]) == slice[0][sliceIndex], "Kandel/baseOverflow");
```
lnist: extract to values to local variables

src/strategies/offer_maker/market_making/Kandel/ExplicitKandel.sol
```diff
@@ -0,0 +1,153 @@
...
+    return _baseOfIndex;
+  }
+
+  function quoteDist() external view onlyAdmin returns (uint96[] memory) {
+    return _quoteOfIndex;
+  }
+
+  ///@notice checks whether offer whose logic is being executed is currently the best on Mangove
+  ///@param order the taker order that is being executed
+  ///@dev `isBest` => `order.offer` is the best bid/ask of this strat (but the converse is not true in general).
```
lnist:
```suggestion
  ///@notice checks whether offer whose logic is being executed is currently the best on Mangrove (may not be true during a snipe)
  ///@param order the taker order that is being executed
  ///@dev `isBest` => `order.offer` is the best bid/ask of this strat (but the converse is not true in general).
```

src/strategies/offer_maker/market_making/Kandel/ExplicitKandel.sol
```diff
@@ -0,0 +1,153 @@
...
+
+    // letting dual offer complements taker's liquidity with pending when current offer is not sniped.
+    // Adding pending to compute dual offer's volume would allow taker to drain liquidity
+    // by sniping an offer far from the mid price for a low quantity, dual offer would then only be posted for a
+    // proportionally low volume if not complemented with pending.
+    uint pending = _isBest(order) ? getPending(dualBa) : 0;
+
+    if (shouldGive >= maxDualGives + pending) {
+      if (dualBa == OrderType.Ask) {
+        pendingBase -= uint128(pending);
```
lnist: why the uint128 here? and why not use setPending(dualBa, 0)?

src/strategies/offer_maker/market_making/Kandel/ExplicitKandel.sol
```diff
@@ -0,0 +1,153 @@
...
+    if (index == 0) {
+      emit AllAsks(MGV, BASE, QUOTE);
+    }
+    if (index == NSLOTS - 1) {
+      emit AllBids(MGV, BASE, QUOTE);
+    }
+    dualIndex = dual(ba, index);
+    dualBa = dual(ba);
+
+    (MgvStructs.OfferPacked dualOffer, MgvStructs.OfferDetailPacked dualOfferDetails) = getOffer(dualBa, dualIndex);
```
lnist: will dualOffer always exist at this point?

src/strategies/offer_maker/market_making/Kandel/ExplicitKandel.sol
```diff
@@ -0,0 +1,153 @@
...
+    args.wants = args.gives == shouldGive ? shouldWant : (maxDualGives * shouldWant) / shouldGive;
+    args.fund = 0;
+    args.noRevert = true;
+    args.gasreq = dualOfferDetails.gasreq();
+    args.gasprice = dualOfferDetails.gasprice();
+    args.pivotId = dualOffer.gives() > 0 ? offerIdOfIndex(dualBa, dualIndex) : dualOffer.next();
+  }
+
+  function depositFunds(OrderType ba, uint amount) external override {
+    require(push({token: ba == OrderType.Ask ? BASE : QUOTE, amount: amount}) == amount, "Kandel/depositFailed");
```
lnist: why the custom error?

src/strategies/offer_maker/market_making/Kandel/ExplicitKandel.sol
```diff
@@ -0,0 +1,153 @@
...
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+pragma solidity ^0.8.10;
+
+import {CoreKandel, IMangrove, IERC20, AbstractKandel, MgvLib, MgvStructs} from "./abstract/CoreKandel.sol";
+import "mgv_src/strategies/utils/TransferLib.sol";
+import {console} from "forge-std/console.sol";
+
+contract ExplicitKandel is CoreKandel {
```
lnist: ExplicitDistributionKandel? FixedKandel? FixedDistributionKandel? ConstantDistributionKandel? maybe natspec would make it clearer, explicit can refer to many things :)

src/strategies/offer_maker/market_making/Kandel/ExplicitKandel.sol
lnist: I think I now understand that no, it does not always exist - and this is intentional in that all things are just 0 which also works. It means we'll use 0 for pivotid, but what else can we do.
It seems we do a lot of work even though we will (often?) end up with a crossed book on partial fills - i.e. should we check whether dual offer is alive before doing this work?
The code `dualIndex = dual(ba, index)` is that encoding the transportLogic of moving the bid one point? Then I would expect that method to be overridable but maybe that is not necessary. Should dual be renamed dualIndex?
I think the logic around _transportLogic and reposting is the core of the complexity, and maybe it would help if we could go through it in a call.

script/strategies/Kandel/KdlAnyDist.sol
jkrivine: Ok the status of these scripts needs to be discussed anyway. We probably want a ts API

script/strategies/Kandel/KdlDeploy.s.sol
jkrivine: Yes these foundry scripts can be used as a spec for writing the SDK and for demo purposes

script/strategies/Kandel/KdlGeometricDist.sol
jkrivine: yes so `quoteDist` is initially a (unit) price distribution. We convert this distribution into a wants/gives so that `quoteDist` can be used for wants (in asks) and gives (in bids).

script/strategies/Kandel/KdlPopulates.s.sol
jkrivine: In this script it's a free parameter. So at this point it's up to the deployer to take into account potential scalling

script/strategies/Kandel/KdlPopulates.s.sol
jkrivine: Good point. The idea is to have the bid/ask already covered when it will be posted by the offer logic.

jkrivine: * in case of partial fill the bid is not off the book so you need to cover for the dual ask
* even if the bid is completely filled you would need to tell the offer logic to deprovision it in order to provision for the ask which would be extremely gas costly

script/strategies/Kandel/KdlPopulates.s.sol
jkrivine: We have a function in the current SDK that evaluates pivot based on the cache. The above is a poor man's solution to have that accessible in a foundry script

src/strategies/offer_maker/abstract/Direct.sol
jkrivine: Right. I think the idea of pull/push is always to return transfered amount so that caller may decide to either revert or try another method. The above is just making this uniform. BTW we have not reviewed `Direct` we need to do this.

src/strategies/offer_maker/market_making/Kandel/ExplicitKandel.sol
jkrivine: yes. I actually verified the Kandel deploy... with console.sol in it.

src/strategies/offer_maker/market_making/Kandel/ExplicitKandel.sol
jkrivine: Yeah all these names suck. Explicit mean here that you don't have a function that performs a computation in memory to return the price and volume at slot i. This would be the case if you were implementing a geometric progression for instance, something like:
`price(i,ratio) = price(0)*ratio**i` (with all the overflow issues)

src/strategies/offer_maker/market_making/Kandel/ExplicitKandel.sol
jkrivine: didn't get this

src/strategies/offer_maker/market_making/Kandel/ExplicitKandel.sol
jkrivine: yes let's do this!

src/strategies/offer_maker/market_making/Kandel/ExplicitKandel.sol
jkrivine: pendingBase is `uint128` so I need to cast pending which is a `uint256`.
I don't think I can use `setPending` here because `pending` might be 0.

src/strategies/offer_maker/market_making/Kandel/abstract/AbstractKandel.sol
jkrivine: Yes code rushed in order to be able to demo a first concept. But proper commenting is on todo list

src/strategies/offer_maker/market_making/Kandel/abstract/CoreKandel.sol
jkrivine: makes sense

src/strategies/offer_maker/market_making/Kandel/abstract/CoreKandel.sol
jkrivine: I think dual is fine, but you're right is should maybe be parametric in the jump `k` which might not be always 1

src/strategies/offer_maker/market_making/Kandel/abstract/CoreKandel.sol
jkrivine: you can manually populate offers.

src/strategies/offer_maker/market_making/Kandel/abstract/CoreKandel.sol
jkrivine: you mean if from and to don't make sense?

src/strategies/offer_maker/market_making/Kandel/abstract/CoreKandel.sol
jkrivine: good catch!

src/strategies/offer_maker/market_making/Kandel/abstract/CoreKandel.sol
jkrivine: An natural invariant is that the sum of all offer gives + pending is equal to the balance of the contract's reserve (in base and quote). If this is no the case at some point, we will now how much is missing to restore the strat.

src/strategies/offer_maker/market_making/Kandel/Kandel.sol
jkrivine: Would you be up for refactoring this?

src/strategies/offer_maker/market_making/Kandel/Kandel.sol
jkrivine: I think it makes little difference. But maybe better for uniform programming

src/strategies/offer_maker/market_making/Kandel/Kandel.sol
jkrivine: Yes, I think common practice is to use `_` prefix to denote an internal function. Maybe we should update natspec?

src/strategies/offer_forwarder/abstract/Forwarder.sol
```diff
@@ -237,10 +237,16 @@ abstract contract Forwarder is IForwarder, MangroveOffer {
...
+  ///@param outbound_tkn the outbound token of the offer list.
+  ///@param inbound_tkn the inbound token of the offer list.
+  ///@param offerId the identifier of the offer in the (`outbound_tkn`,`inbound_tkn`) offer list
+  ///@param deprovision if set to `true` if offer owner wishes to redeem the offer's provision.
+  ///@return freeWei the amount of native tokens (in WEI) that have been retrieved by retracting the offer.
+  ///@dev An offer that is retracted without `deprovision` is retracted from the offer list, but still has its provisions locked by Mangrove.
+  ///@dev Calling this function, with the `deprovision` flag, on an offer that is already retracted must be used to retrieve the locked provisions.
+  function _retractOffer(IERC20 outbound_tkn, IERC20 inbound_tkn, uint offerId, bool deprovision)
+    internal
     mgvOrOwner(outbound_tkn, inbound_tkn, offerId)
```
jkrivine: I'd remove the `mgvOrOwner` from this internal function and adds it to public wrappers (including MangroveOrder). This is what we did for `_updateOffer`. This is because you might want a public wrapper that is `onlyOwner` for instance.

src/strategies/offer_maker/OfferMaker.sol
```diff
@@ -70,4 +70,12 @@ contract OfferMaker is ILiquidityProvider, Direct {
       offerId
     );
   }
+
+  ///@inheritdoc ILiquidityProvider
+  function retractOffer(IERC20 outbound_tkn, IERC20 inbound_tkn, uint offerId, bool deprovision)
+    external
+    returns (uint freeWei)
+  {
+    return _retractOffer(outbound_tkn, inbound_tkn, offerId, deprovision);
```
jkrivine: add `onlyAdmin`

src/strategies/MangroveOrder.sol
```diff
@@ -85,6 +85,21 @@ contract MangroveOrder is Forwarder, IOrderLogic {
...
+  ///@param outbound_tkn the outbound token of the offer list.
+  ///@param inbound_tkn the inbound token of the offer list.
+  ///@param offerId the identifier of the offer in the (`outbound_tkn`,`inbound_tkn`) offer list
+  ///@param deprovision if set to `true` if offer owner wishes to redeem the offer's provision.
+  ///@return freeWei the amount of native tokens (in WEI) that have been retrieved by retracting the offer.
+  ///@dev An offer that is retracted without `deprovision` is retracted from the offer list, but still has its provisions locked by Mangrove.
+  ///@dev Calling this function, with the `deprovision` flag, on an offer that is already retracted must be used to retrieve the locked provisions.
+  function retractOffer(IERC20 outbound_tkn, IERC20 inbound_tkn, uint offerId, bool deprovision)
+    public
+    returns (uint freeWei)
```
jkrivine, line 99: add `onlyOwner` (see comment below)

src/strategies/offer_forwarder/OfferForwarder.sol
```diff
@@ -73,6 +73,14 @@ contract OfferForwarder is ILiquidityProvider, Forwarder {
     _updateOffer(args, offerId);
   }
 
+  ///@inheritdoc ILiquidityProvider
+  function retractOffer(IERC20 outbound_tkn, IERC20 inbound_tkn, uint offerId, bool deprovision)
+    external
+    returns (uint freeWei)
```
jkrivine, line 80: add `onlyOwner` (see comment below)

src/strategies/offer_maker/abstract/Direct.sol
```diff
@@ -116,22 +116,27 @@ abstract contract Direct is MangroveOffer {
...
+  function _retractOffer(
     IERC20 outbound_tkn,
     IERC20 inbound_tkn,
     uint offerId,
     bool deprovision // if set to `true`, `this` contract will receive the remaining provision (in WEI) associated to `offerId`.
-  ) public override mgvOrAdmin returns (uint free_wei) {
-    free_wei = MGV.retractOffer(address(outbound_tkn), address(inbound_tkn), offerId, deprovision);
-    if (free_wei > 0) {
-      require(MGV.withdraw(free_wei), "Direct/withdrawFail");
+  ) internal mgvOrAdmin returns (uint freeWei) {
```
jkrivine: remove `mgvOrAdmin`

src/toy_strategies/offer_forwarder/AmplifierForwarder.sol
```diff
@@ -190,6 +190,13 @@ contract AmplifierForwarder is Forwarder {
     }
   }
 
+  function retractOffer(IERC20 outbound_tkn, IERC20 inbound_tkn, uint offerId, bool deprovision)
+    public
+    returns (uint freeWei)
```
jkrivine, line 197: add `onlyOwner`

src/toy_strategies/offer_maker/Amplifier.sol
```diff
@@ -173,6 +173,13 @@ contract Amplifier is Direct {
     }
   }
 
+  function retractOffer(IERC20 outbound_tkn, IERC20 inbound_tkn, uint offerId, bool deprovision)
+    public
+    returns (uint freeWei)
```
jkrivine, line 179: add `onlyAdmin`

src/toy_strategies/offer_maker/tutorial/OfferMakerTutorial.sol
```diff
@@ -70,6 +70,14 @@ contract OfferMakerTutorial is Direct, ILiquidityProvider {
     );
   }
 
+  ///@inheritdoc ILiquidityProvider
+  function retractOffer(IERC20 outbound_tkn, IERC20 inbound_tkn, uint offerId, bool deprovision)
+    external
+    returns (uint freeWei)
```
jkrivine, line 78: add `onlyAdmin`

src/toy_strategies/offer_maker/tutorial/OfferMakerTutorialResidual.sol
```diff
@@ -70,6 +70,14 @@ contract OfferMakerTutorialResidual is Direct, ILiquidityProvider {
     );
   }
 
+  ///@inheritdoc ILiquidityProvider
+  function retractOffer(IERC20 outbound_tkn, IERC20 inbound_tkn, uint offerId, bool deprovision)
+    external
+    returns (uint freeWei)
```
jkrivine, line 78: add `onlyAdmin`

jkrivine (commented):
I cannot require changes since github believes this is my own PR...

src/strategies/offer_forwarder/abstract/Forwarder.sol
lnist: Thanks you are completely right - I have made the suggested changes

src/strategies/offer_maker/market_making/Kandel/Kandel.sol
jkrivine: It's functional style naming scheme f of i is f(i)

PR #180, opened 2022-12-24
test: replace toEthUnit with toUnit

adhusson:
properly display a uint in a token's display units


adhusson:
added a `dp` decimal places arg; if a nonzero number is truncated, ~`(...)`~ `…` (unicode ellipsis character) is appended to the returned string for clarity. Not as well thought-out as the rest, more likely to have an error

adhusson:
added comments :)

jkrivine (approved):

PR #185, opened 2023-01-13
Transmit makerdata

adhusson:
provide `makerPosthook` with `makerData` whether the call to `makerExecute` has failed or not.


jkrivine (approved):

src/MgvOfferTaking.sol
```diff
@@ -453,7 +453,7 @@ abstract contract MgvOfferTaking is MgvHasOffers {
       /* `success` is true: trade is complete */
       if (success) {
         /* In case of success, `retdata` encodes the gas used by the offer. */
```
lnist: extend this comment?

adhusson: done!

PR #195, opened 2023-01-19
Feat: usual dao

peterMangrove:


src/usual/PLUsMgvStrat.sol
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  function setFee(uint16 fee) public onlyAdmin {
+    require(fee <= maxFee, "PLUsMgvStrat/maxFee");
+    _fee = fee;
+    emit SetFee(fee);
+  }
+
+  function setUsualDapp(address usualDapp) public onlyAdmin {
+    _usualDapp = usualDapp;
```
jkrivine, line 76: you probably want to log `SetUsusalDapp` here as well.

src/usual/PLUsMgvStrat.sol
```diff
@@ -0,0 +1,124 @@
...
+    _usUSD.transfer(to, fee);
+    emit FeeWithdrawn(fee);
+  }
+
+  function __put__(uint amount, MgvLib.SingleOrder calldata order) internal virtual override returns (uint) {
+    uint fee = (amount * _fee) / 10_000;
+    _usUSD.transfer(offerIdToOwner[order.offerId], amount - fee);
+    emit FeePaid(fee);
+    return 0;
+  }
```
jkrivine, line 112: I'd use internal functions `creditFee`, `debitFee` that log `CreditFee` and `DebitFee` and perform the required transfers, so that `withdrawFees` is simply an admin restricted wrapper of `debitFee` and `__put__` calls `creditFee`.

src/usual/PLUsMgvStrat.sol
```diff
@@ -0,0 +1,124 @@
...
+
+  function __get__(uint amount, MgvLib.SingleOrder calldata order) internal virtual override returns (uint missing) {
+    _pLUsDAOToken.transferFrom(offerIdToOwner[order.offerId], address(this), amount);
+    return 0;
+  }
```
jkrivine, line 118: there's the possibility to save one (hot) storage read here (so maybe not crucial), if you perform the transfer of the fee to the owner and the transferFrom owner -> this in the LastLook hook. You can then factor the `offerIdToOwner` read.

src/usual/PLUsMgvStrat.sol
```diff
@@ -0,0 +1,124 @@
...
+        gasreq: type(uint).max, // uses the old gasreg
+        gasprice: 0, // igonred
+        pivotId: pivotId,
+        noRevert: true,
+        fund: msg.value
+      }),
+      offerId
+    );
+  }
+}
```
jkrivine, line 188: don't you need to override `__activate__` and `__checkList__` ? (i.e do `_pLUsDAOToken`
and `_metaPLUsDAOToken` etc. have the necessary approvals)?

src/toy/ERC20.sol
```diff
@@ -216,7 +216,7 @@ contract ERC20Base is Context, IERC20 {
    * - `to` cannot be the zero address.
    * - `from` must have a balance of at least `amount`.
    */
-  function _transfer(address from, address to, uint amount) internal virtual {
+  function _transfer(address from, address to, uint amount) internal virtual returns (bool) {
```
jkrivine: if you change the interface you inherit from, this might not work with standard ERC20 contracts. Do openZep contracts return bool here?

src/usual/MetaPLUsDAOToken.sol
```diff
@@ -0,0 +1,76 @@
...
+    require(
+      (
+        (owner == address(_pLUsMgvStrat) && recipient == _mangrove)
+          || owner == _mangrove && recipient == address(_pLUsTakerProxy)
+      ) && msg.sender == _mangrove,
+      "MetaPLUsDAOToken/nonMangroveTransfer"
+    );
+
+    if (owner == _mangrove) {
+      address currentTaker = _pLUsTakerProxy.currentTaker();
```
jkrivine, line 96: This requires advanced scrutiny because the `marketOrder` call in `_pLUsTakerProxy` is possibly reentrant due to the bounty transfer.

src/usual/PLUsMgvStrat.sol
peterMangrove: Good point.

src/toy/ERC20.sol
peterMangrove: You are right, it does not look like the internal function for transfer returns a bool. https://docs.openzeppelin.com/contracts/2.x/api/token/erc20#ERC20-_transfer-address-address-uint256-

src/usual/PLUsMgvStrat.sol
peterMangrove: Done

src/usual/PLUsMgvStrat.sol
peterMangrove: No i don't believe a have to here. Because Mangrove is only trying to transfer Meta token from the strat. It is then the Meta token contract itself that tries to transfer the PLUsDAOToken. And the approve function on the Meta token also approves the the PLUsDAOToken. This means when the Meta Token is trying to transfer from the strat to the taker, then the strat already has approve this by calling activate.

src/usual/MetaPLUsDAOToken.sol
peterMangrove: Interesting point. But i do not see an issue in the taker calling `marketOrder` on `_pLUsTakerProxy` again, when receiving the bounty. And it is not possible to do it in a posthook, since the Meta token only allows the transfer from our strat or mgv, therefore it would never call the posthook.

src/usual/PLUsMgvStrat.sol
peterMangrove: Oh, and i do an approve in the constructor of PLUsMgvStrat, that makes i possible.

PR #196, opened 2023-01-24
scripting & JS improvements

adhusson:
* Export SimpleTestMaker bytecode
* Open markets in MangoveJsDeploy script
* TestMaker can revert individual offers
* Misc script improvements

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # next version
 
+- Export SimpleTestMaker bytecode
+- Open markets in MangoveJsDeploy script
+- TestMaker can revert individual offers
+
 # 1.4.0 (December 2022)
 
 - Added Polygon mainnet deployment addresses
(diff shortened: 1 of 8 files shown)


lnist (commented):
The reason to keep the global shouldrevert around is for compatibility reasons?
The shouldabort is removed - are we not using it, then ok.
ShouldFail is only global - I assume that is intended since it sets allowance . We could, though, set allowance to amount if not shouldfail, then it could still be per offer? You decide whether the feature is relevant.

adhusson:
* yes, `shouldRevert`-global is kept for compatibility, can be removed later
* `shouldAbort` was a misnomer, it just returned some data instead of no data (I can guess the historical reason why)
* You're right that `shouldFail` can be replaced with setting allowance to 0. Also right now the allowance remains 0 after the call! And you can't reliably reset the allowance in the posthook since the posthook may be asked to fail. I'm removing it.

adhusson:
Nevermind `shouldFail` is used deep in the guts of `MakerDeployer`, I'll leave that to another day.

peterMangrove (approved):

PR #202, opened 2023-01-26
Create github-projects-sync.yml

espendk:
Add missing GitHub Projects synchronization workflow.


dontrolle:
CI was failing in tests - which shouldn't be affected at all - just reran CI. Let's see what happens.

dontrolle (approved):
Looks good now. So the prev failing CI was probably a fluke.

PR #226, opened 2023-02-16
Split KandelSeeder to reduce size

lnist:
Introduced an AaveKandelSeeder and removed duplicate AUDIT_TRAIL file.


test/strategies/unit/AavePooledRouter.t.sol
```diff
@@ -5,8 +5,11 @@ import "./OfferLogic.t.sol";
 import {AavePooledRouter} from "mgv_src/strategies/routers/integrations/AavePooledRouter.sol";
 import {PinnedPolygonFork} from "mgv_test/lib/forks/Polygon.sol";
 import {AllMethodIdentifiersTest} from "mgv_test/lib/AllMethodIdentifiersTest.sol";
+import {PoolAddressProviderMock} from "mgv_script/toy/AaveMock.sol";
 
 contract AavePooledRouterTest is OfferLogicTest {
+  bool useForkAave = true; // false will make test_aave_generates_yield fail since mock tokens do not support yield.
```
jkrivine: Arguably that test should be removed because it is essentially testing the correctness of AAVE not the aave pooler

jkrivine (approved):

PR #230, opened 2023-02-20
feat: reader.MGV is public

adhusson:
Reduces odds of giving an incompatible reader/mgv to scripts (just mgv is enough)


dontrolle (approved):
lgtm

PR #237, opened 2023-02-24
feat: script for mangrove.js (gets slot necessary to deal token)

adhusson:


peterMangrove (approved):
Would very much like an explanation of how it works, at some point 😄

adhusson:
Explained! https://github.com/mangrovedao/mangrove-core/blob/master/script/lib/GetTokenDealSlot.sol

PR #252, opened 2023-03-14
Feat/auditPhase1Patches

jkrivine:


addresses/deployed/maticmum.json
```diff
@@ -1,22 +1,66 @@
...
+    "address": "0x3EFed2C3291E62a2c5B493737b78F1af18812b97",
     "name": "MangroveOrderEnriched"
+  },
+  {
+    "address": "0x67CbCa7f1294cfF275c2e156D23F754Ba8Cd1D67",
+    "name": "Kandel_WETH_USDC"
+  },
+  {
+    "address": "0x349f8e8BbD5f1478A65fCCf0D354E24Ce0cac866",
+    "name": "Kandel_WETH_DAI_on_v7"
```
lnist: I would remove all these spurious addresses.

src/strategies/offer_maker/abstract/Direct.sol
```diff
@@ -18,6 +18,9 @@ import {IOfferLogic} from "mgv_src/strategies/interfaces/IOfferLogic.sol";
 
 ///@title `Direct` strats is an extension of MangroveOffer that allows contract's admin to manage offers on Mangrove.
 abstract contract Direct is MangroveOffer {
+  ///@notice `reserveId` is set in the constructor
+  event SetReserveId(address indexed reserveId);
```
lnist, line 23: Consider adding base/quote event to HasIndexedBidsAndAsks
Consider adding mgv event somewhere so that we know which mgv a contract is attached to. We can derive it from seeder, but an event is easier to discover.

addresses/deployed/maticmum.json
lnist: See https://github.com/mangrovedao/mangrove-core/pull/259/files

src/strategies/offer_maker/abstract/Direct.sol
lnist: See https://github.com/mangrovedao/mangrove-core/pull/259/files

PR #253, opened 2023-03-14
feat: no env vars config section

adhusson:


lnist (commented):
Looks good - in case rpc_endpoints still fail in CI we'll just add them to the profile.

lnist (approved):

PR #264, opened 2023-03-17
feat(Kandel): Add populate/retract start/end events

lnist:


src/strategies/offer_maker/market_making/kandel/abstract/DirectWithBidsAndAsksDistribution.sol
```diff
@@ -90,7 +98,7 @@ abstract contract DirectWithBidsAndAsksDistribution is Direct, HasIndexedBidsAnd
 
       populateIndex(OfferType.Ask, offerIdOfIndex(OfferType.Ask, index), index, args);
     }
-    emit Populate();
+    emit PopulateEnd();
```
jkrivine, line 101: So if you update by chunks, e.g [0, 50[ and [50,100[ you'll get:
`Populate(); PopulateEnd(); Populate(); PopulateEnd()`. Is this expected?
We could use `Populate(bytes memory indices)` with:
```solidity
bytes memory indices_enc = abi.encode(indices);
emit Populate(indices_enc)
```

src/strategies/offer_maker/market_making/kandel/abstract/DirectWithBidsAndAsksDistribution.sol
```diff
@@ -148,5 +157,6 @@ abstract contract DirectWithBidsAndAsksDistribution is Direct, HasIndexedBidsAnd
         _retractOffer(outbound_tknBid, inbound_tknBid, offerId, true);
       }
     }
+    emit RetractEnd();
```
jkrivine, line 160: Same remark here.

src/strategies/offer_maker/market_making/kandel/abstract/DirectWithBidsAndAsksDistribution.sol
lnist: It is more expensive to write the data and it does not offer more info. Either the populate will also emit SetIndexMapping in which case we will know the ids, or it will emit offerWrite which has the offerid and old SetIndexMapping can be used to map that to indices. Same for retract - it will retract offers for which SetIndexMapping has already been emitted.
But I agree that it is not clear whether it is a full populate or first of a batch of populates. There is not enough data in the call to decide that.  Note: UX wise we have agreed that its fine that we do not show a single log when a batch is required - it is ok that it ends up being two different logs.

PR #266, opened 2023-03-18
fix(Kandel): Start/End events and increase precision

lnist:


src/strategies/offer_maker/market_making/kandel/abstract/CoreKandel.sol
```diff
@@ -47,7 +47,7 @@ abstract contract CoreKandel is DirectWithBidsAndAsksDistribution, TradesBaseQuo
   function logPopulateStatus(uint offerId, OfferArgs memory args, bytes32 populateStatus) internal {
     if (
       populateStatus == REPOST_SUCCESS || populateStatus == NEW_OFFER_SUCCESS
-        || populateStatus == "mgv/writeOffer/density/tooLow"
+        || populateStatus == "mgv/writeOffer/density/tooLow" || populateStatus == 0x0
```
jkrivine: When do we emit nothing in populate? Maybe we should ?

lnist: Its in all the gives or wants being 0 branches of populateIndex in DirectWithBidsAndAsksDistribution - do you think we should return something there?

lnist: Updated with return value instead of 0x0

jkrivine (approved):

PR #268, opened 2023-03-20
feat/oracle1559

adhusson:
- chore: cleanup MgvOracle auth
- feat: automated MgvOracle through EIP1559

### DIFF
--- a/src/periphery/MgvOracle1559.sol
+++ b/src/periphery/MgvOracle1559.sol
@@ -1,0 +1,72 @@
+// SPDX-License-Identifier:	AGPL-3.0
+
+// MgvOracle.sol
+
+// Copyright (C) 2021 ADDMA.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published
+// by the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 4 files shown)


src/periphery/MgvOracle1559.sol
```diff
@@ -0,0 +1,72 @@
...
+    return tx.gasprice - block.basefee;
+  }
+
+  // Update gas scaling factor
+  function setGasScaling(uint gasScaling) external {
+    authOrMutatorOnly();
+    lastReceivedGasScaling = gasScaling;
+  }
+
+  // You must make sure that the chosen priorityFee and gasScaling cannot trigger overflow when calling the read function. Otherwise, a Mangrove with useOracle set to true will be locked.
```
espendk, line 62: This seems a bit risky. I suppose the mutator is responsible for ensuring that overflow cannot happen; Would be good to state explicitly what invariants must hold in terms of the parameters the mutator controls.

src/periphery/MgvOracle1559.sol
```diff
@@ -0,0 +1,72 @@
+// SPDX-License-Identifier:	AGPL-3.0
```
espendk, line 1: This should be BUSL-1.1 which  `MgvOracle` and the other core contracts were also updated to recently.

src/periphery/MgvOracle1559.sol
```diff
@@ -0,0 +1,72 @@
...
+// by the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
```
espendk, line 18: Should be removed.

PR #288, opened 2023-04-12
refactor: Rename MumbaiDeploy to MumbaiMangroveDeployer

espendk:
We've decided to use the following naming convention for deployment scripts:

- `<Contract>Deployer`: The general deployment script for `<Contract>`
- `<Chain><Contract>Deployer`: The chain-specific deployment script for `<Contract>` on `<Chain>`.


adhusson:
Should this still be merged?

espendk:
I think it should as it's still an improvement: Improved naming and consistency between file name and contract name 🙂 And I'm not sure we've settled on a final solution, so this will stick around for little while.

adhusson (approved):

PR #291, opened 2023-04-15
feat: Configure oracle on deploy

espendk:
This PR simplifies deployment and configuration related to MgvOracle:

1. Previously, the MgvOracle mutator was set to `chief` on deploy and then changed to `gasbot` when Mangrove was configured to use the oracle. Now, the mutator is immediately set to gasbot.

2. The script to configure Mangrove to use an oracle has been simplified and now supports any oracle, not just MgvOracle.

3. Mangrove is now configured to use MgvOracle as part of the deployment script (`MangroveDeployer`).

### DIFF
--- a/script/MangroveDeployer.s.sol
+++ b/script/MangroveDeployer.s.sol
@@ -20,12 +20,13 @@ contract MangroveDeployer is Deployer {
     innerRun({
       chief: envHas("CHIEF") ? vm.envAddress("CHIEF") : broadcaster(),
       gasprice: envHas("GASPRICE") ? vm.envUint("GASPRICE") : 1,
-      gasmax: envHas("GASMAX") ? vm.envUint("GASMAX") : 2_000_000
+      gasmax: envHas("GASMAX") ? vm.envUint("GASMAX") : 2_000_000,
+      gasbot: envHas("GASBOT") ? vm.envAddress("GASBOT") : fork.get("Gasbot")
     });
     outputDeployment();
   }
 
-  function innerRun(address chief, uint gasprice, uint gasmax) public {
+  function innerRun(address chief, uint gasprice, uint gasmax, address gasbot) public {
     broadcast();
     if (forMultisig) {
       mgv = new Mangrove{salt:salt}({governance: chief, gasprice: gasprice, gasmax: gasmax});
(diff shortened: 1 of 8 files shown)


script/MangroveDeployer.s.sol
```diff
@@ -20,32 +20,40 @@ contract MangroveDeployer is Deployer {
...
+      oracle = new MgvOracle{salt: salt}({governance_: chief, initialMutator_: gasbot});
     } else {
-      mgv = new Mangrove({governance: chief, gasprice: gasprice, gasmax: gasmax});
+      oracle = new MgvOracle({governance_: chief, initialMutator_: gasbot});
+    }
+    fork.set("MgvOracle", address(oracle));
+
+    broadcast();
+    if (forMultisig) {
+      mgv = new Mangrove{salt:salt}({governance: broadcaster(), gasprice: gasprice, gasmax: gasmax});
```
lnist, line 40: I see we invoke broadcaster() in other places after broadcast() - the code for broadcaster is not trivial - maybe we should add a comment to it that it does not cause transactions so that this is obvious?

script/MangroveDeployer.s.sol
```diff
@@ -20,32 +20,40 @@ contract MangroveDeployer is Deployer {
...
+    } else {
+      mgv = new Mangrove({governance: broadcaster(), gasprice: gasprice, gasmax: gasmax});
     }
     fork.set("Mangrove", address(mgv));
 
+    broadcast();
+    mgv.setMonitor(address(oracle));
+    broadcast();
+    mgv.setUseOracle(true);
```
lnist, line 49: This means we'll always use an oracle for instance also for instance for MangroveJsDeploy where I do not see it being used today. Is that intended? Maybe we have JS-tests that will break if they try to set gasprices

espendk: Good point. The intention was to have one script that deploys & configures all core&periphery contracts (Mangrove, MgvReader, MgvOracle, and MangroveOrder (tbd)) in a default setup across environments.

There are indeed JS-tests that set the gasprice.

Since we're planning to use a gas oracle in all environments, I'd suggest we adapt those JS tests to set the gasprice via the oracle.

adhusson (approved):
Looks fine as-is to me!

PR #293, opened 2023-04-15
feat: Update market status in MgvReader as part of the ActivateMarket script

espendk:


lnist (approved):

lnist:
If updateMarket was permissioned then I would argue the steps should be kept separate. But since its permissionless then we can just as well do both here.

PR #308, opened 2023-05-04
docs: Change license of Mangrove core to Business Source License 1.1

Louis-Amas:


espendk (commented):
It seems some of the files that should be covered by BUSL are missing:

billede

espendk (changes requested):
The LICENSE file should still contain all the other licenses (which we still refer to in other files), the BUSL license should just be added (instead of replacing everything).

Louis-Amas:
> It seems some of the files that should be covered by BUSL are missing:
> 
> billede

I used the one you updated below in the Asana task
billede

espendk:
> > It seems some of the files that should be covered by BUSL are missing:
> > billede
> 
> I used the one you updated below in the Asana task billede

which lists the exact same files as your list ;-)

src/AbstractMangrove.sol
```diff
@@ -1,4 +1,4 @@
-// SPDX-License-Identifier:	AGPL-3.0
+// SPDX-License-Identifier:	BUSL-1.1
```
espendk:
```suggestion
// SPDX-License-Identifier: BUSL-1.1
```

LICENSE
espendk, line 1: I'll push a revision of this.

espendk (changes requested):

espendk (approved):
Phil has reviewed and approves.

PR #311, opened 2023-05-15
feat(Core): Add deprovision flag to OfferRetract event

lnist:
When an offer is retracted we cannot discern whether it was deprovisioned by looking at events.

We do emit a `Credit` event in case deprovision is true, so one could try to discern deprovisioning based on whether that event is emitted immediately prior to the `OfferRetract` event. However, that event is also emitted by, e.g., `fund`, so looking at the event sequence within a transaction is not enough in case both retract and fund is invoked in the same TX. This is actually possible with Kandel's populate which can retract an offer, and fund mangrove in one go.

The solution is to add the "deprovision" flag to the event. This is a very tiny change, so should be good wrt audits, and the cost of the event is only applied when retractOffer is invoked.

With this change we can for all offers discern their locked provision via events, and thus the sum of that and mgv.balanceOf(maker) is the maker's entire balance on mangrove.

Locked provision is `
```solidity
provision = 10 ** 9 * offerDetail.gasprice() //gasprice is 0 if offer was deprovisioned
    * (offerDetail.gasreq() + offerDetail.offer_gasbase());
```

We do not emit offer_gasbase during `OfferWrite`, but at that point it is equal to the offer lists `offer_gasbase()`, and we do get events when that change.

1. Live offers locks provision - OfferWrite events can monitor this.
2. Taken offers locks provision (if successful), so same calculation as during initial OfferWrite applies
3. Failed offers subtracts penalty, and frees the rest of the provision to be read by `balanceOf`.
4. RetractOffer can be discerned with the new change.

Note, in the future we may want to extend the Debit, Credit events with Outbound, Inbound, OfferId for more explicit tracking. For now we will deduce provision based on the circumstantial evidence of other events.

### DIFF
--- a/src/IMangrove.sol
+++ b/src/IMangrove.sol
@@ -25,7 +25,7 @@ interface IMangrove {
     uint takerGives,
     bytes32 mgvData
   );
-  event OfferRetract(address indexed outbound_tkn, address indexed inbound_tkn, uint id);
+  event OfferRetract(address indexed outbound_tkn, address indexed inbound_tkn, uint id, bool deprovision);
   event OfferSuccess(
     address indexed outbound_tkn, address indexed inbound_tkn, uint id, address taker, uint takerWants, uint takerGives
   );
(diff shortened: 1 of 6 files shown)


peterMangrove (approved):
lgtm

adhusson (approved):
looks good :)

PR #321, opened 2023-05-17
Feat/improve preproc

adhusson:
- Remove `solpp` (the package is deprecated)
- In preparation for ticks, enable arbitrary-sized struct fields
- bools are 1 bit (they were 8 bits)


src/preprocessed/MgvGlobal.post.sol
```diff
@@ -40,104 +28,126 @@ struct GlobalUnpacked {
...
-    return (((GlobalPacked.unwrap(__packed) << useOracle_before) >> (256-useOracle_bits)) > 0);
+    return (((GlobalPacked.unwrap(__packed) << useOracle_before) >> (256 - useOracle_bits)) > 0);
   }}
+
   function useOracle(GlobalPacked __packed,bool val) internal pure returns(GlobalPacked) { unchecked {
-    return GlobalPacked.wrap((GlobalPacked.unwrap(__packed) & useOracle_mask)
-                                | ((uint_of_bool(val) << (256-useOracle_bits) >> useOracle_before)));
+    uint __clean_struct = GlobalPacked.unwrap(__packed) & useOracle_mask;
+    uint __clean_field  = (uint_of_bool(val) << (256 - useOracle_bits)) >> useOracle_before;
+    return GlobalPacked.wrap(__clean_struct | __clean_field);
```
lnist: Previously we did it all in a return, now we assign and then finally return. Have you verified gas consumption?

tsconfig.json
```diff
@@ -0,0 +1,107 @@
+{
+  "compilerOptions": {
+    /* Visit https://aka.ms/tsconfig to read more about this file */
+
+    /* Projects */
```
lnist: slim this file?

lnist (commented):
I assume this is highly covered by tests, but some explicit foundry fuzz/invariant testing of our structures would be great, so unless we already have that, then consider adding them :)

adhusson:
Putting this as a top-level comment for visibility, but this is a reply to @lnist's remark on how gas usage has evolved.

My initial observation is that gas usage has increased by a nonmateiral amount, e.g. a market order test that consumes about 6M gas now costs ~400 additional gas. This was not due to the `pack` function though.

After extensive testing, here are my findings:
1. Moving from 8bit bools to 1bit bools, increases gas use a small amount. I've looked at the optimized assembly and don't understand why.
2. Using temporary variables in the setters (e.g. `Local.active(bool)`) was the other part of the added gas cost. I will revert this to an inline version.
3. `Local.pack` without a temporary variable (ie. everything inlined of the form `return (((0 |  field1) | field2) | field3)`) does not work without `via_ir` and for now we do not use `via_ir`. `Local.pack` hasn't been used so far which is why it was not found until now. So I'll keep the temp var.

lnist (approved):

test/preprocessed/MgvGlobalTest.post.sol
```diff
@@ -0,0 +1,161 @@
...
+    }
+  function test_set_notify(MgvStructs.GlobalPacked packed,bool notify) public {
+      MgvStructs.GlobalPacked original = packed.notify(packed.notify());
+      assertEq(original.notify(),packed.notify(), "original: bad notify");
+
+      MgvStructs.GlobalPacked modified = packed.notify(notify);
+
+      assertEq(modified.notify(),notify,"modified: bad notify");
+
+      assertEq(modified.monitor(),modified.monitor(),"modified: bad monitor");
```
lnist: Should modified be packed on one side og the assert? This seems to only assert get does not modify?

adhusson: Thanks! goal was to ensure that the setter above did not modify the other values. The `modified` on the right should be `packed`.

lnist (approved):

PR #324, opened 2023-05-19
docs(license): License cleanup

espendk:
This PR cleans up two issues with licenses:

1. Some file headers contained incompatible SPDX identifiers and license texts
It is sufficient to include the SPDX licence identifier in the header as the license terms are stated in the LICENSE file.
We've therefore minimized the license information in the file headers to just the SPDX identifier comment.

1. Invalid SPDX identifier 'UNLICENSED' used in some file headers
'UNLICENSED' is an invalid SPDX identifier with ambiguous meaning: Some interpret it as public domain others as the opposite, i.e. proprietary.

    Our intent has been for files tagged with UNLICENSED to be public domain. The Unlicense license captures that well, so we've put all files previously tagged with UNLICENSED under the Unlicense license.


adhusson (approved):
good to go

PR #329, opened 2023-05-21
feat(MgvOracle): MgvOracle initial gasprice and ownership transfer 

espendk:
Allow initial gas price to be set in MgvOracle.
Allow governance of MgvOracle to transfer ownership of the oracle.


peterMangrove (approved):
Lgtm

PR #334, opened 2023-05-23
KandelSeederDeployer improvements

espendk:
This PR includes two fixes and an improvement to the KandelSeederDeployer scripts:

# Fixes
**Workaround for forge bug in KandelSeederDeployer**
forge has a bug where it fails to increment the broadcaster nonce when deploying a contract which itself deploys a contract.
This is bug is triggered when deploying AaveKandelSeeder which deploys a router contract.

A workaround has been implemented which forces the nonce to be incremented if forge has not done it.

**Ensure market is open in KandelSeederDeployer smoketest**
The smoketest assumed that the WETH/DAI market would be open, but that need not be the case, eg on a new Mangrove instance.
The smoketest now pranks MgvGovernance and opens the market.

# Improvement
The script now deploys a Kandel and an AaveKandel instance. By deploying these instances as part of the KandelSeederDeployer script, the contracts will be verified and later instances will be recognized as matching their code.

### DIFF
--- a/script/strategies/kandel/deployers/KandelSeederDeployer.s.sol
+++ b/script/strategies/kandel/deployers/KandelSeederDeployer.s.sol
@@ -2,8 +2,8 @@
 pragma solidity ^0.8.13;
 
 import {Script, console} from "forge-std/Script.sol";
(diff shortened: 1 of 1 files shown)


dontrolle (approved):
lgtm

PR #335, opened 2023-05-23
Feat/density automation

jkrivine:
This PR addresses fixes density computation for testnet deployment. It does the following:
* uses onchain oracle to convert token values into USD and matic into USD
* uses the oracle's prices to infer density.
* allows gasprice override to during market opening (this allows one to consider a gasprice for density different from Mangrove's gasprice).
* refines computation of gasbase during semibook opening

### DIFF
--- a/script/core/ActivateMarket.s.sol
+++ b/script/core/ActivateMarket.s.sol
@@ -17,13 +17,15 @@ import {ActivateSemibook} from "./ActivateSemibook.s.sol";
  forge script --fork-url mumbai ActivateMarket*/
 
 contract ActivateMarket is Deployer {
+  uint currentGasprice;
+
   function run() public {
     Mangrove mgv_ = Mangrove(envAddressOrName("MGV", "Mangrove"));
     (MgvStructs.GlobalPacked global,) = mgv_.config(address(0), address(0));
+    currentGasprice = global.gasprice();
 
     innerRun({
       mgv: mgv_,
-      gasprice: global.gasprice(),
       reader: MgvReader(envAddressOrName("MGV_READER", "MgvReader")),
       tkn1: IERC20(envAddressOrName("TKN1")),
       tkn2: IERC20(envAddressOrName("TKN2")),
@@ -35,6 +37,7 @@ contract ActivateMarket is Deployer {
 
   /* Activates a market on mangrove. Two semibooks are activated, one where the first tokens is outbound and the second inbound, and the reverse.
     mgv: mangrove address
+    gaspriceOverride: overrides current mangrove's gasprice for the computation of density - default innerRun uses mangrove's gasprice
     tkn1: first tokens
     tkn2: second tokens,
     tkn1_in_gwei: price of one tkn1 (display units) in gwei
@@ -51,7 +54,7 @@ contract ActivateMarket is Deployer {
   */
   function innerRun(
     Mangrove mgv,
-    uint gasprice,
+    uint gaspriceOverride,
     MgvReader reader,
     IERC20 tkn1,
     IERC20 tkn2,
@@ -61,7 +64,7 @@ contract ActivateMarket is Deployer {
   ) public {
     new ActivateSemibook().innerRun({
       mgv: mgv,
-      gasprice: gasprice,
+      gaspriceOverride: gaspriceOverride,
       outbound_tkn: tkn1,
       inbound_tkn: tkn2,
       outbound_in_gwei: tkn1_in_gwei,
@@ -70,7 +73,7 @@ contract ActivateMarket is Deployer {
 
     new ActivateSemibook().innerRun({
       mgv: mgv,
-      gasprice: gasprice,
+      gaspriceOverride: gaspriceOverride,
       outbound_tkn: tkn2,
       inbound_tkn: tkn1,
       outbound_in_gwei: tkn2_in_gwei,
@@ -78,5 +81,20 @@ contract ActivateMarket is Deployer {
     });
 
     new UpdateMarket().innerRun({tkn0: tkn1, tkn1: tkn2, reader: reader});
+  }
+
+  /**
+   * innerRun with no gasprice override (Mangrove's gasprice is used to compute density)
+   */
+  function innerRun(
+    Mangrove mgv,
+    MgvReader reader,
+    IERC20 tkn1,
+    IERC20 tkn2,
+    uint tkn1_in_gwei,
+    uint tkn2_in_gwei,
+    uint fee
+  ) public {
+    innerRun(mgv, currentGasprice, reader, tkn1, tkn2, tkn1_in_gwei, tkn2_in_gwei, fee);
   }
 }
--- a/script/core/ActivateMarket.s.sol
+++ b/script/core/ActivateMarket.s.sol
@@ -18,8 +18,12 @@ import {ActivateSemibook} from "./ActivateSemibook.s.sol";
 
 contract ActivateMarket is Deployer {
   function run() public {
+    Mangrove mgv_ = Mangrove(envAddressOrName("MGV", "Mangrove"));
+    (MgvStructs.GlobalPacked global,) = mgv_.config(address(0), address(0));
+
     innerRun({
-      mgv: Mangrove(envAddressOrName("MGV", "Mangrove")),
+      mgv: mgv_,
+      gasprice: global.gasprice(),
       reader: MgvReader(envAddressOrName("MGV_READER", "MgvReader")),
       tkn1: IERC20(envAddressOrName("TKN1")),
       tkn2: IERC20(envAddressOrName("TKN2")),
@@ -47,6 +51,7 @@ contract ActivateMarket is Deployer {
   */
   function innerRun(
     Mangrove mgv,
+    uint gasprice,
     MgvReader reader,
     IERC20 tkn1,
     IERC20 tkn2,
@@ -56,6 +61,7 @@ contract ActivateMarket is Deployer {
   ) public {
     new ActivateSemibook().innerRun({
       mgv: mgv,
+      gasprice: gasprice,
       outbound_tkn: tkn1,
       inbound_tkn: tkn2,
       outbound_in_gwei: tkn1_in_gwei,
@@ -64,6 +70,7 @@ contract ActivateMarket is Deployer {
 
     new ActivateSemibook().innerRun({
       mgv: mgv,
+      gasprice: gasprice,
       outbound_tkn: tkn2,
       inbound_tkn: tkn1,
       outbound_in_gwei: tkn2_in_gwei,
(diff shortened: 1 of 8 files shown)


script/deployers/MumbaiMangroveFullTestnetDeployer.s.sol
```diff
@@ -51,31 +64,41 @@ contract MumbaiMangroveFullTestnetDeployer is Deployer {
...
+
+    uint[] memory prices = priceOracle.getAssetsPrices(dynamic([address(dai), address(usdc), address(weth)]));
+    maticPrice = priceOracle.getAssetPrice(fork.get("WMATIC"));
+
+    // 1 token_i = (prices[i] / 10**8) USD
+    // 1 USD = (10**8 / maticPrice) Matic
+    // 1 token_i = (prices[i] * 10**9 / maticPrice) gwei of Matic
     new ActivateMarket().innerRun({
       mgv: mgv,
+      gasprice: 140,
```
espendk: Why does this not use the same gasprice as Mangrove?

src/strategies/vendor/aave/v3/IPriceOracleGetter.sol
```diff
@@ -5,26 +5,32 @@ pragma solidity ^0.8.10;
...
   /**
    * @notice Returns the asset price in the base currency
    * @param asset The address of the asset
    * @return The price of the asset
-   **/
+   *
+   */
   function getAssetPrice(address asset) external view returns (uint);
+
+  function getAssetsPrices(address[] calldata _assets) external view returns (uint[] memory);
```
espendk: comment please :-)

script/core/ActivateSemibook.s.sol
```diff
@@ -14,26 +14,37 @@ uint constant COVER_FACTOR = 1000;
...
       inbound_tkn: IERC20(envAddressOrName("INBOUND_TKN")),
       outbound_in_gwei: vm.envUint("OUTBOUND_IN_GWEI"),
       fee: vm.envUint("FEE")
     });
   }
 
-  function innerRun(Mangrove mgv, IERC20 outbound_tkn, IERC20 inbound_tkn, uint outbound_in_gwei, uint fee) public {
+  function innerRun(
+    Mangrove mgv,
+    uint gasprice,
```
espendk: Why allow the gasprice used in market activation to differ from Mangrove's gasprice? Seem it'll make it easy to open a market with inconsistent parameters?

espendk:
> * allows gasprice override to during market opening (this allows one to consider a gasprice for density different from Mangrove's gasprice).

It also requires callers of the market activation script to feed in a gasprice. This seems to create a risk of markets being inconsistent with Mangrove.

If we indeed want this option, I don't think it should be the default. Default should be a market config that is consistent with Mangrove's config.

script/deployers/MumbaiMangroveFullTestnetDeployer.s.sol
jkrivine: I tried to explain in the description of the PR above. The idea is to be able to set a realistic density (the one we expect to see on mainnet) while keeping a low gasprice in order to avoid requiring tester to provide too many matics.

script/core/ActivateSemibook.s.sol
jkrivine: This is actually a way to override. `gasprice` is actually used for both bounty provision computation and density. It would totally make sense to require density to be higher/lower than the current gasprice.

jkrivine:
> > * allows gasprice override to during market opening (this allows one to consider a gasprice for density different from Mangrove's gasprice).
> 
> It also requires callers of the market activation script to feed in a gasprice. This seems to create a risk of markets being inconsistent with Mangrove.
> 
> If we indeed want this option, I don't think it should be the default. Default should be a market config that is consistent with Mangrove's config.

But this *is* actually the default. If you call `ActivateMarket.run()` or `ActivateSemiBook.run()`. You can override `gasprice` only if you bypass the `run` and directly call to `innerRun`

espendk:
> But this is actually the default. If you call ActivateMarket.run() or ActivateSemiBook.run(). You can override gasprice only if you bypass the run and directly call to innerRun.

It's only the default of you call the script from the command line. When composing deployer scripts (as we do and which is the purpose of  innerRun) there's no default.

jkrivine:
> It's only the default of you call the script from the command line. When composing deployer scripts (as we do and which is the purpose of innerRun) there's no default.

But why would you call `innerRun` instead of `run` in that case?

script/deployers/MumbaiMangroveFullTestnetDeployer.s.sol
espendk: Ok. This explanation should be in this script as it's non-obvious and will soon be forgotten :-)

jkrivine: I'll push something

espendk:
> > It's only the default of you call the script from the command line. When composing deployer scripts (as we do and which is the purpose of innerRun) there's no default.
> 
> But why would you call `innerRun` instead of `run` in that case?

Because `run` serves two purposes that only the outermost script should do (see also the `README` in the  scripts dir):
1. Resolves command line parameters/env vars
2. Outputs deployment.

script/core/ActivateSemibook.s.sol
espendk: Since Solidity doesn't allow optional parameters, I suggest reverting the `innerRun` signature and instead allowing the override in a `innerRunWithOverrides` function where `gasprice` can be overridden.

script/deployers/MumbaiMangroveFullTestnetDeployer.s.sol
```diff
@@ -29,16 +33,25 @@ import {MgvReader} from "mgv_src/periphery/MgvReader.sol";
  * - open markets: DAI/USDC, WETH/DAI, WETH/USDC
  */
 contract MumbaiMangroveFullTestnetDeployer is Deployer {
+  uint internal maticPrice;
```
lnist, line 38: add comment describing the unit of this

script/deployers/MumbaiMangroveFullTestnetDeployer.s.sol
```diff
@@ -29,16 +33,25 @@ import {MgvReader} from "mgv_src/periphery/MgvReader.sol";
...
+
   function run() public {
     runWithChainSpecificParams();
     outputDeployment();
   }
 
+  function toGweiOfMatic(uint price) internal view returns (uint) {
```
lnist, line 45: add comment on unit of price

script/deployers/MumbaiMangroveFullTestnetDeployer.s.sol
```diff
@@ -29,16 +33,25 @@ import {MgvReader} from "mgv_src/periphery/MgvReader.sol";
...
+    return (price * 10 ** 9) / maticPrice;
+  }
+
   function runWithChainSpecificParams() public {
     // Deploy Mangrove
     new MumbaiMangroveDeployer().runWithChainSpecificParams();
     Mangrove mgv = Mangrove(fork.get("Mangrove"));
     MgvReader reader = MgvReader(fork.get("MgvReader"));
+    IPriceOracleGetter priceOracle =
+      IPriceOracleGetter(IPoolAddressesProvider(fork.get("Aave")).getAddress("PRICE_ORACLE"));
```
lnist, line 55: Are we fairly confident on the prices aave provide? but I see we only use it for Mumbai right now, so that probably makes it ok.

jkrivine: Yes we can be confident. Actually the prices on mumbai are fixed (1 USD for stables and 4000 for ether) so the oracle solution is even better on mainnet.

script/core/ActivateSemibook.s.sol
jkrivine: > Since Solidity doesn't allow optional parameters, I suggest reverting the innerRun signature and instead allowing the override in a innerRunWithOverrides function where gasprice can be overridden.

Addressed in the latest push

adhusson:
I have comments too 😃 I find the `innerRunWithOverrides` confusing. I agree with the purpose of run function but I think that also includes "resolves defaults". So I would have a single `innerRun` with a `gasprice` argument. And then `run` would do the default-resolution with

```solidity
innerRun({
   ...,
   gasprice: vm.envOr("GASPRICE",global.gasprice())
});
```

Also currently ActivateSemibook's `innerRun` takes a gasprice argument so at least they should be in sync, in whichever direction?

espendk:
> but I think that also includes "resolves defaults"

As mentioned above with that is the composition of deployers. Other deployers would need to bake in the defaults that must be passed to `innerRun` and so the defaults will spread - and probably be filled out wrong.

I agree that that `innerRunWithOverrides` is not ideal, but we currently only need this in one place and I don't see a way to do it which doesn't negatively impact many places. And I don't believe we should optimize the code for allowing inconsistent values across deployers - this should be the exception, not the rule.

So for now, I think it's the better compromise to have `innerRunWithOverrides`.

> Also currently ActivateSemibook's innerRun takes a gasprice argument so at least they should be in sync, in whichever direction?

Of course. I assumed this would be updated consistently with `ActivateMarket`, ie have a `innerRunWithOverrides` :-)

jkrivine:
> > but I think that also includes "resolves defaults"
> 
> As mentioned above with that is the composition of deployers. Other deployers would need to bake in the defaults that must be passed to `innerRun` and so the defaults will spread - and probably be filled out wrong.
> 
> I agree that that `innerRunWithOverrides` is not ideal, but we currently only need this in one place and I don't see a way to do it which doesn't negatively impact many places. And I don't believe we should optimize the code for allowing inconsistent values across deployers - this should be the exception, not the rule.
> 
> So for now, I think it's the better compromise to have `innerRunWithOverrides`.
> 
> > Also currently ActivateSemibook's innerRun takes a gasprice argument so at least they should be in sync, in whichever direction?
> 
> Of course. I assumed this would be updated consistently with `ActivateMarket`, ie have a `innerRunWithOverrides` :-)

If I propagate `innerRunWithOverrides` everywhere the code starts to be super ugly and redundant.
Function overload is the standard way to deal with optional parameters in solidity. I think reverting to two `innerRun`, one that allows overriding gasprice for density and the other using mangrove's gasprice sounds much better.

espendk:
> I think reverting to two `innerRun`, one that allows overriding gasprice for density and the other using mangrove's gasprice sounds much better.
> 

Agreed!

script/core/ActivateMarket.s.sol
```diff
@@ -72,4 +76,20 @@ contract ActivateMarket is Deployer {
...
+   */
+  function innerRun(
+    Mangrove mgv,
+    MgvReader reader,
+    IERC20 tkn1,
+    IERC20 tkn2,
+    uint tkn1_in_gwei,
+    uint tkn2_in_gwei,
+    uint fee
+  ) public {
```
espendk: Since this is the primary function signature for calling this script from other scripts, I think it should be above the special case with overrides + the doc should be complete.

Same for the corresponding function in `ActivateSemibook`.

espendk (approved):
👍🙏

PR #341, opened 2023-05-26
fix: disable test that fails when optimizer=true

adhusson:


test/strategies/unit/AavePooledRouter.t.sol
```diff
@@ -220,7 +220,9 @@ contract AavePooledRouterTest is OfferLogicTest {
     assertEq(pooledRouter.sharesOf(dai, maker1), 0, "Incorrect shares");
   }
 
-  function test_mockup_marketOrder_gas_cost() public {
+  // disabled; only works without optimizer=false
```
espendk, line 223: Mismatch with title - maybe this instead?
```suggestion
  // disabled; only works with optimizer=false
```

lnist: Could we do this with a profile instead - set optimization levels via profiles, and then read some profile-specific flag to skip the test? The test is there to make sure our gasreqs are meaningful, so I dislike disabling it.

adhusson:
closing as stale

PR #348, opened 2023-05-30
script to activate a market on matic chain

jkrivine:
Script to activate a given market on matic chain (using matic price as reference for gas price)

### DIFF
--- a/script/deployers/MumbaiActivateMarket.s.sol
+++ b/script/deployers/MumbaiActivateMarket.s.sol
@@ -1,0 +1,96 @@
+// SPDX-License-Identifier:	AGPL-3.0
+pragma solidity ^0.8.13;
+
+import {ToyENS} from "mgv_lib/ToyENS.sol";
+
+import {Deployer} from "mgv_script/lib/Deployer.sol";
+
+import {MumbaiMangroveDeployer} from "mgv_script/core/deployers/MumbaiMangroveDeployer.s.sol";
+import {MumbaiMangroveOrderDeployer} from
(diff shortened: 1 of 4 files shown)


script/deployers/MumbaiActivateMarket.s.sol
```diff
@@ -0,0 +1,84 @@
...
+    activateMarket(token0, token1, price0, price1);
+    outputDeployment();
+    smokeTest(token0, token1);
+  }
+
+  function toGweiOfMatic(uint price) internal view returns (uint) {
+    return (price * 10 ** 9) / maticPrice;
+  }
+
+  function activateMarket(IERC20 token0, IERC20 token1, uint price0, uint price1) public {
```
espendk, line 55: Should be `innerRun` I think

script/deployers/MumbaiActivateMarket.s.sol
```diff
@@ -0,0 +1,84 @@
...
+    Mangrove mgv = Mangrove(fork.get("Mangrove"));
+    MgvReader reader = MgvReader(fork.get("MgvReader"));
+    MangroveOrder mangroveOrder = MangroveOrder(fork.get("MangroveOrder"));
```
espendk, line 58: I think these should be resolved in `run`

espendk (changes requested):

espendk (approved):
As discussed in DM, let's merge as is. I'll suggest some adjustments and further development  on the Mumbai scripts in a new PR.

PR #351, opened 2023-05-30
fix: adapt MangroveJs&TestToken,test MangroveJs, partial revert of 326bcca140e4c737a8d225e844cd7c84094e12a3

adhusson:
This commit does the following:
- TestToken has not mint frequency limit when its `mintLimit = type(uint).max`
- MangroveJs is tested
- MangroveJs warns when its configured `chief` is not the broadcaster (it's possible but incoherent).
- MangroveJs removes all mint limits on its Test tokens
- Partial revert of 326bcca140e4c737a8d225e844cd7c84094e12a3: remove the added line `fork.set("MgvGovernance", chief);`

1. On that topic, to check my understanding: this line is not necessary thanks to the other half of 326bcca140e4c737a8d225e844cd7c84094e12a3, right?

2. Also, I would recommend the following, if agreed we can include it somewhere in the doc:
    - `MgvGovernance` should not be used by scripts except at the top level, most likely  by scripts that are chain-specific and bundle their entire config.
    - Other scripts should inherit who governance is from either a) an argument or b) `mgv.governance()`
    - The way to reference `MgvGovernance` from the outside is as follows (see related [comment](https://github.com/mangrovedao/mangrove-core/blob/af2ffc8bc44cf6bbbeaa9b0e51fda3c1657de61e/script/lib/Deployer.sol#L137-L142)):

      ```bash
      CHIEF=MgvGovernance forge script MangroveJsDeploy ...
      ```
      The Deployer will pickup the `CHIEF` value and look for the corresponding address.

3. I'm suspicious of MangroveJsDeploy even having a `chief` argument since it doesn't even work when `chief != broadcaster()`, what do you think?

### DIFF
--- a/script/toy/MangroveJs.s.sol
+++ b/script/toy/MangroveJs.s.sol
@@ -16,6 +16,7 @@ import {IMangrove} from "mgv_src/IMangrove.sol";
 import {Deployer} from "mgv_script/lib/Deployer.sol";
 import {ActivateMarket} from "mgv_script/core/ActivateMarket.s.sol";
 import {PoolAddressProviderMock} from "mgv_script/toy/AaveMock.sol";
+import "forge-std/console.sol";
 
 /* 
 This script prepares a local server for testing by mangrove.js.
@@ -26,13 +27,13 @@ interact with.  b) For any additional deployments needed, those files should be
 hosted in mangrove.js.*/
 
 contract MangroveJsDeploy is Deployer {
-  TestToken tokenA;
-  TestToken tokenB;
-  IERC20 dai;
-  IERC20 usdc;
-  IERC20 weth;
-  SimpleTestMaker simpleTestMaker;
-  MangroveOrder mgo;
+  TestToken public tokenA;
+  TestToken public tokenB;
+  IERC20 public dai;
+  IERC20 public usdc;
+  IERC20 public weth;
+  SimpleTestMaker public simpleTestMaker;
+  MangroveOrder public mgo;
 
   function run() public {
     innerRun({chief: broadcaster(), gasprice: 1, gasmax: 2_000_000, gasbot: broadcaster()});
@@ -40,7 +41,9 @@ contract MangroveJsDeploy is Deployer {
   }
 
   function innerRun(address chief, uint gasprice, uint gasmax, address gasbot) public {
-    fork.set("MgvGovernance", chief);
+    if (chief != broadcaster()) {
+      console.log("Warning: chief is not the broadcaster; deployed contracts will not obey their orders.");
+    }
     MangroveDeployer mgvDeployer = new MangroveDeployer();
 
     mgvDeployer.innerRun({chief: chief, gasprice: gasprice, gasmax: gasmax, gasbot: gasbot});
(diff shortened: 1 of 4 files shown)


script/toy/MangroveJs.s.sol
```diff
@@ -26,21 +27,23 @@ interact with.  b) For any additional deployments needed, those files should be
...
   function run() public {
     innerRun({chief: broadcaster(), gasprice: 1, gasmax: 2_000_000, gasbot: broadcaster()});
     outputDeployment();
   }
 
   function innerRun(address chief, uint gasprice, uint gasmax, address gasbot) public {
-    fork.set("MgvGovernance", chief);
+    if (chief != broadcaster()) {
+      console.log("Warning: chief is not the broadcaster; deployed contracts will not obey their orders.");
+    }
```
espendk: Maybe failing fast is better? This way it will be clear, that a core assumption is not met. Or, as you suggest in the PR description, to simply remove the `chief` argument to prevent the issue altogether.

script/toy/MangroveJs.s.sol
```diff
@@ -58,7 +61,11 @@ contract MangroveJsDeploy is Deployer {
       symbol: "TokenA",
       _decimals: 18
     });
+
+    broadcast();
+    tokenA.setMintLimit(type(uint).max);
     fork.set("TokenA", address(tokenA));
+
     broadcast();
     tokenA.setMintLimit(type(uint).max);
```
espendk: This seems like a merge error? No need to set mint limit twice, right?

Same for Token B below.

espendk (changes requested):

espendk:
Very nice!

>     * Partial revert of [326bcca](https://github.com/mangrovedao/mangrove-core/commit/326bcca140e4c737a8d225e844cd7c84094e12a3): remove the added line `fork.set("MgvGovernance", chief);`
> 
> 
>     1. On that topic, to check my understanding: this line is not necessary thanks to the other half of [326bcca](https://github.com/mangrovedao/mangrove-core/commit/326bcca140e4c737a8d225e844cd7c84094e12a3), right?

Yes, it was redundant.

>     2. Also, I would recommend the following, if agreed we can include it somewhere in the doc:
>        
>        * `MgvGovernance` should not be used by scripts except at the top level, most likely  by scripts that are chain-specific and bundle their entire config.

That makes sense. And good idea to document it. There's a README in the scripts dir where it could go.
Not sure how to state it as a general principle. Maybe scripts should only read fork names at the outermost `run` ?

>        * The way to reference `MgvGovernance` from the outside is as follows (see related [comment](https://github.com/mangrovedao/mangrove-core/blob/af2ffc8bc44cf6bbbeaa9b0e51fda3c1657de61e/script/lib/Deployer.sol#L137-L142)):
>          ```shell
>          CHIEF=MgvGovernance forge script MangroveJsDeploy ...
>          ```

For deployments, we should limit the use of env vars to secrets in order to ensure traceability and reproducibility.
That's why I put MgvGovernance in the context addresses.

>     3. I'm suspicious of MangroveJsDeploy even having a `chief` argument since it doesn't even work when `chief != broadcaster()`, what do you think?

Indeed. It sounds like the safest and clearest solution is to remove the `chief` argument.

espendk (approved):
lgtm

PR #358, opened 2023-06-08
gas: use more stack args for better()

adhusson:
improve on #357: use 3 stack args instead of 1 memory args that gets accessed everytime when calling `better`.

strict gas improvement, notably:

| name              | delta to develop (15c0e78b3d7e9ac5c791d269d36b45dfb3d27b1e)   | delta to #357 |
|-------------------|---------|-------------------|
| insert, move by 15 offers |  - 8.40% | -0.91% |
| insert, move by 49 offers |  - 8.12% | -1.07% |

Size is 23.074 kB (vs 23.067 on #357)

PR #360, opened 2023-06-10
chore/0.8.20

adhusson:
new default evm version (shanghai) is only deployed on eth mainnet, so this commit also specifies the former evm version as the default.

Note: when applied to #358 , Mangrove contract size is improved (from 22.986 kB to 22.872 kB), but gas tests regress:

image

### DIFF
--- a/src/toy_strategies/utils/SimpleOracle.sol
+++ b/src/toy_strategies/utils/SimpleOracle.sol
@@ -11,9 +11,9 @@ contract SimpleOracle is IOracle, AccessControlled {
   mapping(address => uint96) internal priceData;
 
   constructor(address base_, address admin) AccessControlled(admin) {
+    base_token = IERC20(base_);
     try IERC20(base_).decimals() returns (uint8 d) {
       require(d != 0, "Invalid decimals number for Oracle base");
-      base_token = IERC20(base_);
     } catch {
       revert("Invalid Oracle base address");
     }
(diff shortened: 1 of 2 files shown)


src/toy_strategies/utils/SimpleOracle.sol
```diff
@@ -11,9 +11,9 @@ contract SimpleOracle is IOracle, AccessControlled {
   mapping(address => uint96) internal priceData;
 
   constructor(address base_, address admin) AccessControlled(admin) {
+    base_token = IERC20(base_);
     try IERC20(base_).decimals() returns (uint8 d) {
```
espendk: Would perhaps be easier to read, if the `try` used the `base_token`:
```suggestion
    base_token = IERC20(base_);
    try base_token.decimals() returns (uint8 d) {
```

espendk (commented):
Are there any risks/issues to consider besides the default evm version and gas usage?
Wdyt of the changes in gas usage?

adhusson:
There is a "tradition" of not using the very latest version of Solidity (trying to extract some Lindy out of older versions) yes :)

espendk:
> There is a "tradition" of not using the very latest version of Solidity (trying to extract some Lindy out of older versions) yes :)

Ok :-) But you think we should upgrade now?
To be clear, I'm not opposed at all, would just like to understand the significance and pros/cons of this upgrade.

adhusson:
closed as we're on 0.8.20 already now

PR #363, opened 2023-06-12
doc: explain how to ignore oracle values

adhusson:


espendk (commented):
structs.js has the following description of `useOracle` which could maybe also mention that density is optional for oracles:

> If useOracle is true, the dex will use the monitor address as an oracle for gasprice and density, for every outbound_tkn/inbound_tkn pair.

espendk (approved):

PR #371, opened 2023-06-15
feat(ci): Simplify CI by moving away from pull_request_target

lnist:
Based on https://securitylab.github.com/research/github-actions-preventing-pwn-requests/ (more details in our context in https://github.com/mangrovedao/mangrove-ts/issues/691 ) pull_request_target can be used if fork PRs must access secrets or write to repository:
> Due to the dangers inherent to automatic processing of PRs, GitHub’s standard pull_request workflow trigger by default prevents write permissions and secrets access to the target repository. However, in some scenarios such access is needed to properly process the PR. To this end the pull_request_target workflow trigger was introduced.

Standard pull_request is much safer and along with that we can remove a lot of old workarounds.
Additionally, our old pull_request_target workflow had a flaw that you could make a forked PR that is innocent, wait for someone to add labels, and then push extra code to extract secrets after the safe label had been added.

Note, RPC node urls are still referenced in the workflow - for our own PRs it has access to our secrets and use those. For fork-PRs those are empty, and instead we now default to llamarpc which has a free demo endpoint. This is probably not as stable as alchemy, but for now we do not have a lot of PR traffic from Forks so this seems like a simple solution.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -65,12 +65,11 @@ jobs:
       run: yarn run build
 
     # Pinned forks are used in tests and utilize secrets. For PRs from forks these are not available
-    # Let the tests attempt to run with the cached state, but if they fail, then the tests have to be disabled
-    # until the PR is merged.
+    # Let the tests attempt to run with demo endpoints which are more unstable.
     - name: Mangrove Solidity Tests
       run: yarn run test
       env:
-        POLYGON_NODE_URL: ${{ secrets.POLYGON_NODE_URL || 'http://localhost:4444' }}
+        POLYGON_NODE_URL: ${{ secrets.POLYGON_NODE_URL || 'https://polygon-mumbai.g.alchemy.com/v2/demo' }}
         MUMBAI_NODE_URL: ${{ secrets.MUMBAI_NODE_URL || 'http://localhost:4444' }}
 
     # For push runs we also create a coverage report
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -8,7 +8,7 @@ on:
       - develop
       - 'run-ci/**'
   pull_request:
-    branches: [ master, develop, test-pr ]
+    branches: [ master, develop, test-pr, feat/simplifyci ]
     types: [opened, synchronize, reopened, labeled]
 
 concurrency: 
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -8,7 +8,7 @@ on:
       - develop
       - 'run-ci/**'
   pull_request:
-    branches: [ master, develop, test-pr, feat/simplifyci ]
+    branches: [ master, develop, test-pr ]
     types: [opened, synchronize, reopened, labeled]
 
 concurrency: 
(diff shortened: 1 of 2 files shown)


.github/workflows/node.js.yml
```diff
@@ -7,8 +7,8 @@ on:
       - master
       - develop
       - 'run-ci/**'
-  pull_request_target:
-    branches: [ master, develop, test-pr ]
+  pull_request:
+    branches: [ master, develop, test-pr, feat/simplifyci ]
```
dontrolle: you want to keep the `feat/simplifyci` branch mentioned here for a reason?

.github/workflows/node.js.yml
```diff
@@ -86,61 +42,40 @@ jobs:
...
-    # == Git checkout ==
     - name: Checkout
       uses: actions/checkout@v3
-      # Workaround for https://github.com/npm/cli/issues/2610
       with:
-        persist-credentials: false
-        ref: ${{ env.GIT_REF_TO_TEST }}
         submodules: recursive
 
-    - name: Reconfigure git to use HTTP authentication
```
dontrolle, line 104: This is not an issue any longer?

.github/workflows/node.js.yml
```diff
@@ -179,25 +114,13 @@ jobs:
     # == check precommit works ==
     - run: yarn run precommit
 
-    - name: Report Failure to Slack
```
dontrolle, line 182: Hmm, some might protest removing the Slack notifications. I don't think they're a perfect solution, but I agree we need to have some kind of monitoring and alerts on broken builds.

lnist: We have already removed them in most other repos. Its replaced by other monitoring. This is a leftover

.github/workflows/node.js.yml
lnist: The linked issue mentions npm, and I don't see any issues with building, so I don't think so? If you know of a concrete issue that does not work in the pipeline even though its green then I'm all ears :D

.github/workflows/node.js.yml
lnist: No, good catch :) That was for testing

test/script/toy/MangroveJs.t.sol
```diff
@@ -8,6 +8,17 @@ import {MangroveTest} from "mgv_test/lib/MangroveTest.sol";
 import "forge-std/console.sol";
 
 contract MangroveJsDeployTest is MangroveTest {
```
dontrolle: Was this supposed to be committed?

.github/workflows/node.js.yml
dontrolle: I don't recall any particular issues about this. But just to confirm - maybe @espendk does?

espendk: The build was broken due to some dependencies that didn't work in GitHub Actions. If the build works without that workaround I'm all for removing it.

test/script/toy/MangroveJs.t.sol
lnist: No :)

dontrolle (approved):
🌴

PR #375, opened 2023-06-16
deploying Kandel on Px Markets

jkrivine:
This branch contains the address of the kandel that was deployed on Px token markets on Polygon mainnet


espendk:
@jkrivine is this still relevant and should be merged?

PR #380, opened 2023-06-26
feat: Mangrove v2

adhusson:
# How to contribute
- current stage: ~design stage~ ~code structure stage~ documentation stage.
- Make a PR containing your change or feature, even in a rough state, with this branch (`feat/ticks`) as target. Your PR will be used for discussion/referencing code, etc.

# TODO after code freeze
- [x] Fix graphics shown in MgvDoc.html
   - contact map is obsolete for instance. When we update it, we should migrate it to a format that anyone can update (eg DrawIO); Currently, it's an Omnigraffle file that @jkrivine has the sources for.
   - Updated here: https://github.com/mangrovedao/mangrove-core/pull/648
- [ ] Address https://github.com/mangrovedao/mangrove-core/pull/380#discussion_r1342628281
	- This is a [solcco](https://github.com/adhusson/solcco) issue, hoping that @adhusson can fix 🤞
- [ ] 2nd round review of MgvDoc
  - Espen got to `Constants.sol` ...
  - We should review the occurrences of ", and" as some have been observed in the past to be inappropriate (maybe a "use Oxford comma everywhere" search-replace gone wrong?).

# Contents of this PR

## Ticks
Prices are represented as ticks. Ticks represent powers of 1.0001. Offers have their outgoing volume (`gives`) and their `tick` stored. `wants` is derived from `gives`&`tick`.

Tick max/min value is +/- 1048576, so the max/min price is about `3.4 * 10^45`.

The offer are stored inside the leaves of a "tick tree". The nodes of each level but the last are bitmaps with size equal to the number of children of a node. For instance the root of the tree looks like this:

```
                                ┌──────┐
                                │      │
        is 0 because all bits   │ 0  1 │  is 1 because at least 1 bit
        of the left child are 0 │      │  of the right child is 1
                                └┬────┬┘
                                 │    │
                                 │    │
┌─────────────────────────────┐  │    │  ┌─────────────────────────────┐
│                             │  │    │  │                             │
│ 0 0 0 0 0 0 ... 0 0 0 0 0 0 ┼──┘    └──┤ 0 1 0 1 1 0 ... 0 0 0 0 1 1 │
│                             │          │                             │
└─────────────────────────────┘          └─────────────────────────────┘
```

Leaves of the tree contain 4 pairs of offer IDs. Each pair represents the ID of the first and last offer of a tick (a price point). If a tick is empty, the pair is (0,0).

Bitmaps are defined like this:
* If a tick has at least one offer, and its leaf is the `N`th child of a node, that node has a `1` in its `N`th position (and a 0 otherwise).
* If a node has at least a 1, and is the `N`th child of another node, that node has a `1` in its `N`th position (and a 0 otherwise).

A _tree branch_ is the level3, the current level2, the current level1, the current level0, and the position in the current leaf of the current best offer. With these data points one can always compute the current tick.

## Dynamic ticks
To accomodate higher price volatility, each offer list is now parameterized by a `tickScale` (in addition to `outbound_tkn` and `inbound_tkn`). All prices are computed as `1.0001^(tick * tickScale)`. See #444 for more.

## New density
Density used to be on 112 bits, now it's on 9 bits. It is stored as a float with 2 bits for the mantissa, 7 for the exponent, and an exponent bias of 32. To play around with densities and challenge this format for the density, you can [use this Observable notebook](https://observablehq.com/d/44bda3a2282aa521).

For convenience, the density is given to mangrove as a 96.32 fixed-point number (it matches the range of the floating-point format and is just the right size to prevent overflows when computing amounts). There is a `DensityLib` for format conversion and for deducing densities from token&gas prices.

## New offer_gasbase
Gasbase used to be on 24 bits, now it's on 9 bits. It is stored in 1k gas units and has been renamd `kilo_offer_gasbase`. There are helpers functions (getter/setter) named `offer_gasbase` that take care of conversion. Numbers are rounded down.

## Replace `snipes` with `cleanByImpersonation` function
NB: This is not tick related, but will be included in the same release and has therefore been made on this PR.

`snipes` and `snipesFor` have been removed from Mangrove and replaced with a `cleanByImpersonation` function that only allows failing offers to be executed. This replaces the functionality of `MgvReader` which has been removed.

###  Cleaning of multiple offers is more resilient
Before, `MgvCleaner.collect*` would revert if execution of any offer succeeded. The new `cleanByImpersonation` is more resilient as it will not revert the whole tx if an offer succeeds but will revert just the execution of that offer and continue cleaning of the next offers. 

Thus a cleaner that attempts to clean multiple offers in one tx will be able to successfully clean some offers despite some offers failing to be cleaned.

### Taker impersonation no longer checks permits
The new `cleanByImpersonation` function has integrated the `taker` parameter of the old `snipesFor`. This allows cleaning by impersonation, ie where another taker's funds and approval of Mangrove for transfers of quote can be used to execute an offer and make it fail (which was already possible via `MgvCleaner.collectByImpersonation`).

This has been integrated into the `cleanByImpersonation` function instead of having a separate `cleanFor` function because
1. no permit/allowance is needed nor does it have to be checked and
2. cleaners are expected to be advanced users, so requiring an explicit `taker` parameter with non-obvious semantics seems ok.

## Prevent makers from reading offer list during `makerExecute`
Makers can no longer read the offer list during `makerExecute` for the following reasons:

1. To prevent `makerExecute` from detecting cleaning
2. To avoid the gas cost of keeping the offer list up-to-date for `makerExecute`. A number of optimizations are in place which mean that `local`, leafs, and levels may not be up-to-date.

Two measures have been added to achieve this:
- The existing offer list lock is now enforced as a read-lock as well in all the view functions.
- All fields in the `SingleOrder` struct that pertain to the offer list/tick tree structure are cleared (set to zero) before it is passed to `makerExecute`.

### Completed TODOs
- [x] Make clz/fls specialized for smaller sizes (depends on the sizes of level0, level1, etc)
- (would work for original mangrove too): use packed offers in-transit (ofp etc)
  - not interesting
- [x] simplify densityFixed usage (no more `    uint densityFixed = (10 ** 7) << DensityLib.FIXED_FRACTIONAL_BITS;`) 
  - #559

### Questions
- [x] Design questions
	- [x] Is the ~500k MAX_TICK enough? (means prices go up to `5.8 * 10^22`). It could be made a lot higher at the cost of more complex code, should check later.
		- We've decided to increase the range of ticks so that the full supported logPrice range fits in the tree
		- See discussion on Slack: https://mangroveitops.slack.com/archives/C05LDSH7FHD/p1694523876856419
	- [x] How to implement dynamic ticks? Current plan: have one offerlist per pair,ticksize but have a canonical ticksize for each market
		- `tickScale` parameter introduced: https://github.com/mangrovedao/mangrove-core/pull/444
	- [x] What is the lowest tick? 1bp is common on uniswap but lower may be good (has tick range tradeoffs)
		- We've decided to stick with 1 (relative) bps for now, see discussion on Slack: https://mangroveitops.slack.com/archives/C05LDSH7FHD/p1691670160846889
	- [x] Should market orders be « average limit » or « max limit »?
	  * Answered: we implemented "max limit"
	- [x] Do we need a way to avoid stackoverflow for market orders? Especially with only snipe-for-fail?
		- Implemented in https://github.com/mangrovedao/mangrove-core/pull/461
- [x] Optimizations
	- Optimize away rereads of `ofp.local` in `writeOffer`
	  - Not useful
	- Construct a new `local` right away instead of repeatedly updating `sor.local` at the end of market order
	  - not worth it
	- [x] Should fields&leafs be "touched", ie with a bit set to 1 (which is ignored for all other purposes), in order to save gas? Otherwise fields/leafs going from empty to nonempty cost ~22k gas (amortized cost is not better). 
		- Implemented in https://github.com/mangrovedao/mangrove-core/pull/530
	- Do offers need to contain their logPrice? They location in the tree reveals the price anyway (similarly the tokens of an offer are given by where the offer is stored).
	  - They do, you can access an offer without knowing its price (while you can't access an offer without knowing its tokens)
- [x] Code improvements
	- [x] Choose between a reverting log2 function (fewer surprises) and a non-reverting one (less code and less gas)
	  - Using a non-reverting one
	- Should ticks be stored as uints to simplify tick tree-related operations? avoids level2 shift headache at least
		- Issue created here: https://github.com/mangrovedao/mangrove-core/issues/511
	- Reorg structs for readability (multiple files)?
		- Issue created here: https://github.com/mangrovedao/mangrove-core/issues/475
- [x] Is a tick offset/shift needed to bring the initial market mid price close to tick 0?
	- Could allow us to use a smaller relative tick size, eg 0.1 bps	
	- We decided not to do this, but to expand the tick tree instead
	- There's a POC PR on this here: https://github.com/mangrovedao/mangrove-core/pull/532
- Either reduce the tree height, or find how to traverse it less. e.g.: only go up to the level needed for displacing an offer, keep part of the tree dirty when running a market order, etc., is that possible?
  - WONTFIX for this version
- [x] Should we remove extra methods `ByVolume` `ByPrice` and replace them either with a proxy or with a `using * for *` library that adds methods to Mangrove, or with something else that doesn't impact the UX
  - `ByPrice` has been removed, `ByVolume` has been kept

  
### Main change
- Add `TickLib.sol` for ticks
- Add Dynamic Ticks (see #444 for details)
- add `[prev,next]OfferId(offer)` to MgvReader
- remove `pivotId` everywhere
- density is now a 9 bit float
- gasbase is now kilo_gasbase and on 10 bits
- there is no `local.best` nor `local.tick`

### Small fixes
- `density` error message uses additional custom code (not hardcoded anymore)

### File-level changes
- remove most deployers, strats, & MangroveJs (will adapt to ticks later)
- move `TransferLib` to lib/
- update forge-std (more pure functions)
- move to solidity 0.8.18

### Tooling
- add `offer.isLive()` (remove `mgv.isLive(offer)`)
- Add `toString` library & preprocessing utilities
- preprocessing
	- add `toString` functions
	- allow fields with `int` base type
	- allow fields with user-defined types
	- allow additional custom code for preprocessed structs
	- add struct field getters in js code

### Testing
- increase ratios between amounts in some tests due to loss of precision caused by ticks
- add "relative error" (`relError`) function (in basis points) for easier approximate equality tests
- add `Pair` struct & user-defined functions for easier testing (e.g. `pair.offers(...)`, `pair.nextOfferId(offer)`, ...)
- add `Debug.sol` all-purpose debugging file

# TODOs once mangrove-core solution has stabilized
Once the Solidity solution for the core has stabilized, other components must be updated to match the new API:
- [ ] Deployment
- [ ] Release
- [ ]Strat lib
  - Should be audited
- [ ] MangroveOrder
  - Should be audited
- [ ] Kandel
  - Should be audited
- [ ] mangrove.js
- [ ] web app
- [ ] docs


s4nxbt:
With Vincent and Hamza, we are starting to review this PR - "Ticks represent powers of 1.0001."
Are 1bps ticks the small size for every market ? For stable / stable we found that it would be better if it was 1/10 bps.

adhusson:
> we are starting to review this PR

awesome!

> Are 1bps ticks the small size for every market ?

No, there will be multiple tick sizes (cf TODO section)

espen:
Wrong @espen

espendk:
> Wrong @espen

whoops, sorry! I'll fix it

preprocessing/structs.ts
```diff
@@ -1,23 +1,88 @@
...
+
+   The sequence diagram of a market order is as follows:
+   <img src="./sequenceChart.png" width="190%"></img>
+
+
+   ## `tickSpacing`
+
+   The granularity of available price points in an offer list is controlled by the `tickSpacing` parameter. With a high `tickSpacing`, fewer price points are available, but the gas cost of market orders is lower since a smaller part of the tick tree has to be explored.
+
+   The available prices in an offer list are `1.0001^i` for all `MIN_TICK <= i <= MAX_TICK` such that `i % tickSpacing = 0`.
```
espendk, line 33: Perhaps we should use "ratio" instead of "price"? That would be more consistent with the code.
If we do, we should probably explain why we do this and what the relationship between "ratio" and "price" is.

NB: I think it's ok that we still use "price points" when explaining "ticks" at the more abstract level.

NNB: "price" is used many times all over 😬 I'll only mention it here.

preprocessing/structs.ts
```diff
@@ -96,121 +161,159 @@ const id_field = (name: string) => {
...
-     */
-    { name: "lock", bits: 1, type: "bool" },
-    /* * `best` holds the current best offer id. Has size of an id field. *Danger*: reading best inside a lock may give you a stale value. */
-    id_field("best"),
-    /* * `last` is a counter for offer ids, incremented every time a new offer is created. It can't go above $2^{32}-1$. */
-    id_field("last"),
-  ],
+      fields.gasreq,
+      /*
+        * <a id="structs.js/gasbase"></a>  `offer_gasbase` represents the gas overhead used by processing the offer inside Mangrove + the overhead of initiating an entire order, in 1k gas increments.
```
```suggestion
        * <a id="structs.js/gasbase"></a>  `kilo_offer_gasbase` represents the gas overhead used by processing the offer inside Mangrove + the overhead of initiating an entire order, in 1k gas increments.
```

preprocessing/structs.ts
```diff
@@ -96,121 +161,159 @@ const id_field = (name: string) => {
...
+
+      The gas considered 'used' by an offer is the sum of
+      * gas consumed during the call to the offer
+      * `offer_gasbase`
+      
+    (There is an inefficiency here. The overhead could be split into an "offer-local overhead" and a "general overhead". That general overhead gas penalty could be spread between all offers executed during an order, or all failing offers. It would still be possible for a cleaner to execute a failing offer alone and make them pay the entire general gas overhead. For the sake of simplicity we keep only one "offer overhead" value.)
+
+    If an offer fails, `gasprice` mwei is taken from the
```
espendk: I just realized "m" is the SI symbol for "milli" and we mean "mega" ("M"), so should be "Mwei".

preprocessing/structs.ts
```diff
@@ -96,121 +161,159 @@ const id_field = (name: string) => {
...
+    provision per unit of gas used. `gasprice` should approximate the average gas
+    ratio at offer creation time.
```
```suggestion
    provision per unit of gas used. `gasprice` should approximate the average gas
    price at offer creation time.
```

preprocessing/structs.ts
```diff
@@ -96,121 +161,159 @@ const id_field = (name: string) => {
...
+
+    `kilo_offer_gasbase` is the actual field name, and is _9 bits wide_ and represents 1k gas increments. The accessor `offer_gasbase` returns `kilo_offer_gasbase * 1e3`.
+
+    `kilo_offer_gasbase` is also the name of a local Mangrove
+    parameters. When an offer is created, their current value is copied from Mangrove local configuration. The maker does not choose it.
```
```suggestion
    `kilo_offer_gasbase` is also the name of a local Mangrove
    parameter. When an offer is created, its current value is copied from Mangrove local configuration. The maker does not choose it.
```

preprocessing/structs.ts
espendk: I just saw this further down:
> `kilo_offer_gasbase` is the actual field name

So if it's intentional to state the field name as `offer_gasbase` I guess that's ok. But then the last part of the description ("in 1k gas increments")  should probably be remove.

preprocessing/structs.ts
```diff
@@ -96,121 +161,159 @@ const id_field = (name: string) => {
...
+      When an offer fails, the following amount is given to the taker as compensation:
+    ```
+    (gasused + offer_gasbase) * gasprice * 1e6
+    ```
+
+    where `offer_gasbase` and `gasprice` are Mangrove's current configuration values.  The rest is given back to the maker.
+
+      */
+      fields.kilo_offer_gasbase,
+      /* * `gasprice` is in mwei/gas and _26 bits wide_, which accommodates 0.001 to ~67k gwei / gas.  `gasprice` is also the name of a global Mangrove parameter. When an offer is created, the offer's `gasprice` is set to the max of the user-specified `gasprice` and Mangrove's global `gasprice`. */
```
espendk: hmm, unit prefixes (m and g) are a bit confusing here, since they're used in the SI sense. But "gwei" seems often used instead of "Gwei" (though that's also in use according to google).

I'd prefer to stick to SI prefixes to avoid confusion.

```suggestion
      /* * `gasprice` is in Mwei/gas and _26 bits wide_, which accommodates 0.001 to ~67k Gwei / gas.  `gasprice` is also the name of a global Mangrove parameter. When an offer is created, the offer's `gasprice` is set to the max of the user-specified `gasprice` and Mangrove's global `gasprice`. */
```

preprocessing/structs.ts
```diff
@@ -96,121 +161,159 @@ const id_field = (name: string) => {
...
+using OfferDetailUnpackedExtra for OfferDetailUnpacked global;
+`,
+  },
+
+  /* ## Global Configuration
+   Configuration information for an offer list is split between a `global` struct (common to all offer lists) and a `local` struct specific to each offer list. Global configuration fields are:
+   */
+  global: {
+    fields: [
+      /* * The `monitor` can provide realtime values for `gasprice` and `density` to Mangrove. It can also receive liquidity events notifications. */
```
```suggestion
      /* * The `monitor` can provide real-time values for `gasprice` and `density` to Mangrove. It can also receive liquidity event notifications. */
```

preprocessing/structs.ts
```diff
@@ -96,121 +161,159 @@ const id_field = (name: string) => {
...
+      { name: "monitor", bits: 160, type: "address" },
+      /* * If `useOracle` is true, the dex will use the monitor address as an oracle for `gasprice` and `density`, for every outbound_tkn/inbound_tkn pair, except if the oracle-provided values do not pass a check performed by Mangrove. In that case the oracle values are ignored. */
```
```suggestion
      /* * If `useOracle` is true, the DEX will use the monitor address as an oracle for `gasprice` and `density`, for every `outbound_tkn`/`inbound_tkn` pair, except if the oracle-provided values do not pass a check performed by Mangrove. In that case the oracle values are ignored. */
```

espendk: Generally, "dex" -> "DEX".

preprocessing/structs.ts
```diff
@@ -96,121 +161,159 @@ const id_field = (name: string) => {
...
+      { name: "useOracle", bits: 1, type: "bool" },
+      /* * If `notify` is true, the dex will notify the monitor address after every offer execution. */
+      { name: "notify", bits: 1, type: "bool" },
+      /* * The `gasprice` is the amount of penalty paid by failed offers, in mwei per gas used. `gasprice` should approximate the average gas price and will be subject to regular updates. */
+      fields.gasprice,
+      /* * `gasmax` specifies how much gas an offer may ask for at execution time. An offer which asks for more gas than the block limit would live forever on the book. Nobody could take it or remove it, except its creator (who could cancel it). In practice, we will set this parameter to a reasonable limit taking into account both practical transaction sizes and the complexity of maker contracts.
+      */
+      { name: "gasmax", bits: 24, type: "uint" },
+      /* * `dead`: if necessary, Mangrove can be entirely deactivated by governance (offers can still be withdrawn and bounties can still be withdrawn). Once killed, Mangrove must be redeployed. It cannot be resurrected. */
```
```suggestion
      /* * `dead`: if necessary, Mangrove can be entirely deactivated by governance (offers can still be retracted and provisions can still be withdrawn). Once killed, Mangrove must be redeployed; It cannot be resurrected. */
```

preprocessing/structs.ts
```diff
@@ -96,121 +161,159 @@ const id_field = (name: string) => {
...
+      { name: "dead", bits: 1, type: "bool" },
+      /* * `maxRecursionDepth` is the maximum number of times a market order can recursively execute offers. This is a protection against stack overflows. */
+      { name: "maxRecursionDepth", bits: 8, type: "uint" },      
+      /* * `maxGasreqForFailingOffers` is the maximum gasreq failing offers can consume in total. This is used in a protection against failing offers collectively consuming the block gas limit in a market order. Setting it too high would make it possible for successive failing offers to consume up to that limit then trigger a revert (thus the failing offer would not be removed). During a market order, Mangrove keeps a running sum of the `gasreq` of the failing offers it has executed, and stops the market order when that sum exceeds `maxGasreqForFailingOffers`. */
```
```suggestion
      /* * `maxGasreqForFailingOffers` is the maximum gasreq failing offers can consume in total. This is used in a protection against failing offers collectively consuming the block gas limit in a market order. Setting it too high would make it possible for successive failing offers to consume up to that limit then trigger a revert (thus the failing offer would not be removed). During a market order, Mangrove keeps a running sum of the `gasreq` of the failing offers it has executed and stops the market order when that sum exceeds `maxGasreqForFailingOffers`. */
```

preprocessing/structs.ts
```diff
@@ -96,121 +161,159 @@ const id_field = (name: string) => {
...
+      * gas consumed during the call to the offer
+      * `offer_gasbase`
+      
+    (There is an inefficiency here. The overhead could be split into an "offer-local overhead" and a "general overhead". That general overhead gas penalty could be spread between all offers executed during an order, or all failing offers. It would still be possible for a cleaner to execute a failing offer alone and make them pay the entire general gas overhead. For the sake of simplicity we keep only one "offer overhead" value.)
+
+    If an offer fails, `gasprice` mwei is taken from the
+    provision per unit of gas used. `gasprice` should approximate the average gas
+    ratio at offer creation time.
+
+    `kilo_offer_gasbase` is the actual field name, and is _9 bits wide_ and represents 1k gas increments. The accessor `offer_gasbase` returns `kilo_offer_gasbase * 1e3`.
```
```suggestion
    `kilo_offer_gasbase` is the actual field name, is _9 bits wide_, and represents 1k gas increments. The accessor `offer_gasbase` returns `kilo_offer_gasbase * 1e3`.
```

preprocessing/structs.ts
```diff
@@ -96,121 +161,159 @@ const id_field = (name: string) => {
...
-
-    When an offer fails, the following amount is given to the taker as compensation:
-   ```
-   (gasused + offer_gasbase) * gasprice
-   ```
-
-   where `offer_gasbase` and `gasprice` are Mangrove's current configuration values.  The rest is given back to the maker.
+  offerDetail: {
+    fields: [
+      /* * `maker` is the address that created the offer. It will be called when the offer is executed, and later during the posthook phase. */
```
```suggestion
      /* * `maker` is the address that created the offer. It will be called when the offer is executed and later during the posthook phase. */
```

I think at some point a "we should should Oxford comma" search-replace went wrong - there are a number of ", and" occurences in non-listing contexts.

preprocessing/structs.ts
```diff
@@ -96,121 +161,159 @@ const id_field = (name: string) => {
...
+      { name: "active", bits: 1, type: "bool" },
+      /* * `fee`, in basis points, of `outbound_tkn` given to the taker. This fee is sent to Mangrove. Fee is capped to ~2.5%. */
+      { name: "fee", bits: 8, type: "uint" },
+      /* * `density` is similar to a 'dust' parameter. We prevent spamming of low-volume offers by asking for a minimum 'density' in `outbound_tkn` per gas requested. For instance, if `density` is worth 10, `offer_gasbase == 5000`, an offer with `gasreq == 30000` must promise at least _10 × (30000 + 5000) = 350000_ `outbound_tkn`. _9 bits wide_.
+
+      We store the density as a float with 2 bits for the mantissa, 7 for the exponent, and an exponent bias of 32, so that density ranges from $2^{-32}$ to $1.75 \times 2^{95}$. For more information, see `DensityLib`.
+      
+      */
+      { name: "density", bits: 9, type: "Density", underlyingType: "uint"},
+      /* To save on gas, Mangrove cache the entire tick tree branch of the bin that contains the best offer in each offer list's `local` parameter. Taken together, `binPosInLeaf`, `level3`, `level2`, `level1` and `root` provide the following info:
```
```suggestion
      /* To save gas, Mangrove caches the entire tick tree branch of the bin that contains the best offer in each offer list's `local` parameter. Taken together, `binPosInLeaf`, `level3`, `level2`, `level1`, and `root` provide the following info:
```

preprocessing/structs.ts
```diff
@@ -96,121 +161,159 @@ const id_field = (name: string) => {
...
+      - What the current bin is (see `BinLib.bestBinFromLocal`)
+      - When a leaf is emptied and the next offer must be fetched, the information in the fields `level3`, `level2`, `level1` and `root` avoid multiple storage reads
```
espendk, line 287: Spacing is weird here in the generated docs:

billede

preprocessing/structs.ts
```diff
@@ -96,121 +161,159 @@ const id_field = (name: string) => {
...
+      /* * `fee`, in basis points, of `outbound_tkn` given to the taker. This fee is sent to Mangrove. Fee is capped to ~2.5%. */
+      { name: "fee", bits: 8, type: "uint" },
+      /* * `density` is similar to a 'dust' parameter. We prevent spamming of low-volume offers by asking for a minimum 'density' in `outbound_tkn` per gas requested. For instance, if `density` is worth 10, `offer_gasbase == 5000`, an offer with `gasreq == 30000` must promise at least _10 × (30000 + 5000) = 350000_ `outbound_tkn`. _9 bits wide_.
+
+      We store the density as a float with 2 bits for the mantissa, 7 for the exponent, and an exponent bias of 32, so that density ranges from $2^{-32}$ to $1.75 \times 2^{95}$. For more information, see `DensityLib`.
+      
+      */
+      { name: "density", bits: 9, type: "Density", underlyingType: "uint"},
+      /* To save on gas, Mangrove cache the entire tick tree branch of the bin that contains the best offer in each offer list's `local` parameter. Taken together, `binPosInLeaf`, `level3`, `level2`, `level1` and `root` provide the following info:
+      - What the current bin is (see `BinLib.bestBinFromLocal`)
```
espendk, line 286:
```suggestion
      - What the current bin is (see `BinLib.bestBinFromLocal`).
```

preprocessing/structs.ts
espendk, line 287:
```suggestion
      - When a leaf is emptied and the next offer must be fetched, the information in the fields `level3`, `level2`, `level1` and `root` avoid multiple storage reads.
```

espendk: Ah, just realized that the first two bullets pertain to a different set of fields than the third.
Not that important I guess.

preprocessing/structs.ts
```diff
@@ -96,121 +161,159 @@ const id_field = (name: string) => {
...
+      { name: "level2", bits: 64, type: "Field", underlyingType: "uint" },
+      { name: "level1", bits: 64, type: "Field", underlyingType: "uint" },
+      { name: "root", bits: 2, type: "Field", underlyingType: "uint" },
+      /* * `offer_gasbase` represents the gas overhead used by processing the offer inside Mangrove + the overhead of initiating an entire order. Mangrove considers that a failed offer has used at least `offer_gasbase` gas. The actual field name is `kilo_offer_gasbase` and the accessor `offer_gasbase` returns `kilo_offer_gasbase*1e3`. Local to an offer list, because the costs of calling `outbound_tkn` and `inbound_tkn`'s `transferFrom` are part of `offer_gasbase`. Should only be updated when ERC20 contracts change or when opcode prices change. */
+      fields.kilo_offer_gasbase,
+      /* * If `lock` is true, orders may not be added nor executed, nor the offer list read by external contracts.
+
+        Reentrancy during offer execution is not considered safe:
+      * during execution, an offer could consume other offers further up in the list, effectively frontrunning the taker currently executing the offer.
```
```suggestion
      * during execution, an offer could consume other offers further up in the list, effectively front-running the taker currently executing the offer.
```

preprocessing/structs.ts
```diff
@@ -96,121 +161,159 @@ const id_field = (name: string) => {
...
+      * it could also cancel other offers, creating a discrepancy between the advertised and actual market price at no cost to the maker.
+      * an offer insertion consumes an unbounded amount of gas (because it has to be correctly placed in the book).
```
espendk: This is no longer an issue, right?

preprocessing/structs.ts
```diff
@@ -96,121 +161,159 @@ const id_field = (name: string) => {
...
+      { name: "binPosInLeaf", bits: 2, type: "uint" },
+      { name: "level3", bits: 64, type: "Field", underlyingType: "uint" },
+      { name: "level2", bits: 64, type: "Field", underlyingType: "uint" },
+      { name: "level1", bits: 64, type: "Field", underlyingType: "uint" },
+      { name: "root", bits: 2, type: "Field", underlyingType: "uint" },
+      /* * `offer_gasbase` represents the gas overhead used by processing the offer inside Mangrove + the overhead of initiating an entire order. Mangrove considers that a failed offer has used at least `offer_gasbase` gas. The actual field name is `kilo_offer_gasbase` and the accessor `offer_gasbase` returns `kilo_offer_gasbase*1e3`. Local to an offer list, because the costs of calling `outbound_tkn` and `inbound_tkn`'s `transferFrom` are part of `offer_gasbase`. Should only be updated when ERC20 contracts change or when opcode prices change. */
+      fields.kilo_offer_gasbase,
+      /* * If `lock` is true, orders may not be added nor executed, nor the offer list read by external contracts.
+
+        Reentrancy during offer execution is not considered safe:
```
espendk, line 298: Maybe we should also mention, that it may be possible to detect cleaning.

preprocessing/structs.ts
```diff
@@ -96,121 +161,159 @@ const id_field = (name: string) => {
...
+      * during execution, an offer could consume other offers further up in the list, effectively frontrunning the taker currently executing the offer.
+      * it could also cancel other offers, creating a discrepancy between the advertised and actual market price at no cost to the maker.
+      * an offer insertion consumes an unbounded amount of gas (because it has to be correctly placed in the book).
+
+  Note: An optimization in the `marketOrder` function relies on reentrancy being forbidden.
+      */
+      { name: "lock", bits: 1, type: "bool" },
+      /* * `best` holds the current best offer id. Has size of an id field. *Danger*: reading best inside a lock may give you a stale value. */
+      // id_field("best"),
```
espendk: Formatting is off in the generated html:

billede

preprocessing/structs.ts
```diff
@@ -96,121 +161,159 @@ const id_field = (name: string) => {
...
+    ],
+    /* Import additional libraries for `Local` and `LocalExtra`. */
+    additionalDefinitions: (struct) => `import {Density, DensityLib} from "mgv_lib/DensityLib.sol";
+import {Bin,BinLib,Field} from "mgv_lib/BinLib.sol";
+/* Globally enable global.method(...) */
+import {LocalExtra,LocalUnpackedExtra} from "mgv_lib/LocalExtra.sol";
+using LocalExtra for Local global;
+using LocalUnpackedExtra for LocalUnpacked global;
+`,
+  }
```
espendk: This is included in the generated html and looks weird:

billede

espendk (commented):
Checkpoint after reviewing structs.ts

lib/Constants.sol
```diff
@@ -0,0 +1,88 @@
...
+uint constant TOPBIT = 0x8000000000000000000000000000000000000000000000000000000000000000;
+uint constant NOT_TOPBIT = 0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff;
+
+/* **sizes must match field sizes in `structs.ts` where relevant** */
+
+/* Number of bits used to represent a tick */
+uint constant TICK_BITS = 21;
+/* Number of bits used to represent an offer id */
+uint constant OFFER_BITS = 32;
+/* Maximum possible size of a field -- protects against wrong code changes. Constraint given by BitLib.ctz64 */
```
espendk:
```suggestion
/* Maximum possible size of a field -- protects against wrong code changes. Constraint given by `BitLib.ctz64` */
```

preprocessing/structs.ts
adhusson: Yeah having ratio everywhere read made it hard to read, I'll just add a comment that relates the two (edit especially because there are other ratios in the code, since it's such a generic word, so just saying the ratio would be confusing on top of being distracting)

adhusson: f0685d78b90166d1ac483051c112be75d4378626

preprocessing/structs.ts
adhusson: wait why remove "in 1k gas increments"?

preprocessing/structs.ts
adhusson: ooooh damn good catch.. yeah for gwei it works because there's no possible confusion?

adhusson: e3e70f9e451b8912157f8e89a4e8e42a258885c5

preprocessing/structs.ts
adhusson: wdym?

preprocessing/structs.ts
adhusson: the entire thing should be removed

preprocessing/structs.ts
espendk: Ok. I think I saw that we use "ratio" in some places instead of "price", we should check that and make it consistent.

preprocessing/structs.ts
adhusson: it is supposed to be included, by weird do you mean it's syntax-highlighted when it shouldn't? If yes I would leave it as-is, it's an issue with the highlighter solcco uses

preprocessing/structs.ts
espendk: hmm maybe that can be retained. But I must admit that I find the mixing of `kilo_offer_gasbase` and `offer_gasbase` in these comments confusing; Would be nice to only use one term for this.

preprocessing/structs.ts
adhusson: will change to Mangrove

preprocessing/structs.ts
espendk: I think "gwei" works because it's a commonly used way of writing it. But I don't like it ;-)

espendk: > [e3e70f9](https://github.com/mangrovedao/mangrove-core/commit/e3e70f9e451b8912157f8e89a4e8e42a258885c5)

That also added DensityLib at the new location 😬

preprocessing/structs.ts
espendk: We should explain that if we didn't have the read-lock it might be possible for malicious makers to detect cleaning and thus avoid it.

preprocessing/structs.ts
espendk: I mean the comments on the left-hand side that it causes look weird and shouldn't be there IMO.

espendk (commented):
Review checkpoint to get my responses to @adhusson's published (reviewing while discussing previous comments doesn't seem to be well-supported)

preprocessing/structs.ts
adhusson: done in 4a009d046831f675e34e1506f69b428656097842

preprocessing/structs.ts
adhusson: I think with this c4d8864019af90153548254b93ebf4bbedde7c33 it will be better

preprocessing/structs.ts
adhusson: oh damn! but it got deleted in cbbc3af7d92f241b7960a843fa71e4a4ebb262f0

preprocessing/structs.ts
espendk: Just had a look and I think this "price in comments, ratio in code" is super confusing :-/

preprocessing/structs.ts
adhusson: 001eed1d07d2d87eaea79dc0fada7ee5aa3da451

preprocessing/structs.ts
adhusson: ahhh yes I see it now! this shoudl be fixed in solcco, it's a flaw in its string-detection code. I'm thinking of leaving as-is for now, wdyt?

espendk: Sure, let's add it to the post freeze backlog :-)

lnist (approved):

preprocessing/structs.ts
espendk: @adhusson is this something you can fix? Or should we ask the solcco maintainer? oh, wait... 😉

PR #395, opened 2023-07-04
Feat/market health scripts

jkrivine:
This PR introduces the `/script/core/monitors/` folder that contains scripts to introspect the state of opened markets on Mangrove. The first of these scripts is `marketHealth` which performs snipes to emulate a market order and evaluates required gas costs and how many offer would be failing.


script/core/monitors/MarketHealth.s.sol
```diff
@@ -0,0 +1,156 @@
...
+
+import {Deployer} from "mgv_script/lib/Deployer.sol";
+import {Test2, console2 as console} from "mgv_lib/Test2.sol";
+import {UpdateMarket} from "mgv_script/periphery/UpdateMarket.s.sol";
+import {MgvReader, VolumeData, IMangrove} from "mgv_src/periphery/MgvReader.sol";
+import {IERC20} from "mgv_src/IERC20.sol";
+import {MgvStructs} from "mgv_src/MgvLib.sol";
+
+/**
+ * @notice Script to obtain data about a given mangrove half book. Data is outputted to terminal as space separated values.
```
adhusson: I think "offer list" has become the canonical term

script/core/monitors/MarketHealth.s.sol
```diff
@@ -0,0 +1,156 @@
+// SPDX-License-Identifier:	AGPL-3.0
```
adhusson, line 1: Maybe `script/periphery`? After the repo split they will probably not be in mangrove-core

script/core/monitors/MarketHealth.s.sol
```diff
@@ -0,0 +1,156 @@
...
+
+/**
+ * @notice Script to obtain data about a given mangrove half book. Data is outputted to terminal as space separated values.
+ * @notice retrieved data are:
+ *  "tot_bounty": collected bounty
+ *  "tot_failures": total failures
+ *  "tot_gas_fail": gas consumed by failing offer
+ *  "tot_gas_req": gas required by makers
+ *  "tot_gas_used": gas consumed for the market order
+ *  "tot_successes": number of successful offers
```
adhusson: I would remove the `tot_` prefix

script/core/monitors/MarketHealth.s.sol
```diff
@@ -0,0 +1,156 @@
...
+ *  "volume_sent": volume sent (in display units of inbound token)
+ *
+ * @notice environment variables that control the script are:
+ * @param VOLUME the amount of outbound token that are required from the market
+ * @param TKN_IN the inbound token (token that are sent by the tester)
+ * @param TKN_OUT the outbound token (token that are required by the tester)
+ */
+/**
+ * Usage: testing status of buy orders on the WMATIC,USDT market for volumes of up to 100,000 WMATIC (display units)
+ *  VOLUME=$(cast ff 18 100000) TKN_IN=USDT TKN_OUT=WMATIC \
```
adhusson, line 43: Maybe easier to let the user provide `VOLUME` in display units and then manually multiply by `decimals`

script/core/monitors/MarketHealth.s.sol
```diff
@@ -0,0 +1,156 @@
...
+contract MarketHealth is Test2, Deployer {
+  // needed if some offer fail, in order to receive bounty
+  uint[] internal failingIds;
+
+  receive() external payable {}
+
+  function run() public {
+    IERC20 inbTkn = IERC20(envAddressOrName("TKN_IN"));
+    // dealing hopefully enough inbound token to execute a market order
+    deal(address(inbTkn), address(this), 10_000_000 * 10 ** inbTkn.decimals());
```
adhusson, line 55: Would `type(uint).max` work?

foundry.toml
```diff
@@ -7,7 +7,7 @@ test='test'
 out='out'
 libs=['lib']
 cache_path='cache'
-fs_permissions = [{ access = "read-write", path = "./addresses/"}, { access = "read", path = "./out/" }]
+fs_permissions = [{ access = "read-write", path = "./addresses/"}, { access = "read-write", path = "./script/core/monitors/"}, { access = "read", path = "./out/" }]
```
adhusson: A directory outside of the source subtrees would be nice, in particular because those directories are not ignored by git. Maybe `./monitor_results` and add `monitor_results` to `.gitignore`?

script/core/monitors/MarketHealth.s.sol
adhusson: The doc could use a small example of what the script outputs -- at the moment I have trouble having a clear picture.

script/core/monitors/MarketHealth.s.sol
jkrivine: But the script is about testing the state of mangrove-core. So not sure about this point

script/core/monitors/MarketHealth.s.sol
jkrivine: How do you give a  `VOLUME`  of 1.3 BTC then?

script/core/monitors/MarketHealth.s.sol
jkrivine: Probably would as well but since I have the required value I might as well use it.

script/core/monitors/MarketHealth.s.sol
adhusson: yes makes sense!

adhusson (approved):

PR #399, opened 2023-07-06
Fix/ci foundry

Louis-Amas:


jkrivine:
Let me try to add the `analytics` folder to see if that solves the issue as well

PR #418, opened 2023-07-25
feat: read addresses from node_modules/mangrovedao/mangrove-core

peterMangrove:
At the moment we are only able to read addresses from root/addresses/(context/deployed), but when using the mangrove-core package, like in bot arbitrage, then the addresses are no longer in that path. So ATM we have just copied the addresses to the same path, so that it reads the addresses, but this is not very nice and we have to copy addresses very time new addresses gets released, and keeping track of “core” addresses vs your own addresses, can be a mess. Because of this, i’ve added this path root/node_modules/@mangrovedao/mangrove-core/addresses/(deployed/context) so that we can also read the addresses directly from the mangrove-core package. This makes it a lot easier to use the mangrove-core package and its addresses, with your own addresses.

### DIFF
--- a/script/lib/Deployer.sol
+++ b/script/lib/Deployer.sol
@@ -206,10 +206,10 @@ abstract contract Deployer is Script2 {
       line(end ? "  }" : "  },");
     }
     line("]");
-    string memory latestBackupFile = fork.addressesFile("deployed.backup", "-latest");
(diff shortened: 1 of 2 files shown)


adhusson (approved):
Nice! Do we have a forge issue for this?

peterMangrove:
> 

Not sure what this has to do with forge

PR #424, opened 2023-08-01
feat: add info to events

peterMangrove:
This PR will add information to current events or add new events, in order for indexing to be able to get the information it needs, without having to do complex calculations.

### The information needed are as follows:
- MarketOrder information.
  -  What parameters were used to start the MO
- Provision for an offer 
  - How much provision has a specific offer locked
- Add indexed on taker in OfferSuccess.
  - Makes it possible to filter on the taker address

### MarketOrder
This change is adding `outbound_tkn`,`inbound_tkn`, `takerWants`, `takerGives`, `fillWants` and `taker` to the event `OrderStart`. The reason for this change, is that when we index a normal Market Order, then we are not able to tell wether the order was fully filled or only partially filled. The information is needed in the UI, under Order history.

In order to do this, i've also added the event `SnipeStart`. This is because `Snipe` also used the event `OrderStart`, but when we do a snipe, then we do not have the same information available. We could have added more information to the `SnipeStart` event, but as we do currently not have any need for any information about snipes when we index, then this is not done, to save the gas.

Another option to this would be to try and read the transaction and all the method calls in that tx, in order to get the parameters used for the Market Order method. But this is a very complicated thing to do in the indexer and might not even be possible, if we have nested Market Orders in a single tx.

### Provision
This change is creating two new events, `Debit` and `Credit` with offerId. The two original events still exists and are emitted when a debit or credit happens, that is not in context of an offer. When a debit or credit happens in the context of an offer, e.g. offerWrite, offerFail, etc. Then we are now emitting an `Debit` or `Credit` with the offerId. This is done in order to keep track of the provision for each offer. This way we can e.g. show in the UI for Kandel how much each offer has locked in provision. 
There is one corner case were the credit event does not match the offer usecase properly. This is when the offer fails and we credit the maker with the remainder of the `provision - penalty`. In this case the offer has actually be fully deprovision and the `penalty` amount has been sent to the taker. Because of this, the indexer needs to know that this credit event, is actually deprovisioning the offer. We solve this by flipping the offerId to be a negative number. This way the indexer knows that the offer has been deprovisioned. 

Other options:
1. We could calculate the provision of each offer, by using the `gasbase`, `gasprice` and the `gasreq` for each offer. But this would add complexity to the indexer and if we ever want to change of how we calculate provision then this would be a dependancy we have to remember
2. We could add even more events. We could create an event for every single usecase, e.g. `CreditOfferWrite`, `CreditFundMgv`, `CreditWithdrawMgv`,`CreditOfferFail`, etc. This way we would for each event know in exactly what scenario the event was emitted and we would not need to flip the offerId when the offer is deprovisioned. It would most likely also increase readability. But it would most likely also increase the gas cost.

### Index taker
Very simple change, just added indexed to taker in the `OfferSuccess` event. This enables that we can now filter on the taker.

This is a POC, if you have any comments or thoughts, please don't hesitate to post them. 
The PR is inspired by this #144


src/MgvOfferMaking.sol
```diff
@@ -171,7 +171,7 @@ contract MgvOfferMaking is MgvHasOffers {
         provision = 10 ** 9 * offerDetail.gasprice() //gasprice is 0 if offer was deprovisioned
           * (offerDetail.gasreq() + offerDetail.offer_gasbase());
         // credit `balanceOf` and log transfer
-        creditWei(msg.sender, provision);
+        creditWei(msg.sender, provision, int(offerId));
```
espendk: Should the offerId be negative here since the offer is being deprovisioned?

Or did I misunderstand the meaning of negative offer ID's? 😅

espendk:
uh oh, let the event wars begin 🫣😂

But seriously, I think this may be a difficult discussion as witnessed by the discussions last time. So I feel the need to preface my input here with my current perspective on events:

It seems to me that events are difficult to do well, because they on the one hand provide an easy way to observe what happens on the blockchain (removing the need to simulate or query state) while they on the other hand add an overhead with no benefit to the smart contract itself. This creates a tension between ease of use and gas cost (EVM in a nutshell I guess).

My understanding after our discussions last time is that Mangrove events should be minimal. This (I think) implies the following:
1. If a needed piece of information can be deduced by parsing the full sequence of events, we don’t emit that piece of information explicitly.
2. Information should be compressed into as few fields as possible.

I think the negative offer ID’s in the Credit event in this PR is an example of 2.
I’m not sure we’ve been doing this until now?

I’m not a fan as I think it’s very non-obvious. But I see how it compresses information into one field.

How would gascost be affected, if we used a separate event instead of using negative offer ID’s?

src/MgvOfferMaking.sol
peterMangrove: Yes, you might be right, that it would be cleaner to do that here. But it is not needed, as the amount we credit here, is the actual amount that the offer had locked. Where as in the "applyPenalty" scenario, it is the difference of what the taker got minus what  provision was on the offer, that gets credited to the maker. Meaning that by just reading the amount credited in the "applyPenalty" scenario, is not be sufficient to keep a balance of provision of the offer.

An example might be easier to understand:
We post and offer and fund mangrove with 10 matic, (but the offer only needs 1), then these credit/debit event happens.
credit 10 matic to maker. (no offerId known here)
debit 1 matic to maker, with the offerid (lets say offerId = 1)

The offer then gets taken, but fails. We calculate the penalty to 0.5 matic. Then this event happens
credit (1-0.5=0.5) matic to maker with offer id 1.

In this scenario it now looks like the offer still has 0.5 provision locked, but that amount has actually been send as bounty to the taker and the offer is fully deprovisioned.

If we take the same scenario, but the user retracts its own offer and deprovisions it:
Same 2 first events:
credit 10 matic to maker. (no offerId, known here)
debit 1 matic to maker, with the offerid (lets say offerId = 1)

The user now retracts the offer and deprovisions it:
credit 1 matic to maker, with offerId 1

The provision balance of the offer is now 0 and we can therefor see that it is fully deprovisioned.

Because of this, it is only necessary to have a negative offerId to indicate deprovision in the first case. But i would not be opposed to also have it negative in the second case.

peterMangrove:
> uh oh, let the event wars begin 🫣😂
> 
> But seriously, I think this may be a difficult discussion as witnessed by the discussions last time. So I feel the need to preface my input here with my current perspective on events:
> 
> It seems to me that events are difficult to do well, because they on the one hand provide an easy way to observe what happens on the blockchain (removing the need to simulate or query state) while they on the other hand add an overhead with no benefit to the smart contract itself. This creates a tension between ease of use and gas cost (EVM in a nutshell I guess).
> 
> My understanding after our discussions last time is that Mangrove events should be minimal. This (I think) implies the following:
> 
> 1. If a needed piece of information can be deduced by parsing the full sequence of events, we don’t emit that piece of information explicitly.
> 2. Information should be compressed into as few fields as possible.
> 
> I think the negative offer ID’s in the Credit event in this PR is an example of 2. I’m not sure we’ve been doing this until now?
> 
> I’m not a fan as I think it’s very non-obvious. But I see how it compresses information into one field.
> 
> How would gascost be affected, if we used a separate event instead of using negative offer ID’s?

Pretty sure that we have never done anything like this before.
I agree that it is a very non-obvious way of doing it and would be hard for new people to understand. 
If we add a new event, lets say `Deprovision(address maker, uint remainingProvision, uint offerId)`, then i think it only adds gas cost to deployment, but emitting this new event vs emitting the credit event (with offerId), would be the exact same amount of gas.

src/MgvOfferMaking.sol
espendk: I don't feel strongly on how we solve this, I just think it's important that the meaning of the negative offer ID is clear and used consistently.

The current description of the event says:
> If the offerId is negative, then we are fully deprovisioning the offer */

With that description, the offer ID should be negative here, right?

Perhaps there's a more precise way to describe what a negative ID means, so that it does not have to be negative here?

espendk:
> If we add a new event, lets say `Deprovision(address maker, uint remainingProvision, uint offerId)`, then i think it only adds gas cost to deployment, but emitting this new event vs emitting the credit event (with offerId), would be the exact same amount of gas.

Understood. I like this :-)

peterMangrove:
@espendk @lnist I've added another Credit event, that also holds the information about how much penalty the offer/maker paid for a failed offer. By adding this information, then i do no longer need to make the offer Id negative. I now with this new information, know that if the credit event has a penalty, then the offer has been deprovisioned and the remaining provision has been added to the makers balance on Mangrove. 

I believe this fixed the issue we had, plus we now also has the information about how much a specific offer has paid in penalty. If non of you have any objections to this, then it would be great to get this PR merge into the ticks branch.

src/IMangrove.sol
```diff
@@ -13,7 +13,10 @@ import "./MgvLib.sol" as MgvLibWrapper;
 
 interface IMangrove {
   event Approval(address indexed outbound_tkn, address indexed inbound_tkn, address owner, address spender, uint value);
+  event Credit(address indexed maker, uint amount, int offerId);
```
lnist:
```suggestion
  event Credit(address indexed maker, uint amount, uint offerId);
```

lnist (approved):

adhusson (commented):
Thank you for this PR :) Agree adding info necessary for partial fill detection is important as well as indexing better!

re: partial fill detection, is there a method that would still give us partial fill detection but cost less gas? I'm asking the new event has 6 arguments. Would some kind of `PartialFill` event at the end work? If not, then let's go with `OrderStart`! (But I think you probably want to index some fields of `OrderStart`)

For the provision-related events I am sceptical but I may have misunderstood something, here is my rationale below:


> This change is creating two new events, Debit and Credit with offerId. (...) This is done in order to keep track of the provision for each offer. This way we can e.g. show in the UI for Kandel how much each offer has locked in provision.

I don't understand this, the total locked provision of a maker is encoded by the sum of (gasbase+gasreq)*gasprice of their offers. The fact that you have a corner case on OfferFail and have to encode additional info by flipping the offer sign shows that there's a mismatch here. By keeping track of `OfferRetract` `OfferWrite` etc it should be possible to "calculate the provision of each offer, by using the gasbase, gasprice and the gasreq for each offer", without additional RPC requests, right? 

I think this is a clear case where some additional code on the client side is much more worth it than additional code + gas costs on the smart contract. But maybe I missed something here and there is something currently not possible client-side that we want to make possible?

peterMangrove:
> Thank you for this PR :) Agree adding info necessary for partial fill detection is important as well as indexing better!
> 
> re: partial fill detection, is there a method that would still give us partial fill detection but cost less gas? I'm asking the new event has 6 arguments. Would some kind of `PartialFill` event at the end work? If not, then let's go with `OrderStart`! (But I think you probably want to index some fields of `OrderStart`)
> 
> For the provision-related events I am sceptical but I may have misunderstood something, here is my rationale below:
> 
> > This change is creating two new events, Debit and Credit with offerId. (...) This is done in order to keep track of the provision for each offer. This way we can e.g. show in the UI for Kandel how much each offer has locked in provision.
> 
> I don't understand this, the total locked provision of a maker is encoded by the sum of (gasbase+gasreq)*gasprice of their offers. The fact that you have a corner case on OfferFail and have to encode additional info by flipping the offer sign shows that there's a mismatch here. By keeping track of `OfferRetract` `OfferWrite` etc it should be possible to "calculate the provision of each offer, by using the gasbase, gasprice and the gasreq for each offer", without additional RPC requests, right?
> 
> I think this is a clear case where some additional code on the client side is much more worth it than additional code + gas costs on the smart contract. But maybe I missed something here and there is something currently not possible client-side that we want to make possible?

Regarding `OrderStart` and partial fill, then you are right. I actually don't need out_tkn, in_tkn or taker, as they are emitted in the `OrderComplete`, but the `fillVolume`, `maxTick` and `fillWants`, are necessary in my mind. I would rather emit the raw data, than we in the contract check if it is partial fill or not. This way the reader of the events, can do what they want with the information. I'll update the `OrderStart` event

Regarding the Provision and its connection to offers. Then you are right, that we could calculate the provision pr offer, by gasbase, gasreq and gasprice. This would require more complex logic in the indexer/SDK. And that logic would be somewhat bridle as it builds on a lot of other data, emitted by several events, so the chance that we at some point break this logic is not unthinkable. @lnist also added the deprovision flag to an event some time ago, as we did not know if an offer was deprovisioned or not, here he also mentions that adding this extra info to credit and debit could be a next step, in order to be more explicit. https://github.com/mangrovedao/mangrove-core/pull/311

Regarding penalty, here we are not able to figure out what offer paid what penalty with the current events. Because of that i added the extra Credit event, with the penalty and offerId. This way we know exactly what offer caused what penalty.

All in all, we would like to be a little more explicit in the data we emit, as it reduces both risk and complexity of the indexer and SDK logic.

espendk:
@peterMangrove This is deprecated and should be closed, right?

I'll close it and then you can reopen it if needed :-)

PR #425, opened 2023-08-02
remove strats

peterMangrove:
The purpose of this PR is to split Core and Strat contracts in separate repos. This is done to be very clear what is the core of Mangrove and it keeps Core and Strats code from being mixed up together.

This PR removes all strat related files (contracts, scripts, deployers, etc.). Some files needed small changes in order for them to work, with all the removed files.

A following PR in the new [mangrove-strats ](https://github.com/mangrovedao/mangrove-strats)repo will do the same, where all Core related files will be removed.


config.js
```diff
@@ -6,20 +6,7 @@ exports.abi_exports = [
   "MgvOracle",
   "TestToken",
   "IERC20",
-  "MangroveOrder",
```
lnist, line 9: Clean up addresses/*?

test/toy_strategies/TestToken.t.sol
espendk, line 1: Should this have been deleted?

espendk (dismissed):
Nice!

Would be great if the PR described the motivation and steps for this change + pointed to the new strat repo.

We could consider keeping more deployers here, though it might require a bit of refac. Examples:
- MumbaiActivateMarket (cannot activate MangroveOrder)
- MumbaiFaucet* (could be kept unchanged?)
- ERC20Deployer and related

config.js
peterMangrove: Done

test/toy_strategies/TestToken.t.sol
peterMangrove: Reinstanted test, with removed unused imports

lnist (approved):

espendk (approved):
👌

PR #428, opened 2023-08-04
Feat/ticks taker tv

lnist:
This is the taker-side of " Internally implement everything in terms of tick/volume, then add interfaces in gives/wants" from https://github.com/mangrovedao/mangrove-core/pull/380

Comments have in general not been updated.

Commits are hopefully self-explanatory in their title, but some comments:
1 `feat(taker): Change average limit order to limit order and apply tick further out` - Changed to limit order by using tick, note we currently do not have a test for difference between average limit and limit. Also, a snipe-test was changed in a hacky way, should be updated by changing snipe to a tick-based API.
2. `feat(fillVolume): Use fillVolume instead of calculating gives/wants` - here "sor.gives" is renamed to "sor.givesToThisOffer" - this can be changed back but is useful to see changes.
3. `feat(stack): Reduce stack for consumption for market order` - introduces a partialFill for execute - similar result could be achieved by packing gives/wants into a single uint. It leaves a TODO for further improvement.
4. `fix(marketOrder): Do not store initialWants/gives` - sor.givesToThisOffer can be used instead.
5. `feat(marketOrder): Introduce tick-based API and overloads`: Introduction of new APIs

This is still WIP and the last commit opens up the discussion on what the API should be and what should be in libs/helpers and what should be the raw API.

Other commits can also be removed if we decide so.

### DIFF
--- a/lib/TickLib.sol
+++ b/lib/TickLib.sol
@@ -166,6 +166,10 @@ library TickLib {
   // FP.lnWad(BP)
   uint constant lnBP = 99995000333308;
 
+  function inRange(Tick tick) internal pure returns (bool) {
+    return Tick.unwrap(tick) >= MIN_TICK && Tick.unwrap(tick) <= MAX_TICK;
+  }
+
   function tickFromBranch(uint tickPosInLeaf,Field level0, Field level1, Field level2) internal pure returns (Tick) {
     unchecked {
       uint utick = tickPosInLeaf |
@@ -235,8 +239,7 @@ library TickLib {
     - with volumes on 96 bits, and a tick in range, there is no overflow doing bp^tick * 1e18 * volume / 1e18
     */
   function priceFromTick_e18(Tick tick) internal pure returns (uint) {
-    require(Tick.unwrap(tick) >= MIN_TICK,"mgv/tick/tooLow");
-    require(Tick.unwrap(tick) <= MAX_TICK,"mgv/tick/tooHigh");
+    require(inRange(tick),"mgv/priceFromTick/outOfRange");
     // FIXME this must round up so tick(price(tick)) = tick
     // FIXME add a test for this
     // Right now e.g. priceFromTick(1) is too low, and tickFromVolumes(1 ether,Tick(1).outboundFromInbound(1 ether)) is 0 (should be 1)
--- a/lib/TickLib.sol
+++ b/lib/TickLib.sol
@@ -190,35 +190,39 @@ library TickLib {
   // with wants/gives on 96 bits, tick will be < 24bits
   // never overstimates tick (but takes the highest tick that avoids doing so)
   // wants will be adjusted down from the original
+  // FIXME use unchecked math but specify precise bounds on what inputs do not overflow
   function tickFromVolumes(uint inboundAmt, uint outboundAmt) internal pure returns (Tick) {
-    unchecked {
-      (uint num, uint den) = (inboundAmt,outboundAmt);
-      if (inboundAmt < outboundAmt) {
-        (num,den) = (den,num);
-      } else {
-      }
-      int lnPrice = FP.lnWad(int(num * 1e18/den));
-      // why is univ3 not doing that? Because they start from a ratio < 1 ?
-      int lbpPrice = int(FP.divWad(uint(lnPrice),lnBP)/1e18);
-      // note this implies the lowest tick will never be used! (could use for something else?)
-      if (inboundAmt < outboundAmt) {
-        lbpPrice = - lbpPrice;
-      }
-
-      uint pr = Tick.wrap(lbpPrice).priceFromTick_e18();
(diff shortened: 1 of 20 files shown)


lib/TickLib.sol
```diff
@@ -178,6 +178,14 @@ library TickLib {
     }
   }
 
+  function tickFromTakerVolumes(uint takerGives, uint takerWants) internal pure returns (Tick) {
```
adhusson, line 185: With average limit order, it is possible to have 

- wants=N, gives=MAX, fillWants=true (means "buy N at any price")

as well as : 

- wants=0, gives=N, fillWants=false (means "sell N at any price")

I think the current PR reproduces that, but maybe we should add tests to check it?

src/MgvOfferTaking.sol
```diff
@@ -35,11 +36,30 @@ abstract contract MgvOfferTaking is MgvHasOffers {
...
+
+  function marketOrderPrice(address outbound_tkn, address inbound_tkn, uint priceLimit, uint fillVolume, bool fillWants)
     external
     returns (uint, uint, uint, uint)
+  {
+    int maxTick = Tick.unwrap(TickLib.tickFromPrice_e18(priceLimit));
+    return marketOrder(outbound_tkn, inbound_tkn, maxTick, fillVolume, fillWants);
+  }
+
+  function marketOrder(address outbound_tkn, address inbound_tkn, int limitTick, uint fillVolume, bool fillWants)
```
adhusson: needs tests for when tick/fillVolume are outside range

src/MgvOfferTaking.sol
```diff
@@ -35,11 +36,30 @@ abstract contract MgvOfferTaking is MgvHasOffers {
...
+    returns (uint, uint, uint, uint)
+  {
+    require(uint160(takerWants) == takerWants, "mgv/mOrder/takerWants/160bits");
+    require(uint160(takerGives) == takerGives, "mgv/mOrder/takerGives/160bits");
+    uint fillVolume = fillWants ? takerWants : takerGives;
+    int maxTick = Tick.unwrap(TickLib.tickFromTakerVolumes(takerGives, takerWants));
+    return marketOrder(outbound_tkn, inbound_tkn, maxTick, fillVolume, fillWants);
+  }
+
+  function marketOrderPrice(address outbound_tkn, address inbound_tkn, uint priceLimit, uint fillVolume, bool fillWants)
```
adhusson: needs tests for when price/fillVolume are outside range

src/MgvOfferTaking.sol
```diff
@@ -191,24 +209,22 @@ abstract contract MgvOfferTaking is MgvHasOffers {
         /* `execute` will adjust `sor.wants`,`sor.gives`, and may attempt to execute the offer if its price is low enough. It is crucial that an error due to `taker` triggers a revert. That way, [`mgvData`](#MgvOfferTaking/statusCodes) not in `["mgv/notExecuted","mgv/tradeSuccess"]` means the failure is the maker's fault. */
         /* Post-execution, `sor.wants`/`sor.gives` reflect how much was sent/taken by the offer. We will need it after the recursive call, so we save it in local variables. Same goes for `offerId`, `sor.offer` and `sor.offerDetail`. */
 
-        (gasused, makerData, mgvData) = execute(pair, mor, sor);
+        bool partialFill;
```
adhusson: Needs more doc on why this variable is required now

src/MgvOfferTaking.sol
```diff
@@ -191,24 +209,22 @@ abstract contract MgvOfferTaking is MgvHasOffers {
...
-        uint takerGives = sor.gives;
+        /* Keep cached copy of current `sor` values to restore them later to send to posthook. */
         uint offerId = sor.offerId;
         MgvStructs.OfferPacked offer = sor.offer;
         MgvStructs.OfferDetailPacked offerDetail = sor.offerDetail;
 
         /* If an execution was attempted, we move `sor` to the next offer. Note that the current state is inconsistent, since we have not yet updated `sor.offerDetails`. */
         if (mgvData != "mgv/notExecuted") {
-          sor.wants = mor.initialWants > mor.totalGot ? mor.initialWants - mor.totalGot : 0;
+          mor.fillVolume -= mor.fillWants ? sor.wantsFromThisOffer : sor.givesToThisOffer;
```
adhusson: do we know that there will not be an underflow?

adhusson: replying to myself: yes we do! added comments to that effect in `execute`

src/MgvOfferTaking.sol
adhusson: variable removed for now

lib/TickLib.sol
adhusson: it seems ok

src/MgvOfferTaking.sol
adhusson: done

PR #431, opened 2023-08-06
Dynamic ticks

adhusson:
(PR is currently empty, this is a placeholder) 

2 options for dynamic ticks. The choice is about whether the ticksize of a pair (eg. should price points be every base tick, every 2nd base tick, every 10 base tick, etc?) is auto-handled by Mangrove or not.

* **raw**: offerlists are now a triplet `(outbound, inbound, tickScale)`. You can open/close them like before. Users should specify the right `tickScale` when interacting with mangrove — they should know “where the action is”.

* **assisted**: offerlists are _internally_ `(outbound,inbound,tickScale)`, but Mangrove has an internal notion of the “current” `tickScale`. The public-facing ABI only lets people give an `(outbound,inbound)` pair, so people can only interact with the current `tickScale` (except to remove offers that were created when a different `tickScale` was the default). When the `tickScale` is changed by governance, all current offers linger in a “zombie” state where they cannot be taken (but don’t pollute the OB) and can only be retracted.

**raw** has less complex smart contract code but requires more client-side work (and more cognitive overhead in general for integrators) and increases fragmentation risk. It is also more ‘stable’ from the pov of ecosystem development / smart contract: a triplet clearly defines the set of offers you’ll interact with and with what ticksize.

**assisted** has more complex smart contract code but maintains the current straightforward ABI (except for `retractOffer`) and forces all liquidity for a pair to be in one place. It is also less ‘stable’ from the pov of ecosystem development / smart contracts: offers can become un-changeable without a `retractOffer`, and the detailed semantics of the `(outbound,inbound)`` pair they interact with can change with no warning.

Note: by its nature UniV3 goes beyond **raw**: it cannot even prevent the activation of a pool with a given tick size, whereas Mangrove governance can, under the **raw** option, enable/disable `(out,in,tickScale)` triplets.

Since **assisted** is 1) nontrivial codewise and 2) introduces too much smart contract risk to my taste and 3) is less predictable as an ‘ecosystem object’ and also 4) anyway governance can always choose to enable only one triplet at a time, I would like to start work on a raw POC for now. But what do you think?


lnist:
I agree with your assessment. An even more restrictive "raw" option is to have tickScale constant for a Mangrove instance, but then have multiple Mangrove instances, one for each desired tickScale. It would complicate interaction and integration, but can end up with multiple Mangrove instances anyway due to upgrades - this would force our hand in that direction (it is not without cost to deployment, mangrove.js and UI to do this).
A benefit would be simpler maps (saves a bit of hashing) and fewer parameters and very controlled set of tickScales.

PR #433, opened 2023-08-07
feat(maker): Introduce tick-based maker API

espendk:
This PR adds a tick maker API `{new|update}OfferByTick` and renames the old `{new|update}Offer` functions to `{new|update}OfferByVolume`.

## Notes
- Gatekeeping tests for the API should be revisited once input ranges and precisions is stable.
- The `test_snipe_correct_amount_auto` test in `TakerOperations.t.sol` has been commented out as it was failing; It's not clear why.


adhusson (approved):
LGTM, I just restored the `insertionTick` name because I had `tick` for a while and renaming it to `insertionTick` helped me reason about code correctness way more easily

edit: I don't know if you want to merge or squash so I'll let you go over it :)

PR #434, opened 2023-08-07
Feat/ticks snipes

adhusson:
Snipes are now tick-based

New file: `MgvHelpers.sol` gives tools to call `snipesByVolume` or to manually convert a `targets` array to ticks. These utilities are now used by tests.

### DIFF
--- a/src/MgvHelpers.sol
+++ b/src/MgvHelpers.sol
@@ -1,14 +1,12 @@
 // SPDX-License-Identifier: Unlicense
 
-/* `MgvLib` contains data structures returned by external calls to Mangrove and the interfaces it uses for its own external calls. */
-
 pragma solidity ^0.8.10;
 
 import {IMangrove} from "mgv_src/IMangrove.sol";
 import "mgv_lib/TickLib.sol";
 
 library MgvHelpers {
-  // Converts snipe targets from volume-based [id,wants,gives,gasreq] tick-based [id,tick,volume,gasreq]. volume will be wants if fillWsants is true, volume will be gives otherwise.
+  // Converts snipe targets from volume-based [id,wants,gives,gasreq] tick-based [id,tick,volume,gasreq]. volume will be wants if fillWsants is true, volume will be gives otherwise. Note that `tick` is fundamentally an int but arrays are homogenous so must be typed as a uint in the call.
   function convertSnipeTargetsToTicks(uint[4][] memory targets, bool fillWants)
     internal
     pure
--- a/src/MgvHelpers.sol
+++ b/src/MgvHelpers.sol
@@ -1,0 +1,53 @@
+// SPDX-License-Identifier: Unlicense
+
+/* `MgvLib` contains data structures returned by external calls to Mangrove and the interfaces it uses for its own external calls. */
+
+pragma solidity ^0.8.10;
+
+import {IMangrove} from "mgv_src/IMangrove.sol";
+import "mgv_lib/TickLib.sol";
+
+library MgvHelpers {
+  // Converts snipe targets from volume-based [id,wants,gives,gasreq] tick-based [id,tick,volume,gasreq]. volume will be wants if fillWsants is true, volume will be gives otherwise.
+  function convertSnipeTargetsToTicks(uint[4][] memory targets, bool fillWants)
+    internal
+    pure
+    returns (uint[4][] memory)
+  {
+    uint[4][] memory newTargets = new uint[4][](targets.length);
+    // convert targets from [id,wants,gives,gasreq] to [id,tick,volume,gasreq]
+    uint volumeIndex = fillWants ? 1 : 2;
+    for (uint i = 0; i < targets.length; i++) {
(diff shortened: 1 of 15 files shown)


src/MgvHelpers.sol
```diff
@@ -0,0 +1,53 @@
+// SPDX-License-Identifier: Unlicense
+
+/* `MgvLib` contains data structures returned by external calls to Mangrove and the interfaces it uses for its own external calls. */
```
espendk: This comment seems unrelated to this file - perhaps copy/paste leftover?

src/MgvOfferTaking.sol
```diff
@@ -310,8 +310,9 @@ abstract contract MgvOfferTaking is MgvHasOffers {
   /* ## Snipes */
   //+clear+
 
-  /* `snipes` executes multiple offers. It takes a `uint[4][]` as penultimate argument, with each array element of the form `[offerId,takerWants,takerGives,offerGasreq]`. The return parameters are of the form `(successes,snipesGot,snipesGave,bounty,feePaid)`. 
+  /* `snipes` executes multiple offers. It takes a `uint[4][]` as penultimate argument, with each array element of the form `[offerId,tick,fillVolume,offerGasreq]`. The return parameters are of the form `(successes,snipesGot,snipesGave,bounty,feePaid)`.
```
espendk, line 313: `tick` is an `int` in the other interfaces, so slightly surprising for it to a `uint` here. I think it's fine, as snipe will be a specialist call for cleaning, but we should probably call it out explicitly.

src/MgvOfferTaking.sol
```diff
@@ -392,16 +393,16 @@ abstract contract MgvOfferTaking is MgvHasOffers {
...
-          mor.fillVolume = mor.fillWants ? wants : gives;
-          /* Wants gets normalized to the closest tick */
-          // note that `takerGives=0` is not possible here
-          // Tick tt = TickLib.tickFromVolumes(gives,wants);
-          // sor.tick = TickLib.tickFromVolumes(gives,wants);
-          mor.maxTick = TickLib.tickFromTakerVolumes(gives, wants);
+          {
+            Tick tick = Tick.wrap(int(targets[i][1]));
+            require(TickLib.inRange(tick), "mgv/snipes/tick/outOfRange");
+            mor.maxTick = tick;
```
espendk, line 399: I was expecting snipe to only work if the offer was at the specified tick. But it seems we also execute offers where the price is better than expected?

src/periphery/MgvCleaner.sol
```diff
@@ -30,7 +31,8 @@ contract MgvCleaner {
     returns (uint bal)
   {
     unchecked {
-      (uint successes,,,,) = MGV.snipesFor(outbound_tkn, inbound_tkn, targets, fillWants, msg.sender);
+      (uint successes,,,,) =
+        MgvHelpers.snipesForByVolume(address(MGV), outbound_tkn, inbound_tkn, targets, fillWants, msg.sender);
```
espendk, line 35: Assume the MgvCleaner API was left unchanged in anticipation of it being removed?

espendk (approved):
LGTM. Added a few comments.

I think most of the snipe tests are more natural when using the tick API, so I've taken the liberty of migrating them. With the exception of the MgvCleaner tests (and API) which will soon be irrelevant anyway.

As discussed, sniping will become a specialized API for cleaning. So I think we should consider removing the utility methods from MgvHelpers. But perhaps we could move the other ByVolume versions of the maker and taker APIs there?

espendk:
I forgot to mention one thing: Creating the `targets` for `snipes` has become somewhat clunky due to ticks:

```solidity
wrap_dynamic([ofr, uint(Tick.unwrap(tick)), 1 ether, 300_000])
```

I couldn't come up with something better outside of creating a utility method for it. Do you see better options?

src/MgvHelpers.sol
adhusson: yes

src/MgvOfferTaking.sol
adhusson: yes, the idea is that the taker specifies the offer tick as a protection against the offer having changed. But if the offer has _improved_ its price the taker needs no protection

adhusson:
> I forgot to mention one thing: Creating the `targets` for `snipes` has become somewhat clunky due to ticks:
> 
> ```solidity
> wrap_dynamic([ofr, uint(Tick.unwrap(tick)), 1 ether, 300_000])
> ```
> 
> I couldn't come up with something better outside of creating a utility method for it. Do you see better options?

Not with arrays, but we may figure something out with structs!

src/MgvOfferTaking.sol
adhusson: will add a comment

src/periphery/MgvCleaner.sol
adhusson: just minimizing changes :p

PR #437, opened 2023-08-11
Replace `snipes` with `cleanByImpersonation` function

espendk:
This PR removes `snipes` and `snipesFor` from Mangrove and replaces them with a `cleanByImpersonation` function that only allows failing offers to be executed.

# Major changes

In addition to the primary change (not allowing snipes in general) the following changes have been made:

## `MgvCleaner` has been removed
Mangrove core now provides the cleaning functionality that `MgvCleaner` used to provide, so it `MgvCleaner` has been removed.

##  Cleaning of multiple offers is more resilient
Before, `MgvCleaner.collect*` would revert if execution of any offer succeeded. The new `cleanByImpersonation` is more resilient as it will not revert the whole tx if an offer succeeds but will revert just the execution of that offer and continue cleaning of the next offers. 

Thus a cleaner that attempts to clean multiple offers in one tx will be able to successfully clean some offers despite some offers failing to be cleaned.

## Taker impersonation no longer checks permits
The new `cleanByImpersonation` function has integrated the `taker` parameter of the old `snipesFor`. This allows cleaning by impersonation, ie where another taker's funds and approval of Mangrove for transfers of quote can be used to execute an offer and make it fail (which was already possible via `MgvCleaner.collectByImpersonation`).

This has been integrated into the `cleanByImpersonation` function instead of having a separate `cleanFor` function because
1. no permit/allowance is needed nor does it have to be checked and
2. cleaners are expected to be advanced users, so requiring an explicit `taker` parameter with non-obvious semantics seems ok.

# Migration of tests away from snipes
Many tests rely on sniping. To allow gradual migration away from this, the snipe functions have been moved to a `TestSnipingMangrove` contract that `MangroveTest` provides to tests.

To make clear that these tests do something non-standard, the snipe functions have been renamed to `snipesInTest` and `snipesForInTest`.

Migration of tests will be done in later PR's as merging this PR will (1) hopefully stop further use of snipes in tests and (2) enable that we can start updating downstream components to the new MGV V2 API.


src/MgvOfferTaking.sol
```diff
@@ -306,146 +306,97 @@ abstract contract MgvOfferTaking is MgvHasOffers {
...
-      /* Over the course of the snipes order, a penalty reserved for `msg.sender` has accumulated in `mor.totalPenalty`. No actual transfers have occured yet -- all the ethers given by the makers as provision are owned by Mangrove. `sendPenalty` finally gives the accumulated penalty to `msg.sender`. */
-      sendPenalty(mor.totalPenalty);
-      //+clear+
-
-      emit OrderComplete(sor.outbound_tkn, sor.inbound_tkn, taker, snipesGot, snipesGave, mor.totalPenalty, mor.feePaid);
-      totalPenalty = mor.totalPenalty;
-      feePaid = mor.feePaid;
-    }
-  }
+      /* If we removed the `isLive` conditional, a single expired or nonexistent offer in `targets` would revert the entire transaction (by the division by `offer.gives` below since `offer.gives` would be 0). We also check that `gasreq` is not worse than specified. A taker who does not care about `gasreq` can specify any amount larger than $2^{24}-1$. A mismatched price will be detected by `execute`. */
```
adhusson:
```suggestion
      /* FIXME: edit comment: If we removed the `isLive` conditional, a single expired or nonexistent offer in `targets` would revert the entire transaction (by the division by `offer.gives` below since `offer.gives` would be 0). We also check that `gasreq` is not worse than specified. A taker who does not care about `gasreq` can specify any amount larger than $2^{24}-1$. A mismatched price will be detected by `execute`. */
```

src/MgvOfferTakingWithPermit.sol
```diff
@@ -124,23 +124,17 @@ abstract contract MgvOfferTakingWithPermit is MgvOfferTaking {
...
-    unchecked {
-      (successes, takerGot, takerGave, bounty, feePaid) =
-        generalSnipes(outbound_tkn, inbound_tkn, targets, fillWants, taker);
-      /* The sender's allowance is verified after the order complete so that the actual amounts are checked against the allowance, instead of the declared `takerGives`. The former may be lower.
-    
-    An immediate consequence is that any funds available to Mangrove through `approve` can be used to clean offers. After a `snipesFor` where all offers have failed, all token transfers have been reverted, so `takerGave=0` and the check will succeed -- but the sender will still have received the bounty of the failing offers. */
-      deductSenderAllowance(outbound_tkn, inbound_tkn, taker, takerGave);
-    }
-  }
+  // FIXME: This doesn't use permit or allowance at all, so perhaps it's better to move it to MgvOfferTaking? Or simply expose the `taker` parameter there?
```
adhusson: Removing `deductSenderAllowance` is scary even though I do not see an attack right now.

espendk:
~~Just realized that the old snipe tests that test the details of execution (eg events emitted and transfers made) should be migrated as part of this PR. Will look at that Monday!~~

On second thought (and seeing #444 building on this PR) I think we should merge this after review and include the offer execution tests in the upcoming test migration PR. The feature will still be appropriately tested before merging to `develop`.

src/MgvOfferTaking.sol
```diff
@@ -306,146 +306,97 @@ abstract contract MgvOfferTaking is MgvHasOffers {
...
-     From an array of _n_ `[offerId, tick,fillVolume,gasreq]` elements, execute each snipe in sequence. Returns `(successes, takerGot, takerGave, bounty, feePaid)`. 
-
-     Note that if this function is not internal, anyone can make anyone use Mangrove.
-     Note that unlike general market order, the returned total values are _not_ `mor.totalGot` and `mor.totalGave`, since those are reset at every iteration of the `targets` array. Instead, accumulators `snipesGot` and `snipesGave` are used. */
-  function generalSnipes(
+  /* # Cleaning */
+  // FIXME: Document cleaning
+  // FIXME: Maybe `cleanFailingOffer` instead?
+  /* Any `taker` can be impersonated when cleaning because the function reverts if the offer succeeds, cancelling any token transfers. And after a `clean` where the offer has failed, all token transfers have been reverted -- but the sender will still have received the bounty of the failing offers. */
+  // FIXME: This would be simpler if tick had to match exactly: Then `fillWants` wouldn't be needed and `fillVolume` could be simplified to either `takerWants` or `takerGives`.
```
lnist: Do we have a good use case for non-exact match? Simplified code seems like a win.

src/MgvOfferTaking.sol
```diff
@@ -306,146 +306,97 @@ abstract contract MgvOfferTaking is MgvHasOffers {
...
-          }
+        require(mgvData != "mgv/tradeSuccess", "mgv/clean/offerDidNotFail");
+        require(mgvData != "mgv/notExecuted", "mgv/clean/tickTooLow");
 
-          /* In the market order, we were able to avoid stitching back offers after every `execute` since we knew a continuous segment starting at best would be consumed. Here, we cannot do this optimisation since offers in the `targets` array may be anywhere in the book. So we stitch together offers immediately after each `execute`. */
-          if (mgvData != "mgv/notExecuted") {
-            // updates best&tick
-            sor.local = dislodgeOffer(pair, sor.offer, sor.local, true);
-          }
+        /* In the market order, we were able to avoid stitching back offers after every `execute` since we knew a continuous segment starting at best would be consumed. Here, we cannot do this optimisation since offers in the `targets` array may be anywhere in the book. So we stitch together offers immediately after each `execute`. */
```
lnist: also a comment that needs to be updated. You added a FIXME further up, maybe also here? Though I suspect the code has quite a bit of outdated comments now.

src/MgvOfferTaking.sol
```diff
@@ -306,146 +306,97 @@ abstract contract MgvOfferTaking is MgvHasOffers {
...
-          /* In an inverted Mangrove, amounts have been lent by each offer's maker to the taker. We now call the taker. This is a noop in a normal Mangrove. */
-          executeEnd(mor, sor);
+        /* No fees are paid since offer execution failed. */
 
-          /* After an offer execution, we may run callbacks and increase the total penalty. As that part is common to market orders and snipes, it lives in its own `postExecute` function. */
-          if (mgvData != "mgv/notExecuted") {
-            postExecute(mor, sor, gasused, makerData, mgvData);
-          }
+        /* In an inverted Mangrove, amounts have been lent by each offer's maker to the taker. We now call the taker. This is a noop in a normal Mangrove. */
+        // FIXME: Is this needed when offer execution failed? I don't think so, but should be covered by tests
```
lnist: Do we leave this FIXME behind?

src/MgvOfferTaking.sol
```diff
@@ -306,146 +306,97 @@ abstract contract MgvOfferTaking is MgvHasOffers {
...
-          snipesGave += mor.totalGave;
-        }
+        /* After an offer execution, we may run callbacks and increase the total penalty. As that part is common to market orders and snipes, it lives in its own `postExecute` function. */
+        postExecute(mor, sor, gasused, makerData, mgvData);
       }
+
+      bounty = mor.totalPenalty;
+
+      /* Over the course of the snipes order, a penalty reserved for `msg.sender` has accumulated in `mor.totalPenalty`. No actual transfers have occured yet -- all the ethers given by the makers as provision are owned by Mangrove. `sendPenalty` finally gives the accumulated penalty to `msg.sender`. */
+      sendPenalty(bounty);
```
lnist: Previously we could clean multiple offers and then in the end do the transfer (at cast 21000 gas). We now pay that amount of gas for each transfer, thus upping the needed penalty to cover the cost.

Should gasbase compensate a normal market order taker, or a cleaner? If it is to incentivize cleaning, then that is the important one.

src/MgvOfferTaking.sol
```diff
@@ -306,146 +306,97 @@ abstract contract MgvOfferTaking is MgvHasOffers {
...
-          (uint gasused, bytes32 makerData, bytes32 mgvData) = execute(pair, mor, sor);
+      {
+        /* `execute` will adjust `sor.wants`,`sor.gives`, and may attempt to execute the offer if its price is low enough. It is crucial that an error due to `taker` triggers a revert. That way [`mgvData`](#MgvOfferTaking/statusCodes) not in `["mgv/tradeSuccess","mgv/notExecuted"]` means the failure is the maker's fault. */
+        /* Post-execution, `sor.wants`/`sor.gives` reflect how much was sent/taken by the offer. */
+        (uint gasused, bytes32 makerData, bytes32 mgvData) = execute(pair, mor, sor);
 
-          if (mgvData == "mgv/tradeSuccess") {
-            successCount += 1;
-          }
+        require(mgvData != "mgv/tradeSuccess", "mgv/clean/offerDidNotFail");
```
lnist, line 410: This looks correct - however, would it be safer to explicitly check the allowed failure values? 
`require(mgvData == "mgv/makerRevert" || mgvData == "mgv/makerTransferFail" || mgvData == "mgv/makerReceiveFail", "mgv/clean/offerDidNotFail");

src/periphery/MgvCleaner.sol
```diff
@@ -3,17 +3,15 @@ pragma solidity ^0.8.10;
...
- * their associated bounty. It takes an array of offers with same definition as
- * `Mangrove.snipes` and expects them all to fail or not execute. */
+ * their associated bounty. It takes an array of offers and will attempt to clean them individually. */
 
 /* How to use:
    1) Ensure *your* address approved Mangrove for the token you will provide to the offer (`inbound_tkn`).
    2) Run `collect` on the offers that you detected were failing.
 
-   You can adjust takerWants/takerGives and gasreq as needed.
+   Instead of using your own address, you can use `collectByImpersonation` to impersonate another taker (`takerToImpersonate`) who has approved Mangrove for `inbound_tkn`. This allows borrowing that taker's `inbound_tkn` funds for cleaning instead of using `msg.sender`'s funds (who need not have any).
```
lnist: takerToImpersonate is now just `taker`, right?

test/core/Gatekeeping.t.sol
```diff
@@ -607,13 +607,52 @@ contract GatekeepingTest is IMaker, MangroveTest {
...
+    posthook_cb = abi.encodeCall(this.snipesOK, ($(base), $(quote), other_ofr));
+
+    uint ofr = mgv.newOfferByVolume($(base), $(quote), 1 ether, 1 ether, 190_000, 0);
+    assertTrue(tkr.take(ofr, 1 ether), "take must succeed or test is void");
+    assertTrue(mgv.best($(base), $(quote)) == 0, "snipe in posthook must work");
+  }
+
   /* Snipe failure */
 
   function snipesKO(uint id) external {
```
lnist, line 661: is snipesKO replaces by cleanKO?

src/MgvOfferTaking.sol
adhusson: I'm wondering about the opposite: should we completely remove the price match? Initially it's there for "normal snipes" (to protect the taker). In the absence of normal snipes, what happens if we just ignore the price?

src/MgvOfferTaking.sol
adhusson: The intent was to be passively future-proof against additional errors that would be added later. I think that still applies so I would tend towards keeping it as is.

src/MgvOfferTaking.sol
espendk: All of the comments in that file should be updated :-) I made a general comment on this (and @adhusson added a specific one) and I think it'll be easiest to do it in one go once the `feat/ticks` has stabilized. But I can be persuaded otherwise :-)

src/MgvOfferTaking.sol
espendk: Good point, I'll remove it.

adhusson: To answer the FIXME: `executeEnd` is a noop on Mangrove anyway but is useful on InvertedMangrove (it calls the taker contract).

src/MgvOfferTaking.sol
adhusson: I agree -- gasbase should compensate max(market order overhead, cleaning overhead). 

I didn't get the 21k gas reference?

src/MgvOfferTakingWithPermit.sol
adhusson: I would rather keep it in :)

src/MgvOfferTaking.sol
espendk: Good point!

I think this is a strong reason for moving the multi-clean functionality from `MgvCleaner` to the core. Should be pretty quick to do.

Unless @adhusson objects, I'll go ahead and make the change.

adhusson (approved):
lgtm pending comments above

src/MgvOfferTaking.sol
espendk: Hmm, the point here is to revert explicitly if the offer succeeds, because then it shouldn't be cleaned.
But maybe that isn't clear?

src/periphery/MgvCleaner.sol
espendk: Indeed, I'll fix.

test/core/Gatekeeping.t.sol
espendk: Yes, I missed a rename when adapting the reentrancy tests and fixed on the test migration branch - I didn't expect you to notice ;-)

I'll port the fix here!

src/MgvOfferTaking.sol
espendk: I think the cleaner risks spending more gas on failing to make the offer fail: The offer is now more expensive and no longer fails.

src/MgvOfferTakingWithPermit.sol
adhusson: @espendk convinced me

src/MgvOfferTaking.sol
espendk: So `executeEnd` should still be invoked when an offer fails, right?

src/MgvOfferTaking.sol
lnist: The scenario is that the maker updates the offer between us making the decision to clean and actually cleaning - I don't think we can a priori say anything about whether it would still fail. Skipping early saves gas - in case of failure still, we win gas. I don't know what is most common.
I lean towards skipping since information has changed - spending gas on executing the entire offer if it succeeds is quite wasteful.

src/MgvOfferTaking.sol
espendk: Done in [e4cfb66](https://github.com/mangrovedao/mangrove-core/pull/437/commits/e4cfb668bd6ebfb2547952379be91c8c510a1482)

src/MgvOfferTaking.sol
espendk: Reintroduce `executeEnd` in [76393f2](https://github.com/mangrovedao/mangrove-core/pull/437/commits/76393f2b9000721d1456705f839d761e0d1e9244)

src/MgvOfferTaking.sol
espendk: I went for the strict match and `takerWants` instead of `fillVolume` and `fillWants` in [92eff73](https://github.com/mangrovedao/mangrove-core/pull/437/commits/92eff7372aa66d7dd695a6b7a741334d3f9add01)

PR #444, opened 2023-08-13
Dynamic ticks (2)

adhusson:
**To review, read in priority files that don't end in .s.sol, .t.sol, or .post.sol.**

## TODO
Now merged with #380's TODO

## Summary of dynamic ticks
POC for **raw** dynamic ticks: offerLists are now characterised a triplet `(outbound, inbound, tickScale)`. You can open/close them like before. Users should specify the right `tickScale` when interacting with mangrove — they should know “where the action is”.

The `tickScale` value specifies the distance between ticks, in number of price points (a price point is 1 basis point for now).

Example:
* `tickScale=1`: ticks are 1 bp apart
* `tickScale=8`: if offers have tick `i` have price `p`, offers at tick `i+1` have price `p * 1.0001^8`, ie ticks are ~8 bp apart.

Why **raw** price ticks, as opposed to auto-managed price ticks [^1]:
- more ‘stable’ from the pov of ecosystem development / smart contract: a triplet clearly defines the set of offers you’ll interact with and with what ticksize.
- less complex smart contract code in Mangrove
- governance can always choose to enable at most one `tickScale` for any given `(outbound,inbound)` token pair.
- proxy contracts can manage the API complexity of having to specify price ticks

## Summary of packed offerList arguments
Methods such as `marketOrder`, don't start with `(address outbound, address inbound, uint tickScale, ...` but with `(OLKey memory olKey, ...`. OLKey stands for Offer List Key. The definition of `OLKey` is:

```solidity
struct OLKey {
  address outbound;
  address inbound;
  uint tickScale;
}
```
- Calls to Mangrove are much easier to read and write
- Fewer stack variables internally and for callers. 
- Small gas cost increase (~400).
- Prepares for calldata-reducing optimizations that will drastically reduce costs on L2s.

## Other changes
- Use of the `Market` struct (structu

## Questions
- [x] Should events log the `hash(olKey)` instead of `outbound,inbound,tickScale`? This reduces gas costs and means that for events that specify taker & maker, we can also index those fields (there is a maximum of 3 indexed fields per event). Indexed fields can be efficiently queried.
  - Yes, they should, and they do now.
- [ ] Names are hard. The structure for now is:

```solidity
struct OLKey {
  address outbound;
  address inbound;
  uint tickScale;
}

mapping(bytes32 => OfferList) internal offerLists;

struct OfferList {
  MgvStructs.LocalPacked local;
  mapping(uint => OfferData) offerData;
  mapping(int => Leaf) leafs;
  mapping(int => Field) level0;
  mapping(int => Field) level1;
}

struct OfferData {
  MgvStructs.OfferPacked offer;
  MgvStructs.OfferDetailPacked detail;
}
``` 

It seems slightly suboptimal but I don't know in which direction it should change. I would tend to go towards even simpler names as they are "base concepts" of Mangrove:
* `OLKey`->`Key`, 
* `olKey` -> `key`
* `OfferList` -> `List`
* `offerLists` -> `lists`
* `OfferData` -> `Offer`
* `MgvStructs.OfferPacked` -> `OfferCore`
* `MgvStructs.OfferDetailPacked` -> `OfferDetail`

[^1]: **assisted price ticks**: offerlists are _internally_ `(outbound,inbound,tickScale)`, but Mangrove has an internal notion of the “current” `tickScale`. The public-facing ABI only lets people give an `(outbound,inbound)` pair, so people can only interact with the current `tickScale` (except to remove offers that were created when a different `tickScale` was the default). When the `tickScale` is changed by governance, all current offers linger in a “zombie” state where they cannot be taken (but don’t pollute the OB) and can only be retracted.


lib/TickLib.sol
```diff
@@ -160,42 +160,41 @@ library LeafLib {
...
+  // FIXME should depend on the min and max prices 
+  int constant MIN_LOG_PRICE = -524287;
+  int constant MAX_LOG_PRICE = -MIN_LOG_PRICE;
 
-  function inRange(Tick tick) internal pure returns (bool) {
-    return Tick.unwrap(tick) >= MIN_TICK && Tick.unwrap(tick) <= MAX_TICK;
+  function inRange(int logPrice) internal pure returns (bool) {
+    return logPrice >= MIN_LOG_PRICE && logPrice <= MAX_LOG_PRICE;
+  }
+  function logPriceFromPrice_e18(uint price_e18) internal pure returns (int) {
```
lnist, line 175: e18 is still too little right? The price (with the previous code) went up to 10^22.

src/IMangrove.sol
```diff
@@ -8,33 +8,50 @@ pragma solidity >=0.7.0 <0.9.0;
...
   event Credit(address indexed maker, uint amount);
   event Debit(address indexed maker, uint amount);
   event Kill();
   event NewMgv();
   event OfferFail(
     address indexed outbound_tkn,
     address indexed inbound_tkn,
+    uint indexed tickScale,
     uint id,
     address taker,
```
lnist: this is an example where indexed taker would be nice (and maker for that matter - so with OLID we could do that.

src/MgvLib.sol
```diff
@@ -9,6 +9,26 @@ import {IERC20} from "./IERC20.sol";
...
+  uint tickScale;
+}
+
+library OLLib {
+  // The id should be keccak256(abi.encode(olKey))
+  // To save gas, id() directly hashes the memory (which matches the ABI encoding)
+  // If the memory layout changes, this function must be updated
+  function hash(OLKey memory olKey) internal pure returns (bytes32 _id) {
+    assembly ("memory-safe") {
+      _id := keccak256(olKey, 96)
```
lnist, line 27: Do we have a unit test for this?

src/MgvOfferMaking.sol
```diff
@@ -38,62 +37,56 @@ contract MgvOfferMaking is MgvHasOffers {
...
   */
-  function newOfferByVolume(
-    address outbound_tkn,
-    address inbound_tkn,
-    uint wants,
-    uint gives,
-    uint gasreq,
-    uint gasprice
-  ) external payable returns (uint) {
+  function newOfferByVolume(OLKey memory olKey, uint wants, uint gives, uint gasreq, uint gasprice)
```
lnist, line 44: Should it be OLKey calldata (multiple external calls)

test/core/DynamicTicks.t.sol
```diff
@@ -0,0 +1,130 @@
+// SPDX-License-Identifier:	AGPL-3.0
+
+pragma solidity ^0.8.10;
+
+import "mgv_test/lib/MangroveTest.sol";
+import {MgvStructs, MAX_TICK, MIN_TICK, LogPriceLib} from "mgv_src/MgvLib.sol";
+import {DensityLib} from "mgv_lib/DensityLib.sol";
+import {stdError} from "forge-std/stdError.sol";
```
lnist:
```suggestion
import {stdError} from "forge-std/StdError.sol";
```

src/IMangrove.sol
lnist: We should then emit an event when opening a market which has the out/in/scale and OfferListId hash for the indexer to know what its working with.

lib/TickLib.sol
```diff
@@ -267,18 +262,57 @@ library TickLib {
...
+}
+
+library TickLib {
+
+  function inRange(Tick tick) internal pure returns (bool) {
+    return Tick.unwrap(tick) >= MIN_TICK && Tick.unwrap(tick) <= MAX_TICK;
+  }
+
+  function fromLogPrice(int logPrice, uint tickScale) internal pure returns (Tick) {
+    // Do not force logPrices to fir the tickScale (aka logPrice%tickScale==0)
```
lnist: fir? fit?

lnist: Why don't we want exact math here? Would it surprise makers that we don't?

lib/TickLib.sol
```diff
@@ -160,42 +160,41 @@ library LeafLib {
 
 }
 
-library TickLib {
+library LogPriceLib {
   // Could have an INVALID tick that is > 24 bits? But in `local` how do I write "empty"?
   int constant BP = 1.0001 * 1e18;
```
lnist, line 165: does the change allow us to make BP smaller so we can support stablestable? The current change (as I understand it) allows higher prices to add support for more volatile unstableunstable. We discussed not putting more work into supporting stablestable, but what would the impact be if we decrease this number?

lib/TickLib.sol
lnist: With tickscale price can become much much higher, right?

src/MgvRoot.sol
```diff
@@ -35,55 +36,56 @@ contract MgvRoot is HasMgvEvents {
...
+    * `local` is the Mangrove configuration specific to the `outbound,inbound,tickScale` offerList. It contains e.g. the minimum offer `density`. It contains packed information, see [`structs.js`](#structs.js) for more.
     * `offerData` maps from offer ids to offer data.
   */
 
   /* Note that offers are structured into a tree with linked lists at its leves.
      The root is level2, has 256 level1 node children, each has 256 level0 node children, each has 256 leaves, each has 4 ticks (it holds the first and last offer of each tick's linked list).
      level2, level1 and level0 nodes are bitfield, a bit is set iff there is a tick set below them.
   */
-  struct Pair {
+  // FIXME rename OfferList to OLD (means OfferListData) or such
```
lnist, line 49: I must admit that the renaming from Pair to OfferList made it quite hard to find the important bits to review so I am not sure I saw it all :)

src/periphery/MgvReader.sol
```diff
@@ -365,75 +393,69 @@ contract MgvReader {
...
-
-  /// @param tkn0 one token of the market
-  /// @param tkn1 another token of the market
+  /// @param market the market
   /// @return bool Whether the {tkn0,tkn1} market is open.
   /// @dev May not reflect the true state of the market on Mangrove if `updateMarket` was not called recently enough.
-  function isMarketOpen(address tkn0, address tkn1) external view returns (bool) {
-    (tkn0, tkn1) = order(tkn0, tkn1);
-    return marketPositions[tkn0][tkn1] > 0;
+  function isMarketOpen(Market memory market) external view returns (bool) {
```
lnist, line 423: I seem to remember we allowed half-open markets. This seems to not allow it. What am I missing?

src/MgvOfferTaking.sol
```diff
@@ -104,56 +111,51 @@ abstract contract MgvOfferTaking is MgvHasOffers {
...
-  function generalMarketOrder(
-    address outbound_tkn,
-    address inbound_tkn,
-    Tick maxTick,
-    uint fillVolume,
-    bool fillWants,
-    address taker
-  ) internal returns (uint, uint, uint, uint) {
+
+  function generalMarketOrder(OLKey memory olKey, int logPrice, uint fillVolume, bool fillWants, address taker)
```
lnist: I had expected the `int logPrice` to still be `Tick maxTick`. I am not entirely sure whether this change is needed or good - it seems to introduce some imprecision and possibly some extra calculations.

I am not sure I've found all the places where tick<->logprice conversion happens :)

lib/TickLib.sol
adhusson: Yes this is all temporary pending a tick math pr :)

src/MgvLib.sol
adhusson: Yes! Well depends what you want to test but there is a test that checks it corresponds to the hash of the abi encoding

src/MgvOfferMaking.sol
adhusson: Not sure since hash is always taken it needs to be copied to memory anyway, no?

lib/BitLib.sol
```diff
@@ -2,12 +2,13 @@
 pragma solidity ^0.8.17;
 
 library BitLib {
+  // Returns the number of zeroes in x that do not have a 1 to their left, eg ctz(0)=256, ctz(1)=0
```
espendk: Shouldn't `ctz(1) = 255`? None of the zeroes in `0...01` have a 1 to their left?

lib/BitLib.sol
```diff
@@ -2,12 +2,13 @@
...
   function ctz(uint x) internal pure returns (uint c) {
     assembly ("memory-safe") {
       // Isolate the least significant bit
       x := and(x, add(not(x), 1))
 
-      // Get first 3 bits of c
+      // Get first 3 bits of c, this is the unusual part
```
espendk, line 11: "unusual" compared to what? Where does this come from? :-)

lib/TickLib.sol
espendk, line 1: > Yes this is all temporary pending a tick math pr :)

👍 I'll not review this in any detail then.

src/IMangrove.sol
espendk: Agreed. We've often discussed that it would be great to have more indexed fields in our events and this seems a good way to be able to do that.

src/MgvRoot.sol
espendk: If this `OfferList` struct gets renamed, then we could consider renaming `OLKey` to `OfferList` as I think that could make the public API easier to read. Examples:

- `function activate(OfferList memory offerList, uint fee, uint density, uint offer_gasbase) external;`
- `function marketOrderByPrice(OfferList memory offerList, uint maxPrice, uint fillVolume, bool fillWants)`

Technically, it's not the offer list itself but an ID/key. However, the intuition is clearer with this naming: I'm activating an offer list or running a market order on an offer list.

src/periphery/MgvReader.sol
```diff
@@ -1,19 +1,56 @@
...
+
+// flip tkn0/tkn1 of a market. Useful before conversion to OLKey
+// creates a copy
+function flipped(Market memory market) pure returns (Market memory) {
+  return Market(market.tkn1, market.tkn0, market.tickScale);
+}
+
+// convert Market to OLKey
+// creates a copy
+function toOLKey(Market memory market) pure returns (OLKey memory) {
```
espendk, line 47: It could be clearer that you'll get the tkn0/tkn1 OL, not tkn1/tkn0. Eg by changing the name or adding a parameter that allows the caller to specify which side of the market they want.

src/MgvOfferTakingWithPermit.sol
```diff
@@ -75,70 +75,53 @@ abstract contract MgvOfferTakingWithPermit is MgvOfferTaking {
...
+  {
     unchecked {
-      (successes, takerGot, takerGave, bounty, feePaid) =
-        generalSnipes(outbound_tkn, inbound_tkn, targets, fillWants, taker);
+      (successes, takerGot, takerGave, bounty, feePaid) = generalSnipes(olKey, targets, fillWants, taker);
       /* The sender's allowance is verified after the order complete so that the actual amounts are checked against the allowance, instead of the declared `takerGives`. The former may be lower.
     
     An immediate consequence is that any funds available to Mangrove through `approve` can be used to clean offers. After a `snipesFor` where all offers have failed, all token transfers have been reverted, so `takerGave=0` and the check will succeed -- but the sender will still have received the bounty of the failing offers. */
-      deductSenderAllowance(outbound_tkn, inbound_tkn, taker, takerGave);
+      deductSenderAllowance(olKey.outbound, olKey.inbound, taker, takerGave);
```
espendk: So allowance isn't tied to a tickScale. This seems reasonable. Would probably be good to update comments to explain this + the rationale for it.

src/MgvOfferTaking.sol
```diff
@@ -20,81 +32,76 @@ abstract contract MgvOfferTaking is MgvHasOffers {
     uint feePaid; // used globally
     Leaf leaf;
     Field level1;
-    Tick maxTick; // maxTick is the maximum tick that can be reached by the market order as a limit price.
+    int logPrice; // maxTick is the log of the max price that can be reached by the market order as a limit price.
```
```suggestion
    int logPrice; // logPrice is the log of the max price that can be reached by the market order as a limit price.
```

src/MgvOfferTaking.sol
```diff
@@ -20,81 +32,76 @@ abstract contract MgvOfferTaking is MgvHasOffers {
...
   }
 
   /* # Market Orders */
 
   /* ## Market Order */
   //+clear+
 
-  /* A market order specifies a (`outbound_tkn`,`inbound_tkn`) pair, a desired total amount of `outbound_tkn` (`takerWants`), and an available total amount of `inbound_tkn` (`takerGives`). It returns four `uint`s: the total amount of `outbound_tkn` received, the total amount of `inbound_tkn` spent, the penalty received by msg.sender (in wei), and the fee paid by the taker (in wei).
+  /* A market order specifies a (`outbound_tkn`,`inbound_tkn`,`tickScale`) offerList, a desired total amount of `outbound_tkn` (`takerWants`), and an available total amount of `inbound_tkn` (`takerGives`). It returns four `uint`s: the total amount of `outbound_tkn` received, the total amount of `inbound_tkn` spent, the penalty received by msg.sender (in wei), and the fee paid by the taker (in wei).
```
```suggestion
  /* A market order specifies a (`outbound_tkn`,`inbound_tkn`,`tickScale`) offer list, a desired total amount of `outbound_tkn` (`takerWants`), and an available total amount of `inbound_tkn` (`takerGives`). It returns four `uint`s: the total amount of `outbound_tkn` received, the total amount of `inbound_tkn` spent, the penalty received by msg.sender (in wei), and the fee paid by the taker (in wei).
```

src/MgvOfferTaking.sol
espendk: Agreed.

But if we retain log price, I think it should be called `maxLogPrice`, `limitLogPrice` or similar.
(I won't open the "limit order" vs "market order" debate here. Whoops, I just did 🫣)

src/MgvOfferMaking.sol
```diff
@@ -326,7 +304,8 @@ contract MgvOfferMaking is MgvHasOffers {
         }
       }
 
-      // mapping (uint => MgvStructs.OfferPacked) _offers = offers[ofp.outbound_tkn][ofp.inbound_tkn];
+      Tick insertionTick = TickLib.fromLogPrice(insertionLogPrice, ofp.olKey.tickScale);
```
espendk: It feels like we're converting back-and-forth between ticks and log prices. Maybe we should calculate both once and then drag them around in a struct?

espendk:
Names are indeed hard. As noted in a comment above, I think looking at the names from the perspective of the public API could perhaps guide us a bit.

For example, whenever the API needs to know which offer list to operate on, it takes an `OLKey`; I think `OfferList` would be more readable there.

So a proposal could be:
* `OLKey`->`OfferList`, 
* `olKey` -> `offerList`
* `OfferList` -> `OfferListData`
* `offerLists` -> `offerListsData`
* `OfferData` -> `OfferData`
* `MgvStructs.OfferPacked` -> `OfferCoreData`
* `MgvStructs.OfferDetailPacked` -> `OfferDetailData`

src/MgvOfferMaking.sol
lnist: Also if called directly with a hash from the outside? "OLID" approach.

lnist: I did not know you could not do hash of calldata.

src/IMangrove.sol
adhusson: ok, I pushed e97ca3cd12692968decf97902e12c2328547d72a :)

src/MgvOfferMaking.sol
adhusson: fixed by 86f994841ebd482bf95d391b2f2f4a5ce1edc086

lib/TickLib.sol
adhusson: Yeah it's not clear. They have to specify the `tickScale` so they should be aware. And tooling can choose whether to error or allow not-exactly-matching `logPrice`s. 

The scenario I have in mind: offer makers have a default "price computation" function. They can just call it and pass it without thinking about tickScale and get approximated away if they want to. Otherwise they'd have to call `logPrice/tickScale*tickScale` in addition.

WDYT?

lib/TickLib.sol
adhusson: Immediate impact is that total tick range must be increased to get the same price range as we have now, or we must introduce tick shift. Also future increased gas use when computing tickBase^logPrice (~a few hundred gas / tick I estimate)

src/MgvRoot.sol
adhusson: Well the FIXME is an old comment I meant to remove :p `OLKey` is really everywhere / pervasive, so a short name would be better. Also `key` I think is less error-inducing maybe? `Id` works as well. Just `Key` maybe?

src/periphery/MgvReader.sol
adhusson: `tkn0,tkn1` are canonically ordered so I think this does allow half-open markets?

src/MgvOfferTaking.sol
adhusson: @lnist the initial idea is that it is more intuitive to specify the log of the price rather than the index of the offer in the data structure that stores it. Before this PR, tick was always equal to logPrice, so there was no choice to make between the two.

@espendk now that `snipes` are gone, we might as well say `order`? The `marketOrder` function can definitely both work a (now true) limit order and as a market order.

lib/BitLib.sol
adhusson: english spec wrong `s/left/right`, fixed in c85c1e8b40fc1b19cdf37af1e0c4ada36a5b4c4c

lib/BitLib.sol
adhusson: There's a common trick documented in [this paper](http://supertech.csail.mit.edu/papers/debruijn.pdf) and [this page](https://graphics.stanford.edu/~seander/bithacks.html#IntegerLogDeBruijn) (also more on [stackoverflow](https://stackoverflow.com/questions/68677658/how-does-one-find-the-floor-of-the-log-base-2-of-an-n-bit-integer-using-bitwise) and [wikipedia](https://en.wikipedia.org/wiki/De_Bruijn_sequence)) to help you find the position of an isolated one in a bit sequence.

Starting from solady's solidity implementation of this idea (see below in the same file), I have found a trick to be cheaper with 256 words (normally you have to find the first 3 bits of the 1 position "by hand" then you can use deBruijn for the remaining 5 bits). I haven't yet written a complete explainer because I don't think we'll end up using it, but I'll write something up in the future. I don't think we'll use it because it seems like our bitmaps will be on 64 or 128 bits, in which case the method I use here is not the cheapest gaswise.

src/periphery/MgvReader.sol
adhusson: What about `toOLKey` + `toOLKeyFlipped`?

src/MgvOfferTakingWithPermit.sol
adhusson: agreed, it's a TODO

src/MgvRoot.sol
espendk: Alright. I think we're entering bike-shedding territory, especially re. length of names 😅
Re. just using `Key`: It's a generic name that doesn't convey what it relates to. So if the name should be ultra-short I think the current name or an acronym is better as it's less easy to misunderstand.

src/MgvOfferTaking.sol
espendk: > now that `snipes` are gone, we might as well say `order`? The `marketOrder` function can definitely both work a (now true) limit order and as a market order.

True. I think `order` would be clearer. And `limitOrder` even more so, since a DeFi market order in Mangrove will be modeled by adding slippage to the limit logPrice. But `order` could be a good compromise.

lib/BitLib.sol
espendk: Ok great. If we end up keeping (a variant of this) these reference will be great to have in the code.

src/periphery/MgvReader.sol
espendk: Since we're using "market" terminology, we could use "asks" and "bids", eg `asksToOLKey` or `toOLKeyAsks` ?

lib/TickLib.sol
lnist: I'm ok with it being external to the contract.

src/periphery/MgvReader.sol
lnist: Sorry I had forgotten what marketPositions contained :) you are right.

lnist (approved):

src/MgvOfferMaking.sol
adhusson: I don't think we're doing back-and-forth? Also logPrice <> tick conversion is extremely cheap.
1. insertionLogPrice is an argument
2. it is written as the `offerLogPrice` (got a bug to fix there, price should approximated)
3) it is converted to `tick` to be compared with `local.tick()` and used throughout the insertion algo

BTW, expanding our memory structs are costly --  e.g. with yesterday I was replacing `out,in,scale` with `key` in the events we emit, and I tried a) recompute `olKey.hash()` every time or b) add an `olKeyHash` field to the `SingleOrder` and `OfferPack` structs. The `market_order_8` gas test was 6k gas more expensive with b)!

adhusson:
> So a proposal could be:
> 
> OLKey->OfferList,
olKey -> offerList
OfferList -> OfferListData
offerLists -> offerListsData
OfferData -> OfferData
MgvStructs.OfferPacked -> OfferCoreData
MgvStructs.OfferDetailPacked -> OfferDetailData

I think they are way too long :p In my personal experience I have a much harder time committing APIs to memory when their most common names are like that; I also want to work with them less, and I also have a harder time rereading code I wrote that uses those API because the amount of line noise is high. 

Naming is definitely hard though and longer names are sometimes crucial to prevent confusion. Also `OLKey` is not good. Maybe `Market` after all, despite the name clash with the client API?

src/periphery/MgvReader.sol
adhusson: interesting -- internal to MgvReader we don't have a notion of quote/base inside a Market, just tkn0/tkn1. It's just supposed to represent "both offer lists" without the quote/base asymmetry. So maybe we should not use `Market` here actually. I think introducing asks/bids would create too much cognitive overhead here anyway, you'd end up wondering which one flips and which one doesn't.

I don't think it's that confusing as it is, but maybe `toOLKey01` and `toOLKey10` (as a reference to market tkn0/tkn1)?

Or `toOLKeyDirect`/`toOLKeyFlipped`

src/MgvOfferTaking.sol
adhusson: > But if we retain log price, I think it should be called maxLogPrice, limitLogPrice or similar.

Fixed in 6b1891bf75d66f6c11fbe4c42d0a93ee3e490dfa

src/MgvOfferMaking.sol
espendk: Understood 👍

espendk:
> Also `OLKey` is not good. Maybe `Market` after all, despite the name clash with the client API?

The circle is complete 😅 Let's maybe park it for a bit?

foundry.toml
```diff
@@ -7,12 +7,14 @@ test='test'
...
 cache_path='cache'
-fs_permissions = [{ access = "read-write", path = "./addresses/"}, { access = "read-write", path = "./analytics/"}, { access = "read", path = "./out/" }]
-solc_version="0.8.18"
+evm_version='paris'
+fs_permissions = [{ access = "read-write", path = "./addresses/"}, { access = "read", path = "./out/" }]
+solc_version="0.8.21"
 ffi=true
-# optimizer=false
-# optimizer_runs=200
+optimizer=false
```
espendk, line 14: Should this be commented out?

src/IMangrove.sol
espendk, line 1: Events here don't seem to have been updated to match MgvLib

test/core/Gatekeeping.t.sol
```diff
@@ -182,38 +182,38 @@ contract GatekeepingTest is MangroveTest {
     mkr.newOfferByVolume((1 << 96) + 1e28, 1 ether, 10_000, 0);
   }
 
-  function test_makerTick_wider_than_24_bits_fails_newOfferByTick() public {
-    vm.expectRevert("mgv/writeOffer/tick/outOfRange");
-    mkr.newOfferByTick(1 << 23, 1 ether, 10_000, 0);
+  function test_makerTick_wider_than_24_bits_fails_newOfferByLogPrice() public {
+    vm.expectRevert("mgv/writeOffer/logPrice/outOfRange");
+    mkr.newOfferByLogPrice(1 << 23, 1 ether, 10_000, 0);
```
espendk: Should logPrice fit in 24 bits?

I ask because the old test was too permissive: Ticks actually had to fit in 20 bits.

test/core/Gatekeeping.t.sol
```diff
@@ -309,6 +307,25 @@ contract GatekeepingTest is MangroveTest {
     tkr.marketOrder(0, 1 << 160);
   }
 
+  /*
+  // FIXME reactivate this test
```
espendk: There are tests of `clean` parameter validation tests in `TakerOperations.t.sol` , eg `test_gives_volume_above_96bits_fails_clean`.

So not sure these tests should be reactivated?

If they should, they should be placed together with those other clean parameter validation tests.

test/core/HigherTickBecomesBest.t.sol
espendk, line 1: Does this file serve a purpose?

test/core/TakerOperations.t.sol
```diff
@@ -724,9 +729,9 @@ contract TakerOperationsTest is MangroveTest {
     }
     mkr.expect("mgv/tradeSuccess");
     // first two offers are at right price
-    uint takerWants = 0.1 ether;
-    uint takerGives = 2 * takerWants;
-    mgv.marketOrderByVolume{gas: 450_000}($(base), $(quote), takerWants, takerGives, true);
+    uint takerWants = 0.1 ether + 0.1 ether;
+    uint takerGives = 0.1 ether + 0.2 ether;
```
espendk: This deviates from the pattern in the previous two tests.

I changed this because the old tests were unclear and one didn't actually make sense, I believe it was this one: The comment says the "first two offers are at right price" but that is not true now as evidenced by `offersExpectedTaken = 1` above.

I think this change should be reverted or it should be made clearer, what this test is supposed to test.

espendk (commented):
This review focuses on the changes to tests after merging `feat/ticks` into `feat/dynamic-ticks-packed`. I've ignored the gas tests as @lnist  have reviewed and adapted them.

foundry.toml
adhusson: It doesn't really matter for now anyway but yes :)

src/IMangrove.sol
adhusson: yes that's on my TODO but I think they were _already_ out of sync before :p

test/core/Gatekeeping.t.sol
adhusson: removed!

test/core/HigherTickBecomesBest.t.sol
adhusson: No I think it was a local file I had lying around

test/core/TakerOperations.t.sol
adhusson: addessed in cf54a2471e08dd93819e51782d782552aa7e1219

espendk (approved):

PR #445, opened 2023-08-14
feat(gas): Replace old gas tests with more comprehensive suite

lnist:
See test/core/gas/README.md

Adds a ./gas-measurement.sh for dumping measurements to out/gas-measurement.csv

Note: There are quite a bit of tests that inherit from other tests and override setup to change the test case. This is due to only setup being able to vary test data without warming stuff: https://github.com/foundry-rs/foundry/issues/1340 .


gas-measurement.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
```
espendk, line 1: Should this file be marked as executable in git?

espendk (approved):
Awesome!

Comprehension-wise it's quite dense (the inheritance structure complex, which I understand is needed), so perhaps an example of how to read one of the tests and its result could be helpful?


The CSV file could perhaps be made more amenable for analysis of the results.
Eg by using multiple columns/dimensions such that variations of the same test could easily be identified (eg a shared key) and compared (eg by having a column for the dimension set to `true` or by having a `dimensions` column).

This probably shouldn't be done as part of this PR as I think it's better to do it when we do some actual analyses and know what we need.

Relatedly, column headers would be great 😅

billede

gas-measurement.sh
lnist: Yes, thanks :)

lnist:
Added column headers. Let's revisit layout later.

PR #446, opened 2023-08-14
refactor: Simplify offer price match flow

espendk:
Previously, `execute` checked that the offer to be executed had a sufficiently low price. If this wasn't the case, `"mgv/notExecuted"` was returned and used to tell the next recursive call to `internalMarketOrder` that the order was finished.

This flow was somewhat convoluted: The decision on which offers to execute was distributed between multiple functions, despite only the outermost of those actually needing it. And because of this, the execute-offer path of `internalMarketOrder` had two if-stmts just to make sure it **did not** handle an offer that shouldn't be executed.

This commit instead moves the check out one level to `internalMarketOrder` and `clean`. This simplifies the code of all involved functions and reduces gas usage.

The tradeoff is that the price check code is duplicated and that the caller of `execute` must ensure that the taker is not matched with an offer that is too expensive. However, this is seems acceptable as the duplicated code is a trivial function call and `execute` is `internal`.

### DIFF
--- a/src/MgvOfferTaking.sol
+++ b/src/MgvOfferTaking.sol
@@ -174,7 +174,7 @@ abstract contract MgvOfferTaking is MgvHasOffers {
       emit OrderStart();
 
       /* Call recursive `internalMarketOrder` function.*/
-      internalMarketOrder(pair, mor, sor, true);
+      internalMarketOrder(pair, mor, sor);
 
       /* Over the course of the market order, a penalty reserved for `msg.sender` has accumulated in `mor.totalPenalty`. No actual transfers have occured yet -- all the ethers given by the makers as provision are owned by Mangrove. `sendPenalty` finally gives the accumulated penalty to `msg.sender`. */
       sendPenalty(mor.totalPenalty);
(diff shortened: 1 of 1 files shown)


src/MgvOfferTaking.sol
```diff
@@ -233,26 +228,18 @@ abstract contract MgvOfferTaking is MgvHasOffers {
...
+          3. `sor.gives` may have been clamped _down_ during `execute` (to "`offer.wants`" if the offer is entirely consumed, or to `makerWouldWant`, cf. code of `execute`).
         */
-          (sor.offerId, sor.local) = getNextBest(pair, mor, sor.offer, sor.local);
+        (sor.offerId, sor.local) = getNextBest(pair, mor, sor.offer, sor.local);
 
-          sor.offer = pair.offerData[sor.offerId].offer;
-        }
+        sor.offer = pair.offerData[sor.offerId].offer;
 
         /* note that internalMarketOrder may be called twice with same offerId, but in that case `proceed` will be false! */
```
lnist: update comment

espendk: removed in [c263db4](https://github.com/mangrovedao/mangrove-core/pull/446/commits/c263db4e477b936cc3db40f06b213bb91c413b5f)

src/MgvOfferTaking.sol
```diff
@@ -261,13 +247,11 @@ abstract contract MgvOfferTaking is MgvHasOffers {
...
-        /* After an offer execution, we may run callbacks and increase the total penalty. As that part is common to market orders and snipes, it lives in its own `postExecute` function. */
-        if (mgvData != "mgv/notExecuted") {
-          postExecute(mor, sor, gasused, makerData, mgvData);
-        }
+        /* After an offer execution, we may run callbacks and increase the total penalty. As that part is common to market orders and cleaning, it lives in its own `postExecute` function. */
+        postExecute(mor, sor, gasused, makerData, mgvData);
 
         /* #### Case 2 : End of market order */
-        /* If `proceed` is false, the taker has gotten its requested volume, or we have reached the end of the book, we conclude the market order. */
+        /* If `proceed` is false, the taker has gotten its requested volume, no more offers match, or we have reached the end of the book, we conclude the market order. */
```
jkrivine: Remove mention to `proceed` also line 211

jkrivine (changes requested):

src/MgvOfferTaking.sol
espendk: Addressed in [76f3afa](https://github.com/mangrovedao/mangrove-core/pull/446/commits/76f3afa0c8f42fd8459ad450a7b9651e7ccb1d1b)

adhusson:
@espendk I'm getting a feedback from @jkrivine saying that MgvReader does not do the price check in the market order!

espendk:
> @espendk I'm getting a feedback from @jkrivine saying that MgvReader does not do the price check in the market order!

Oh, there's a simulation of market order in `MgvReader` 😳 That'll need to be updated, but that's much broader than this PR. I suggest we add it to the `feat/ticks` task list.

adhusson:
Added to the TODO!

adhusson:
Added to the TODO!Added to the TODO!

adhusson (approved):
let's go

PR #453, opened 2023-08-16
fix: addresses from env file

peterMangrove:


README.md
```diff
@@ -80,6 +80,17 @@ To build, run
 $ yarn build
 ```
 
+## Addresses
+
+When writing scripts that uses the `Generic.sol` script, you can control what addresses are read As default it will try and look into the `{projectRoot}/addresses` folder, this can be disabled by setting `READ_ROOT_ADDRESSES` to false. If you want to read addresses from a different folder, you can set `READ_ADDRESSES_FROM` to the path of the folder you want to read from, the variable is a JSON string, here is an example:
```
espendk: I'd suggest giving the env var a prefix (eg `MGV_`) so we don't pollute the env var namespace.

Also, it looks like it's possible to point to multiple paths, so I suggest tweaking this text a bit:

```suggestion
When writing scripts that uses the `Generic.sol` script, you can control what addresses are read.

By default, it will try and look into the `{projectRoot}/addresses` folder, this can be disabled by setting `READ_ROOT_ADDRESSES` to false.

If you want to read addresses from other folders, you can set `READ_ADDRESSES_FROM` to the paths addresses should be read from. The variable is a JSON string, here is an example:
```

README.md
```diff
@@ -80,6 +80,17 @@ To build, run
...
+
+```shell
+export MGV_ADDRESSES_PATHS='{ "paths": ["/addresses/"] }'
+export READ_ROOT_ADDRESSES=false
```
espendk: I see there's partially a `MGV_` prefix here :-)

README.md
```diff
@@ -80,6 +80,17 @@ To build, run
...
+```
+
+In this example we disable the default path and set the path to `/addresses/`. Remember the path is relativ to the project root. This way you can easily read addresses from multiple sources.
```
espendk:
```suggestion
In this example we disable the default path and set the path to `/addresses/`. Remember the path is relative to the project root. This way you can easily read addresses from multiple sources.
```

espendk (dismissed):
Nifty! Added some minor comments, but lgtm!

README.md
peterMangrove: Fixed

espendk (approved):

PR #454, opened 2023-08-17
Migrate `snipes` and `snipesFor` in tests to `cleanByImpersonation` or `marketOrder*`

espendk:
This PR completes the replacement of `snipes` with `cleanByImpersonation` started in  #437 by migrating all tests away from `snipes` and `snipesFor`. 

TODO:
- [ ]  The main test in `Scenarii.t.sol` relied on snipe. It is not clear to me whether it's a test of sniping (and thus can be deleted) or if it also tests something else (and thus should be removed)? @jkrivine maybe knows?

### DIFF
--- a/test/core/Gatekeeping.t.sol
+++ b/test/core/Gatekeeping.t.sol
@@ -7,11 +7,12 @@ import {MgvStructs, MAX_TICK, MIN_TICK} from "mgv_src/MgvLib.sol";
 import {DensityLib} from "mgv_lib/DensityLib.sol";
 
 // In these tests, the testing contract is the market maker.
-contract GatekeepingTest is IMaker, MangroveTest {
+contract GatekeepingTest is MangroveTest {
   receive() external payable {}
 
   TestTaker tkr;
   TestMaker mkr;
+  TestMaker other_mkr;
   TestMaker dual_mkr;
   address notAdmin;
 
@@ -21,14 +22,17 @@ contract GatekeepingTest is IMaker, MangroveTest {
 
     tkr = setupTaker($(base), $(quote), "taker[$(A),$(B)]");
     mkr = setupMaker($(base), $(quote), "maker[$(A),$(B)]");
+    other_mkr = setupMaker($(base), $(quote), "other_maker[$(A),$(B)]");
     dual_mkr = setupMaker($(quote), $(base), "maker[$(B),$(A)]");
 
     mkr.provisionMgv(5 ether);
+    other_mkr.provisionMgv(5 ether);
     dual_mkr.provisionMgv(5 ether);
 
     deal($(quote), address(tkr), 1 ether);
     deal($(quote), address(mkr), 1 ether);
     deal($(base), address(dual_mkr), 1 ether);
+    deal($(quote), $(other_mkr), 1 ether);
 
     tkr.approveMgv(quote, 1 ether);
 
@@ -329,65 +333,51 @@ contract GatekeepingTest is IMaker, MangroveTest {
     );
   }
 
-  /* # Internal IMaker setup */
-
-  bytes trade_cb;
-  bytes posthook_cb;
-
-  // maker's trade fn for the mgv
-  function makerExecute(MgvLib.SingleOrder calldata) external override returns (bytes32 ret) {
-    ret; // silence unused function parameter
-    bool success;
-    if (trade_cb.length > 0) {
-      (success,) = $(this).call(trade_cb);
-      assertTrue(success, "makerExecute callback must work");
-    }
-    return "";
-  }
-
-  function makerPosthook(MgvLib.SingleOrder calldata order, MgvLib.OrderResult calldata result) external override {
-    bool success;
-    order; // silence compiler warning
-    if (posthook_cb.length > 0) {
-      (success,) = $(this).call(posthook_cb);
-      bool tradeResult = (result.mgvData == "mgv/tradeSuccess");
-      assertTrue(success == tradeResult, "makerPosthook callback must work");
-    }
-  }
-
   /* # Reentrancy */
 
   /* New Offer failure */
 
   function newOfferKO() external {
     vm.expectRevert("mgv/reentrancyLocked");
-    mgv.newOfferByVolume($(base), $(quote), 1 ether, 1 ether, 30_000, 0);
+    mkr.newOfferByVolume($(base), $(quote), 1 ether, 1 ether, 30_000);
   }
 
   function test_newOffer_on_reentrancy_fails() public {
-    mgv.newOfferByVolume($(base), $(quote), 1 ether, 1 ether, 100_000, 0);
-    trade_cb = abi.encodeCall(this.newOfferKO, ());
+    mkr.approveMgv(base, 1 ether);
+    deal($(base), $(mkr), 1 ether);
+
+    uint ofr = mkr.newOfferByVolume($(base), $(quote), 1 ether, 1 ether, 100_000);
+    mkr.setTradeCallback($(this), abi.encodeCall(this.newOfferKO, ()));
     assertTrue(tkr.marketOrderWithSuccess(1 ether), "take must succeed or test is void");
+    assertTrue(mkr.makerExecuteWasCalled(ofr), "ofr must be executed or test is void");
   }
 
   /* New Offer success */
 
   // ! may be called with inverted _base and _quote
   function newOfferOK(address _base, address _quote) external {
-    mgv.newOfferByVolume(_base, _quote, 1 ether, 1 ether, 30_000, 0);
+    mkr.newOfferByVolume(_base, _quote, 1 ether, 1 ether, 30_000);
   }
 
   function test_newOffer_on_reentrancy_succeeds() public {
-    mgv.newOfferByVolume($(base), $(quote), 1 ether, 1 ether, 200_000, 0);
-    trade_cb = abi.encodeCall(this.newOfferOK, ($(quote), $(base)));
+    mkr.approveMgv(base, 1 ether);
+    deal($(base), $(mkr), 1 ether);
+
+    uint ofr = mkr.newOfferByVolume($(base), $(quote), 1 ether, 1 ether, 200_000);
+    mkr.setTradeCallback($(this), abi.encodeCall(this.newOfferOK, ($(quote), $(base))));
     assertTrue(tkr.marketOrderWithSuccess(1 ether), "take must succeed or test is void");
+    assertTrue(mkr.makerExecuteWasCalled(ofr), "ofr must be executed or test is void");
     assertTrue(mgv.best($(quote), $(base)) == 1, "newOfferByVolume on swapped pair must work");
   }
 
   function test_newOffer_on_posthook_succeeds() public {
-    mgv.newOfferByVolume($(base), $(quote), 1 ether, 1 ether, 200_000, 0);
-    posthook_cb = abi.encodeCall(this.newOfferOK, ($(base), $(quote)));
+    mkr.approveMgv(base, 1 ether);
+    deal($(base), $(mkr), 1 ether);
+
+    uint ofr = mkr.newOfferByVolume($(base), $(quote), 1 ether, 1 ether, 200_000);
+    mkr.setPosthookCallback($(this), abi.encodeCall(this.newOfferOK, ($(base), $(quote))));
     assertTrue(tkr.marketOrderWithSuccess(1 ether), "take must succeed or test is void");
+    assertTrue(mkr.makerExecuteWasCalled(ofr), "ofr must be executed or test is void");
     assertTrue(mgv.best($(base), $(quote)) == 2, "newOfferByVolume on posthook must work");
   }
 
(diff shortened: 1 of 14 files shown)


test/core/Gatekeeping.t.sol
```diff
@@ -305,37 +305,10 @@ contract GatekeepingTest is IMaker, MangroveTest {
...
-    testMgv.snipesInTest($(base), $(quote), targets, true);
-  }
-
-  function test_wants_volume_above_96bits_fails_snipes() public {
-    uint ofr = mkr.newOfferByVolume(1 ether, 1 ether, 100_000);
-    uint[4][] memory targets = wrap_dynamic([ofr, 0, 1 << 96, type(uint).max]);
-    vm.expectRevert("mgv/snipes/volume/96bits");
-    testMgv.snipesInTest($(base), $(quote), targets, false);
-  }
-
```
adhusson, line 322: Is there a similar test for clean's max volume?

test/core/Gatekeeping.t.sol
```diff
@@ -305,37 +305,10 @@ contract GatekeepingTest is IMaker, MangroveTest {
...
-    deal($(base), address(mkr), 1 ether);
-    mkr.approveMgv(base, 1 ether);
-    uint ofr = mkr.newOfferByVolume(1 ether, 1 ether, 100_000);
-    Tick offerTick = pair.offers(ofr).tick();
-
-    vm.expectRevert("mgv/lowAllowance");
-    testMgv.snipesForInTest(
-      $(base), $(quote), wrap_dynamic([ofr, uint(Tick.unwrap(offerTick)), 1 ether, 300_000]), true, address(tkr)
-    );
-  }
```
adhusson, line 337: Is there a test that checks clean cannot decrease the `taker` balance?

test/core/Gatekeeping.t.sol
```diff
@@ -442,18 +415,18 @@ contract GatekeepingTest is IMaker, MangroveTest {
     uint other_ofr = mgv.newOfferByVolume($(quote), $(base), 1 ether, 1 ether, 100_000, 0);
 
     trade_cb = abi.encodeCall(this.updateOfferOK, ($(quote), $(base), other_ofr));
-    uint ofr = mgv.newOfferByVolume($(base), $(quote), 1 ether, 1 ether, 400_000, 0);
```
adhusson, line 445: Here and in the other tests where `x_on_y_succeeds|fails` I'm not seeing a check that the callbacks were indeed called. There wasn't a check before but it was very likely since the offer with the callback was being sniped. Here if there is a change to `setUp` in this contract or the parent contracts and a cheap offer is posted to the market and succeeds, then won't the tests become vacuously true?

test/core/MakerOperations.t.sol
```diff
@@ -83,8 +83,8 @@ contract MakerOperationsTest is MangroveTest, IMaker {
     bool funded;
     (funded,) = $(mgv).call{value: 1 ether}("");
     deal($(base), $(this), 1 ether);
-    uint ofr = mgv.newOfferByVolume($(base), $(quote), 0.05 ether, 0.05 ether, 200_000, 0);
-    require(tkr.take(ofr, 0.05 ether), "take must work or test is void");
+    mgv.newOfferByVolume($(base), $(quote), 0.05 ether, 0.05 ether, 200_000, 0);
+    require(tkr.marketOrderWithSuccess(0.05 ether), "take must work or test is void");
   }
```
adhusson: Similarly in this file I this it is less clear that the callback are being called. I think something like a `callbackWasCalled()` on each test would be enough, and `callbackWasCalled` checks that some state variable has been flipped from false to true?

edit: I see we do it in `MakerPosthook.t.sol`!

test/core/MakerPosthook.t.sol
```diff
@@ -422,8 +421,8 @@ contract MakerPosthookTest is MangroveTest, IMaker {
     _posthook = reverting_posthook;
```
adhusson, line 424: Except here -- we don't check the posthook was called

test/core/Scenarii.t.sol
```diff
@@ -273,62 +269,63 @@ contract ScenariiTest is MangroveTest {
...
-      balances.makersBalanceB[bag.snipedId]
-        + (bag.orderAmount * offers[bag.snipedId][Info.makerWants]) / offers[bag.snipedId][Info.makerGives],
-      relError(10),
-      "incorrect maker B balance"
-    );
-    // Testing residual offer
-    (MgvStructs.OfferUnpacked memory ofr,) = mgv.offerInfo($(base), $(quote), bag.snipedId);
-    assertTrue(ofr.gives == 0, "Offer should not have a residual");
-  }
+  // FIXME: Can this meaningfully be converted to clean and/or marketOrder?
```
adhusson, line 272: For this I think we need the advice of @jkrivine

test/core/TakerOperations.t.sol
```diff
@@ -77,99 +80,32 @@ contract TakerOperationsTest is MangroveTest {
...
-    quote.approve($(mgv), 1 ether);
-    (uint successes, uint got, uint gave,,) = testMgv.snipesInTest(
-      $(base), $(quote), wrap_dynamic([ofr, uint(Tick.unwrap(offerTick) - 1), 0.5 ether, 100_000]), false
-    );
-    assertTrue(successes == 0, "Snipe should fail");
-    assertEq(weiBalanceBefore, mgv.balanceOf($(this)), "Taker should not take bounty");
-    assertTrue((got == gave && gave == 0), "Taker should not give or take anything");
-  }
-
+  // FIXME: Not sure what this tests?
```
adhusson: The purpose of this test is to make sure inbound volumes are rounded up when partially taking an offer i.e. you can't have the taker pay 0 if the maker sends > 0 to the taker.

The test sets this up with a wants=9, gives=10 offer,and the taker asks for a volume of 1.

test/core/TakerOperations.t.sol
```diff
@@ -77,99 +80,32 @@ contract TakerOperationsTest is MangroveTest {
...
-    (uint successes, uint got, uint gave,,) = testMgv.snipesInTest(
-      $(base), $(quote), wrap_dynamic([ofr, uint(Tick.unwrap(offerTick) + 1), 0.5 ether, 100_000]), true
-    );
-    assertTrue(successes == 1, "Snipe should not fail");
+    (uint got, uint gave,,) = mgv.marketOrderByTick($(base), $(quote), Tick.unwrap(offerTick) + 1, 0.5 ether, true);
     assertEq(got, 0.5 ether, "Taker did not get correct amount");
     assertEq(gave, 0.5 ether, "Taker did not give correct amount");
   }
 
-  function test_multiple_snipes_fillWants() public {
```
adhusson, line 121: Is there a `clean` version of this test that should exist somewhere?

test/core/TakerOperations.t.sol
```diff
@@ -197,54 +130,86 @@ contract TakerOperationsTest is MangroveTest {
...
-    );
-    assertTrue(successes == 1, "Snipe should not fail");
+    Tick orderTick = TickLib.tickFromPrice_e18(1 ether);
+    (uint got, uint gave,,) = mgv.marketOrderByTick($(base), $(quote), Tick.unwrap(orderTick), 0.1 ether, false);
     assertApproxEqRel(got, 1 ether, relError(10), "Wrong got value");
     assertApproxEqRel(gave, 0.01 ether, relError(10), "Wrong gave value");
     assertTrue(!mgv.offers($(base), $(quote), ofr).isLive(), "Offer should not be in the book");
   }
 
-  function test_snipe_fillGives_zero() public {
```
adhusson, line 211: Similarly, shouldn't we test something similar with clean (ie that cleaning fails in this setup)?

test/core/TakerOperations.t.sol
```diff
@@ -297,14 +280,15 @@ contract TakerOperationsTest is MangroveTest {
 
     expectFrom($(mgv));
     emit OfferFail($(base), $(quote), ofr, $(this), 1 ether, 1 ether, "mgv/makerTransferFail");
-    (uint successes, uint takerGot, uint takerGave,,) =
-      testMgv.snipesInTest($(base), $(quote), wrap_dynamic([ofr, uint(Tick.unwrap(offerTick)), 1 ether, 100_000]), true);
+    vm.expectEmit(true, false, false, false, $(mgv));
+    emit Credit($(refusemkr), 0); // credited amount should be "mkr_provision - penalty"; Not including this in the emit assertion, as penalty is difficult to maintain.
```
adhusson: I don't understand this comment

test/core/TakerOperations.t.sol
```diff
@@ -330,18 +314,17 @@ contract TakerOperationsTest is MangroveTest {
 
     expectFrom($(mgv));
     emit OfferFail($(base), $(quote), ofr, $(this), 1 ether, 1 ether, "mgv/makerTransferFail");
-    (uint successes, uint takerGot, uint takerGave,,) =
-      testMgv.snipesInTest($(base), $(quote), wrap_dynamic([ofr, uint(Tick.unwrap(offerTick)), 1 ether, 100_000]), true);
+    vm.expectEmit(true, false, false, false, $(mgv));
+    emit Credit($(mkr), 0); // credited amount should be "mkr_provision - penalty"; Not including this in the emit assertion, as penalty is difficult to maintain.
```
adhusson: Same as above!

test/core/TakerOperations.t.sol
```diff
@@ -641,37 +593,15 @@ contract TakerOperationsTest is MangroveTest {
...
-    Tick offerTick = pair.offers(ofr).tick();
-    quote.approve($(mgv), 0.5 ether);
-
-    (uint successes,,,,) = testMgv.snipesInTest(
-      $(base), $(quote), wrap_dynamic([ofr, uint(Tick.unwrap(offerTick) - 1), 1 ether, 100_000]), true
-    );
-    assertTrue(successes == 0, "Order should fail when order price is higher than offer");
-  }
-
-  function test_snipe_on_higher_gas_fails() public {
```
adhusson, line 658: Do we have tests for clean on that topic?

test/core/TakerOperations.t.sol
espendk: Ideally, we'd include the credited value in the emitted event. However, the penalty isn't something we can calculate (since it depends on actual gas usage) and I think it's likely to change over time, so putting in a constant would require maintenance.

Do you think we should something different?

test/core/Gatekeeping.t.sol
espendk: Yes, in `TakerOperations.t.sol` there's `test_gives_volume_above_96bits_fails_clean` and `test_gives_tick_outside_range_fails_clean`.

test/core/Gatekeeping.t.sol
espendk: In  `TakerOperations.t.sol` there's `test_cleaning_by_impersonation_succeeds_and_does_not_transfer_funds`.

Do you think we need more/smth different?

test/core/Gatekeeping.t.sol
espendk: The assertion two lines down (L448) is checking that the gasreq of the offer posted in line 442 has been changed as defined in the posthook. Both numbers are defined in the test, are unaffected by `setUp`, and (I believe) can only be changed by the posthook (the offer id is passed from the test). So it should be ok?

espendk: Thinking some more on this: I think was we need is an assertion that the expected offer was executed since it's not guaranteed by `marketOrder` whereas it was with `snipe`. The tests in `Posthook.t.sol` should allow us to assume that the posthook is called if the offer is executed.

test/core/TakerOperations.t.sol
espendk: Yes, `test_clean_multiple_failing_offers` in the same file.

test/core/TakerOperations.t.sol
espendk: The test `test_cleaning_non_failing_offer_fails` tests the same, just in a slightly different way (it was written before migrating the snipes tests).

test/core/TakerOperations.t.sol
espendk: Same answer/question as above :-)

test/core/TakerOperations.t.sol
espendk: Yes, `test_giving_smaller_gasreq_to_clean_fails` in the same file.

adhusson: Sorry I should have checked better that all those tests I asked about had cleaning counterparts!

espendk: np :-)

test/core/Gatekeeping.t.sol
espendk: I'm implementing this and it's cleaner + makes things easier to read

espendk: Addressed in [6d1388d](https://github.com/mangrovedao/mangrove-core/pull/454/commits/6d1388d7700b735d4d362afbe231ab09109b62bd)

test/core/MakerOperations.t.sol
espendk: addressed in [6d1388d](https://github.com/mangrovedao/mangrove-core/pull/454/commits/6d1388d7700b735d4d362afbe231ab09109b62bd)

test/core/MakerPosthook.t.sol
espendk: Not sure how we can check it, since the posthook reverts and thus we cannot retain any state change made in it?

test/core/TakerOperations.t.sol
espendk: I found this pattern in `MakerPosthook.t.sol`, I'll copy that instead:

```
    //TODO: when events can be checked instead of expected, take given penalty instead of ignoring it
    vm.expectEmit(true, true, true, false, $(mgv));
```

espendk: Addressed in [a21d89d](https://github.com/mangrovedao/mangrove-core/pull/454/commits/a21d89dbc0f556eda8cdea09e69e21f89da9b907)

test/core/TakerOperations.t.sol
espendk: Addressed in [a21d89d](https://github.com/mangrovedao/mangrove-core/pull/454/commits/a21d89dbc0f556eda8cdea09e69e21f89da9b907)

test/core/TakerOperations.t.sol
espendk: Ok, thanks for the explanation :-) I've copied it into a comment, see [8f2c2cb](https://github.com/mangrovedao/mangrove-core/pull/454/commits/8f2c2cb6413e55228132de90f317cde596c0fde1)

test/core/MakerPosthook.t.sol
adhusson: I tried to check Mangrove's corresponding log: 3994d96ff8b77f2486f322254aa3c1860b4ad9a5

adhusson (approved):
pending my small test-editing commit lgtm

espendk:
> pending my small test-editing commit lgtm

Perfect, thanks.

I'll merge once @jkrivine has had a chance to weigh in re. Scenarii.t.sol.

espendk:
I merged the PR with the FIXME re. Scenarii in as @jkrivine  is busy and we need to move on :-)
We'll address the Scenarii test later - I'll add a task on it to the ticks PR.

PR #458, opened 2023-08-19
fix: correct leaf and best.prev after partially consumed tick

adhusson:
fix bug found by @espendk 

test checks correctness under a range of conditions (marketORder starts at tick or before tick, leave 1 or >1 offers in tick)

### DIFF
--- a/test/core/TakerOperations.t.sol
+++ b/test/core/TakerOperations.t.sol
@@ -1046,7 +1046,6 @@ contract TakerOperationsTest is MangroveTest {
     assertEq(best.prev(), 0, "best.prev should be 0");
     Leaf emptyLeaf = leaf.setTickFirst(tick, 0).setTickLast(tick, 0);
     assertTrue(emptyLeaf.isEmpty(), "leaf should not have other tick used");
-    // console.log(toString(pair.leafs(leafIndex)));
   }
 }
 
--- a/test/core/TakerOperations.t.sol
+++ b/test/core/TakerOperations.t.sol
@@ -3,7 +3,7 @@
 pragma solidity ^0.8.10;
 
 import "mgv_test/lib/MangroveTest.sol";
-import {MAX_TICK} from "mgv_lib/TickLib.sol";
+import {MAX_TICK, MIN_TICK} from "mgv_lib/TickLib.sol";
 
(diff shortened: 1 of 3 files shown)


test/core/TakerOperations.t.sol
```diff
@@ -1019,6 +1019,35 @@ contract TakerOperationsTest is MangroveTest {
...
+
+    Tick tick = Tick.wrap(_tick);
+
+    uint bestId = leaveOneOnly ? ofr4 : ofr3;
+    MgvStructs.OfferPacked best = pair.offers(bestId);
+    Leaf leaf = pair.leafs(best.tick().leafIndex());
+    assertEq(leaf.firstOfIndex(tick.posInLeaf()), bestId, "wrong first of tick");
+    assertEq(leaf.lastOfIndex(tick.posInLeaf()), ofr4, "wrong last of tick");
+    assertEq(pair.local().tickPosInLeaf(), tick.posInLeaf(), "wrong local.tickPosInleaf");
+    assertEq(best.prev(), 0, "best.prev should be 0");
```
espendk, line 1046: These assertions seem like something we'd like in more places ("assert that this is the best tick and that the first and last offers in it are X and Y"), so maybe we could make it an `assert*` function?

test/core/TakerOperations.t.sol
```diff
@@ -1019,6 +1019,35 @@ contract TakerOperationsTest is MangroveTest {
...
+    Leaf emptyLeaf = leaf.setTickFirst(tick, 0).setTickLast(tick, 0);
+    assertTrue(emptyLeaf.isEmpty(), "leaf should have not other tick used");
```
espendk:
```suggestion
    assertTrue(emptyLeaf.isEmpty(), "leaf should not have other tick used");
```

test/core/TakerOperations.t.sol
```diff
@@ -1019,6 +1019,35 @@ contract TakerOperationsTest is MangroveTest {
...
+    // console.log(toString(pair.leafs(leafIndex)));
```
espendk:
```suggestion
```

espendk (approved):
lgtm. Added some minor suggestions

PR #459, opened 2023-08-19
POC of minimal events

espendk:
POC of minimal events for market orders.

Below is the output of `forge snapshot --diff` comparing `feat/ticks` to this PR.

The successful `*_mo_*` tests (which are probably the most interesting) seem to save 0-3k gas.

```
test_basic_newOffer_sets_best() (gas: 0 (0.000%)) 
test_density_convert_auto(uint128) (gas: 0 (0.000%)) 
test_density_manual() (gas: 0 (0.000%)) 
test_density_multiply_auto(uint8,uint8,uint96) (gas: 0 (0.000%)) 
test_fixedFromParams() (gas: 0 (0.000%)) 
test_multiply_manual() (gas: 0 (0.000%)) 
test_activation_emits_events_in_order() (gas: 0 (0.000%)) 
test_admin_can_withdrawERC20(uint256) (gas: 0 (0.000%)) 
test_clean_on_closed_fails() (gas: 0 (0.000%)) 
test_clean_on_inactive_fails() (gas: 0 (0.000%)) 
test_configInfo(address,address,address,uint128) (gas: 0 (0.000%)) 
test_gives_0_rejected() (gas: 0 (0.000%)) 
test_gov_can_transfer_rights() (gas: 0 (0.000%)) 
test_gov_cant_be_zero() (gas: 0 (0.000%)) 
test_idOverflow_reverts(address,address) (gas: 0 (0.000%)) 
test_initial_allowance_is_zero() (gas: 0 (0.000%)) 
test_kill_is_idempotent() (gas: 0 (0.000%)) 
test_killing_updates_config() (gas: 0 (0.000%)) 
test_makerGasprice_wider_than_16_bits_fails_newOfferByVolume() (gas: 0 (0.000%)) 
test_makerGasreq_at_density_suceeds() (gas: 0 (0.000%)) 
test_makerGasreq_at_gasmax_succeeds_newOfferByVolume() (gas: 0 (0.000%)) 
test_makerGasreq_bigger_than_gasmax_fails_newOfferByVolume() (gas: 0 (0.000%)) 
test_makerGasreq_lower_than_density_fails_newOfferByVolume() (gas: 0 (0.000%)) 
test_makerGasreq_wider_than_24_bits_fails_newOfferByVolume() (gas: 0 (0.000%)) 
test_makerGives_wider_than_96_bits_fails_newOfferByVolume() (gas: 0 (0.000%)) 
[...]

PR #460, opened 2023-08-21
POC of minimal events to support indexer and RPC calls

espendk:
This builds on PR #459 but adding a bit more information, such that consumers can make sense of the events with an indexer or by using RPC calls.

The changes mostly revolves around adding indexed to fields and adding a few fields, in order for RPC calls to work. 
Only 2 events have been added. `CleanStart` and `CleanComplete`, before `OrderStart` and `OrderComplete` was used. But Clean and MarketOrder are 2 different operations and do not hold the same information, because of this these 2 new events have been added.

The biggest improvement is probably the Events.md file, that explains every single event we emit and why.


src/IMangrove.sol
```diff
@@ -12,29 +12,23 @@ import {MgvLib, MgvStructs, IMaker, OLKey} from "./MgvLib.sol";
...
 interface IMangrove {
-  event Approval(
-    address indexed outbound_tkn,
-    address indexed inbound_tkn,
-    uint indexed tickScale,
-    address owner,
-    address spender,
-    uint value
-  );
+  event Approval(address indexed outbound_tkn, address indexed inbound_tkn, address owner, address spender, uint value);
```
peterMangrove: Should we use olKey here?

espendk: Good question. Are approvals per pair or per OLKey? If it's per pair, then we probably shouldn't use olKey here

peterMangrove: What is the OLKey created from? only outbound and inbound? or does it also have other information?

adhusson: they are per pair!

espendk: > What is the OLKey created from? only outbound and inbound? or does it also have other information?

It also has the `tickScale`:

```Solidity
struct OLKey {
  address outbound;
  address inbound;
  uint tickScale;
}
```

Events.md
```diff
@@ -0,0 +1,225 @@
...
+If the posthook of the offer fails. Then we emit `OfferSuccessWithPosthookData` instead of just `OfferSuccess`. This event has one extra field, which is the reason for the posthook failure. By emitting the posthook data, an indexer can keep track of the reason posthook fails, this could for example be used for analytics.
+
+By emitting `offerId`, `wants` and `gives`, an indexer can keep track of whether the offer was partially or fully taken, by looking at the what the offer was posted at.
+
+### OrderComplete
+
+This event is emitted when a market order is finished.
+
+It only emits the total `fee paid`. We need this event, in order to know that a market order is completed. This way an indexer can know exactly in what context we are in.
+The total `fee paid` for that market order, is needed, as we do not emit this any other place. It is not indexed as there is no need for RPC calls to filter on this.
```
espendk: Perhaps this event should have the same indexed fields as `OrderStart` so RPC clients can filter on this and get the complete picture for the order?

espendk (commented):
Great stuff! Added a single question :-)

Events.md
peterMangrove: Might not be a bad idea. @adhusson @lnist   WDYT?

Events.md
```diff
@@ -0,0 +1,225 @@
...
+The `maker` address is indexed so that we can filter on it when doing RPC calls.
+
+These are the scenarios where it can happen:
+
+- Fund Mangrove directly
+- Fund Mangrove when posting an offer
+- When updating an offer
+  - Funding Mangrove or
+  - the updated offer needs less provision, than it already has. Meaning the user gets credited the difference.
+- When retracting an offer and deprovisioning it.Meaning the user gets credited the provision that was locked by the offer.
```
lnist:
```suggestion
- When retracting an offer and deprovisioning it. Meaning the user gets credited the provision that was locked by the offer.
```

Events.md
```diff
@@ -0,0 +1,225 @@
...
+- When an offer fails. The remaining provision gets credited back to the maker
+
+A challenge for an indexer is to know how much provision each offer locks. With the current events, an indexer is going to have to know the liveness, gasreq and gasprice of the offer. If the offer is not live, then also know if it has been deprovisioned. And know what the gasbase of the offerlist was when the offer was posted. With this information an indexer can calculate the exact provision locked by the offer.
+
+Another information that an indexer cannot deduce, is in what scenario the credit event happened. E.g. We don't know if the credit event happened because an offer failed or if the user simply funded Mangrove.
```
lnist:
```suggestion
The indexer also cannot deduce what scenario the credit event happened. E.g., we don't know if the credit event happened because an offer failed or because the user simply funded Mangrove.
```

Events.md
```diff
@@ -0,0 +1,225 @@
...
+
+### NewMgv
+
+Emitted when a new Mangrove is deployed.
+
+### OrderStart
+
+This event is emitted when a market order is started on Mangrove.
+
+It emits the `offerlist key`, the `taker`, the `maxLogPrice`, `fillVolume` and `fillWants`.
```
lnist: offerlist => offerList in general - and it is the hash that gets emitted, right?

Events.md
```diff
@@ -0,0 +1,225 @@
...
+
+The fields `offerlist key` and `taker` are indexed, so that we can filter on them when doing RPC calls.
+
+By emitting this an indexer can keep track of what context the current market order is in. E.g. If a user starts a market order and one of the offers taken also starts a market order, then we can in an indexer have a stack of started market orders and thereby know exactly what offerlist the order is running on and the taker.
```
lnist: In general, E.g. should not be followed by capitalized letter.

Events.md
```diff
@@ -0,0 +1,225 @@
...
+
+It emits `offerlist key`, `taker`, `maker`, the `offerId`, the offers `wants`, `gives`, `penalty` and the `reason` for failure. `Offer list key`, `taker` and `maker` are all fields that we do not need, in order for an indexer to work, as an indexer will be able the get that info from the former `OrderStart` and `OfferWrite` events. But in order for RPC call to filter on this, we need to emit them. Those 3 fields are indexed for the same reason.
+
+If the posthook of the offer fails. Then we emit `OfferFailWithPosthookData` instead of just `OfferFail`. This event has one extra field, which is the reason for the posthook failure. By emitting the posthook data, an indexer can keep track of the reason posthook fails, this could for example be used for analytics.
+
+This event is emitted doring posthook end, we wait to emit this event to the end, because we need the information of `penalty`, which is only available at the end of the posthook. This means that `OfferFail` events are emitted in reverse order, compared to what order they are taken. This is do to the way we handle posthooks. The same goes for `OfferSuccess`.
```
lnist:
```suggestion
This event is emitted doring posthook end, we wait to emit this event to the end, because we need the information of `penalty`, which is only available at the end of the posthook. This means that `OfferFail` events are emitted in reverse order, compared to what order they are taken. This is due to the way we handle posthooks. The same goes for `OfferSuccess`.
```

Events.md
lnist: I agree

Events.md
```diff
@@ -0,0 +1,225 @@
...
+
+The fields `offerlist key` and `taker` are indexed, so that we can filter on them when doing RPC calls.
+
+### CleanComplete
+
+This event is emitted when a Clean operation is completed.
+
+It does not emit any fields. This event is only needed in order to know that the clean operation is completed. This way an indexer can know exactly in what context we are in. It could emit the total bounty received, but in order to know who got the bounty, an indexer would still be needed or we would have to emit the taker address aswell, in order for an RPC call to find the data. But an indexer would still be able to find this info, buy collecting all the previous events. So we do not emit the bounty.
```
lnist:
```suggestion
It does not emit any fields. This event is only needed in order to know that the clean operation is completed. This way an indexer can know exactly in what context we are in. It could emit the total bounty received, but in order to know who got the bounty, an indexer would still be needed or we would have to emit the taker address as well, in order for an RPC call to find the data. But an indexer would still be able to find this info, by collecting all the previous events. So we do not emit the bounty.
```

Events.md
```diff
@@ -0,0 +1,225 @@
...
+We tried to list the main points that we would like to do with events.
+
+1. Use as little gas as possible
+2. An indexer should be able to keep track of the state of Mangrove.
+3. Doing RPC calls directly, should be able to find offers and other information based on offerlist, maker, taker, offer id, etc.
+
+These 3 points all have there own direction and it is therefore not possible to find a solution that is optimal for all 3 points.
+We have therefore tried to find a solution that is a good balance between the all 3 points.
+
+## Events
```
lnist: IMHO all the event-specific documentation should be inlined where the events are defined. @adhusson wdyt?

src/MgvLib.sol
```diff
@@ -138,6 +139,8 @@ interface HasMgvEvents {
 
   event OfferFailWithPosthookData(
     bytes32 indexed olKeyHash,
+    address indexed taker,
```
lnist, line 350: If taker/maker is deemed too much data, we can consider removing maker and having id be indexed - maker would have to make a query for each of their offers, but they could get the data.
Maybe you can use Espen's newly added gas diff feature and see what the effect of these changes are :)

lnist: Though do note that in the past we emitted outbound/inbound indexed, so this is not much more.
One could argue that for something like MangroveOrder indexed offerId could make more sense to query on for the UI.

lnist: This may be the case for most events emitting maker an id. Maybe @adhusson or @espendk has an opinion.

lnist (commented):
Added some typos comments to the doc but did not in general read it for proof-reading, only the stuff I stumbled on.

src/MgvLib.sol
espendk: Would it be cheaper to emit `id` instead of `maker`?

For `MangroveOrder`, I don't think it'll make much difference: Yes, the UI will receive events for offer ids that don't belong to the current user, but it'll be easy to filter on and the number of events will likely not be very high.

Events.md
peterMangrove: Fixed

Events.md
peterMangrove: Fixed

Events.md
peterMangrove: I've implemented this now.

Events.md
peterMangrove: All the documentation has been inlined at their relevant event.

src/MgvLib.sol
peterMangrove: After a talk with Lasse. I then agree that by indexing the id, it will be easier for RPC call filter correctly. As the maker could be a Strat, like MangroveOrder, then it is not really as relevant as the actual id. This way the Strat can emit the relation between the maker and the offerId. It is then doable to do RPC calls and filter on the relevant ids.

lnist (approved):

PR #461, opened 2023-08-21
Feat/theclog

lnist:
Introduces
* TheClog.t.sol to test out of gas and stack too deep clogs due to failing offers.
* MaxDepth in global to configure a maxDepth (Consider name, and should it be compile time constant?)
* Accounts for needed gas for posthooks and verifies that we have enough for execute+posthook or else ends market order
* Factors out the end of market order case into its own function for readability (this is the final commit)

Since this manipulates the same code as feat/simplify-price-match-flow that branch was merged in, do not consider those changes part of this PR. The final commit and the merge makes this hard to review. The most important changes are:
1) Initialization of maxDepth and gasreqLeft in generalMarketOrder
2) Using the values and updating them in internalMarketOrder

The change unclogs at the cost of some accounting, currently taking up some bits for dynamic configuration, and two additional uints in the mor structure.


src/MgvOfferTaking.sol
```diff
@@ -149,6 +151,10 @@ abstract contract MgvOfferTaking is MgvHasOffers {
       sor.inbound_tkn = inbound_tkn;
       Pair storage pair;
       (sor.global, sor.local, pair) = _config(outbound_tkn, inbound_tkn);
+      mor.maxDepth = sor.global.maxDepth();
+      /* We set aside gas for ending a market order. */
+      mor.gasreqLeft = sor.local.offer_gasbase();
```
espendk: This is an over-approximation, right?

src/MgvOfferTaking.sol
```diff
@@ -211,108 +227,96 @@ abstract contract MgvOfferTaking is MgvHasOffers {
...
-        /* If an execution was attempted, we move `sor` to the next offer. Note that the current state is inconsistent, since we have not yet updated `sor.offerDetails`. */
-        if (mgvData != "mgv/notExecuted") {
-          /* It is known statically that `mor.initialGives - mor.totalGave` does not underflow since
-           1. `mor.totalGave` was increased by `sor.gives` during `execute`,
-           2. `sor.gives` was at most `mor.initialGives - mor.totalGave` from earlier step,
-           3. `sor.gives` may have been clamped _down_ during `execute` (to "`offer.wants`" if the offer is entirely consumed, or to `makerWouldWant`, cf. code of `execute`).
-        */
-          (sor.offerId, sor.local) = getNextBest(pair, mor, sor.offer, sor.local);
+      sor.offer = pair.offerData[sor.offerId].offer;
+      mor.gasreqLeft -= gasused;
```
espendk: Here we haven't used all of the `offer_gasbase` gas set aside in line 209 since we're not done processing the offer yet. So I suspect this is an under approximation.

But maybe this is compensated for in line 156 where we set `offer_gasbase` aside for ending the order?

src/MgvOfferTaking.sol
```diff
@@ -188,21 +194,31 @@ abstract contract MgvOfferTaking is MgvHasOffers {
...
+      }
+
+      /* Load additional information about the offer. We don't do it earlier to save one storage read in case `proceed` was false. */
+      sor.offerDetail = pair.offerData[sor.offerId].detail;
+
+      mor.gasreqLeft += sor.offerDetail.gasreq();
+
+      if (gasleft() < mor.gasreqLeft + sor.local.offer_gasbase()) {
+        return endInternalMarketOrder(pair, mor, sor);
+      }
```
espendk: I like how this simplifies this method and makes it clear what the end-conditions are and how they are handled.

Comments need to be updated, but we can do that when/if we decide to move forward with this.

src/MgvGovernable.sol
```diff
@@ -17,6 +17,7 @@ contract MgvGovernable is MgvRoot {
       /* Set initial gasprice and gasmax. */
       setGasprice(_gasprice);
       setGasmax(gasmax);
+      setMaxDepth(40);
```
espendk: Two comments:

1. Perhaps we should name this parameter in a more semantic way. Eg `maxOfferTakeCount`, `maxTakenOffers` or smth. Or perhaps we should say what 'depth' this is referring to.

espendk (commented):
I like this. It's annoying that we have to do it, but as you demonstrate, The Clog is real and IMO we should address it.

Have you measured the impact of gas usage by this change?

src/MgvGovernable.sol
lnist: Agree - @adhusson : Do you have any preferred name base on Espen's suggestions?

src/MgvOfferTaking.sol
lnist: You're on the right track. The reservation in 156 is needed because 258 is an underapproximation. It goes like this when taking two offers:
1. Line 258 reached first time, we have at least `offer_gasbase+gasreq1-gasused1` gas left for rounding off the market order due to what is set aside in 156 and the accounting in internalMarketOrder. It is larger due to line 209 also conditioning on additional offer_gasbase.
2. Line 258 reached the second time we have gasreqLeft be `offer_gasbase +gasreq1-gasused1+gasreq2-gasused2`, however, note that we will not execute offer2 unless gasleft() can also cover an additional offer_gasbase when reaching line 209. 

So we will never start executing an offer unless we also have `offer_gasbase` left on top of the `gasreq` . This should always be sufficient to complete a market order of 1 offer. For many offers this code relies on offer_gasbase being able to cover for the overhead of unwrapping the stack and doing the necessary updates.
Alternatively we should add some constant overhead for each offer on line 258 or let the reservation on line 156 be another value which we know is big enough to cover the unwrap of stack.

Note: gasreq for one offer is the worst case for that offer, so with a set of offers each recursion will set aside too much in case worst case is not hit, so on average a bit extra is set aside for each offer.

src/MgvOfferTaking.sol
lnist: See other comment.

src/MgvOfferTaking.sol
adhusson: Later when we add comments we should remember to include all this information.

src/MgvGovernable.sol
adhusson: I would say `maxRecursionDepth`

adhusson:
Great PR :) I would like to fully review&merge this only once the diff has become "smaller" (ie after we merged @espendk's changes to execution flow) , is this OK?

src/MgvGovernable.sol
lnist: Alright, I'll make the change after you do the thorough review.

lnist:
@adhusson : Definitely fine - hopefully the other is is in soon.

espendk:
Umm, if we need the other change in first, either we have to wait for @jkrivine to have time to review it (including getting acquainted with the new ticks code) or decide it's sufficient that the three of us have approved it 🙂

src/MgvGovernable.sol
lnist: The change has been made.

src/MgvOfferTaking.sol
```diff
@@ -191,13 +191,20 @@ abstract contract MgvOfferTaking is MgvHasOffers {
   /* `internalMarketOrder` works recursively. Going downward, each successive offer is executed until the market order stops (due to: volume exhausted, bad price, or empty book). Then the [reentrancy lock is lifted](#internalMarketOrder/liftReentrancy). Going upward, each offer's `maker` contract is called again with its remaining gas and given the chance to update its offers on the book. */
   function internalMarketOrder(Pair storage pair, MultiOrder memory mor, MgvLib.SingleOrder memory sor) internal {
     unchecked {
+      if (mor.fillVolume == 0 || sor.offerId == 0 || !sor.offer.tick().better(mor.maxTick)) {
+        return endInternalMarketOrder(pair, mor, sor);
```
adhusson: Why moving to an additional function? It uses additional gas & reduces the total available stack space a little bit (edit: I understood, see other comment)

test/core/TheClog.t.sol
```diff
@@ -78,6 +78,22 @@ contract TooDeepRecursionClogTest is MangroveTest, IMaker {
     takeSome(minVolume);
     takeSome(minVolume + 1);
   }
+
+  function testFail_take_more_than_first_fails_for_deep_stack() public {
```
adhusson: `testFail` will be [soon](https://github.com/foundry-rs/foundry/pull/4969) deprecated/discouraged by foundry so I think we could move this test and the one below to a "normal" test?

src/MgvOfferTaking.sol
adhusson: @lnist line numbers have gone stale!(?) I could use a refresh 😅

src/MgvOfferTaking.sol
```diff
@@ -191,13 +197,23 @@ abstract contract MgvOfferTaking is MgvHasOffers {
...
       sor.offerDetail = pair.offerData[sor.offerId].detail;
 
+      mor.gasreqLeft += sor.offerDetail.gasreq();
+
+      if (gasleft() < mor.gasreqLeft + sor.local.offer_gasbase()) {
+        return endInternalMarketOrder(pair, mor, sor);
+      }
+
+      mor.maxRecursionDepth--;
+
```
adhusson: Ok I understand why the function `endInternalMarketOrder` was created!

Since moving to ticks, we have 72 bits available in `offer`. I think moving `gasreq` from `offerDetail` to `offer` make sense. It would reduce gas use in some situations, and mainly here it may remove one call to `endInternalMarketOrder` because we could just add `|| gasleft() < ...` to the big `if` at the beginning of `internalMarketOrder`. With the proviso that updating `mor.gasreqLeft` early is not a problem.

By removing one call to it we would then be able to inline it.

test/core/TakerOperations.t.sol
```diff
@@ -756,13 +758,13 @@ contract TakerOperationsTest is MangroveTest {
   }
 
   function test_unsafe_gas_left_fails_order() public {
-    mgv.setGasbase($(base), $(quote), 1);
     quote.approve($(mgv), 1 ether);
     uint ofr = mkr.newOfferByVolume(1 ether, 1 ether, 120_000, 0);
+    mgv.setGasbase($(base), $(quote), 1);
     Tick offerTick = pair.offers(ofr).tick();
 
     vm.expectRevert("mgv/notEnoughGasForMakerTrade");
```
adhusson, line 774: How can that revert still happen? I think I'm missing something

adhusson (changes requested):
Extremely nice commit-by-commit decomposition :)

I _think_ this could use some additional tests re: checking what Mangrove does when the max stack is reached, when a market order ends partially completed, etc.

edit: also we should keep in mind after merging that we still need to account for the impact on frontend / upfront sticker price for the gas

test/core/TakerOperations.t.sol
lnist: Because I set gasbase to 1 which breaks the accounting by not setting aside enough at the beginning of the market order.

src/MgvOfferTaking.sol
lnist: Are you OK with adding this as a task to our general PR, so that this PR can be completed?

src/MgvOfferTaking.sol
lnist: I'll resolve this comment then.

src/MgvOfferTaking.sol
lnist: Updated in the following:

You're on the right track. The reservation in 156 is needed because 260 is an underapproximation. It goes like this when taking two offers:
1. Line 260 reached first time, we have at least `offer_gasbase+gasreq1-gasused1` gas left for rounding off the market order due to what is set aside in 156 and the accounting in internalMarketOrder. It is larger due to line 211 also conditioning on additional offer_gasbase.
2. Line 260 reached the second time we have gasreqLeft be `offer_gasbase +gasreq1-gasused1+gasreq2-gasused2`, however, note that we will not execute offer2 unless gasleft() can also cover an additional offer_gasbase when reaching line 211. 

So we will never start executing an offer unless we also have `offer_gasbase` left on top of the `gasreq` . This should always be sufficient to complete a market order of 1 offer. For many offers this code relies on offer_gasbase being able to cover for the overhead of unwrapping the stack and doing the necessary updates.
Alternatively we should add some constant overhead for each offer on line 260 or let the reservation on line 156 be another value which we know is big enough to cover the unwrap of stack.

Note: gasreq for one offer is the worst case for that offer, so with a set of offers each recursion will set aside too much in case worst case is not hit, so on average a bit extra is set aside for each offer.

test/core/TakerOperations.t.sol
adhusson: ahh of course :)

test/core/TheClog.t.sol
lnist: done

src/MgvGovernable.sol
```diff
@@ -17,6 +17,8 @@ contract MgvGovernable is MgvRoot {
       /* Set initial gasprice and gasmax. */
       setGasprice(_gasprice);
       setGasmax(gasmax);
+      /* Without optimizer enabled it fails above 79. Set default a bit lower to be safe. */
+      setMaxRecursionDepth(75);
```
espendk: If it's safe w/o optimizer at 79, doesn't that mean that the optimizer would increase the max recursion depth? If so, wouldn't it be safe to set it to 79 as the default?

test/core/gas/OfferPosthookFailGasDelta.t.sol
```diff
@@ -0,0 +1,146 @@
...
+        innerRevert([bytes32("mgv/takerTransferFail"), "", ""]);
+      }
+    }
+  }
+}
+
+/// This test is used by OfferGasBaseBaseTest to measure the gas cost of unrolling the stack during market orders which is dominated by posthook.
+/// Posthook part is most expensive when 1) makerExecute fails (due to penalty accounting) and 2) postHook fails (due to event emitted)
+/// So that is the scenario tested here. However, to measure during unrolling we use the `beforePosthook` hook which is only
+/// invoked for successful offers, so we wrap 19 failing offers in 2 successful offers, which effective measures gas for 19 failures and 1 success.
```
espendk, line 69: nifty!

test/lib/MangroveTest.sol
```diff
@@ -156,7 +156,8 @@ contract MangroveTest is Test2, HasMgvEvents {
     quote: TokenOptions({name: "Quote Token", symbol: "$(B)", decimals: 18}),
     defaultFee: 0,
     gasprice: 40,
-    gasbase: 50_000,
+    //FIXME: measure gasbase and consider using different measurement method (see GasBaseTest.t.sol)
+    gasbase: 150_000,
```
espendk: use different measurement method compared to what?

src/MgvOfferTaking.sol
adhusson: ok! (#481)

src/MgvGovernable.sol
lnist: With the assumption that I haven't overlooked some edge case that is more expensive than the testet, then yes, I would expect so. And we can decrease it runtime if needed.
What do you think? Both you and @adhusson : Do you think it should be this hardcoded value or a deployment parameter?

test/lib/MangroveTest.sol
lnist: We currently use transfers inside ActivateSemibook - IMHO this is wrong, we should use the estimate given by GasBaseTest.
(Even gasBaseTest is not the worst case, that would be with a failing posthook, but since failing posthooks does not happen every single time it may be good).

src/MgvGovernable.sol
espendk: It may change over time with new EIPs, right? That could be a reason for having it a config value.
But that's probably rare, so not sure that consideration should weigh much in the decision.

My intuition is that a deployment param would be good. This will make it easy to vary depending on chain, but w/o the overhead of having it as a rarely (I expect) changed config parameter.

test/lib/MangroveTest.sol
espendk: Ok. Did you run the GasBaseTest and set the parameter based on that?

lnist: Yes, but unoptimized

espendk: Ok, makes sense.

So the FIXME is for having this discussion, right? Or is there more to it?

Maybe it could be changed into a comment describing how to check/update `gasbase` for `MangroveTest`?

src/MgvGovernable.sol
lnist: Constant vs config vs deployment also applies to the 7000 constant for recursion overhead

src/MgvOfferTaking.sol
```diff
@@ -154,8 +172,8 @@ abstract contract MgvOfferTaking is MgvHasOffers {
       OfferList storage offerList;
       (sor.global, sor.local, offerList) = _config(olKey);
       mor.maxRecursionDepth = sor.global.maxRecursionDepth();
-      /* We set aside gas for ending a market order. */
-      mor.gasreqLeft = sor.local.offer_gasbase();
+      /* We have an upper limit om total gasreq for failing offers to avoid failing offers delivering nothing and exhausting gaslimit for transaction. */
```
```suggestion
      /* We have an upper limit on total gasreq for failing offers to avoid failing offers delivering nothing and exhausting gaslimit for the transaction. */
```

src/MgvOfferTaking.sol
```diff
@@ -202,20 +220,15 @@ abstract contract MgvOfferTaking is MgvHasOffers {
   {
     unchecked {
       if (
-        mor.fillVolume == 0 || sor.offerId == 0 || mor.maxRecursionDepth == 0 || sor.offer.logPrice() > mor.maxLogPrice
+        mor.fillVolume == 0 || sor.offer.logPrice() > mor.maxLogPrice || sor.offerId == 0 || mor.maxRecursionDepth == 0
+          || mor.gasreqForFailingOffers > mor.maxGasreqForFailingOffers
       ) {
         return endInternalMarketOrder(offerList, mor, sor);
```
espendk: This can be inlined now, right?

src/MgvOfferTaking.sol
```diff
@@ -202,20 +220,15 @@ abstract contract MgvOfferTaking is MgvHasOffers {
   {
     unchecked {
       if (
-        mor.fillVolume == 0 || sor.offerId == 0 || mor.maxRecursionDepth == 0 || sor.offer.logPrice() > mor.maxLogPrice
+        mor.fillVolume == 0 || sor.offer.logPrice() > mor.maxLogPrice || sor.offerId == 0 || mor.maxRecursionDepth == 0
+          || mor.gasreqForFailingOffers > mor.maxGasreqForFailingOffers
       ) {
```
espendk: `MgvReader` should be updated to reflect this.

preprocessing/structs.ts
```diff
@@ -260,6 +260,8 @@ library OfferDetailUnpackedExtra {
       { name: "dead", bits: 1, type: "bool" },
       /* * `maxRecursionDepth` is the maximum number of times a market order can recursively execute offers. This is a protection against stack overflows. */
       { name: "maxRecursionDepth", bits: 8, type: "uint" },
+      /* * `maxGasreqForFailingOffers` is the maximum gasreq failing offers can consume in total. This is used in a protection against failing offers consuming gaslimit for transaction. Setting it high would make it possible for successive failing offers to consume gaslimit, setting it too low will make a non-healthy book not execute too many offers. Gasmax and maxRecursionDepth bit sizes constrain this  */
```
```suggestion
      /* * `maxGasreqForFailingOffers` is the maximum gasreq failing offers can consume in total. This is used in a protection against failing offers consuming gaslimit for transaction. Setting it too high would make it possible for successive failing offers to consume gaslimit, setting it too low will make a non-healthy book not execute enough offers. `gasmax` and `maxRecursionDepth` bit sizes constrain this.  */
```

src/MgvOfferTaking.sol
lnist: Yes - I think that should be a separate PR

src/MgvOfferTaking.sol
lnist: How much of it? "MAxGasreqForFailingOffers" for instance does not make sense since MgvReader assumes success, right?

espendk: Indeed it assumes success. 

The MgvReader implementation closely mirrors the MgvOfferTaking implementation for ease of validation and maintenance; This PR changes the structure of the code and I think MgvReader should be updated to reflect that change.

test/lib/MangroveTest.sol
lnist: I have updated it with a newly measured value. Note that this is for an entire market order of 1 offer to an empty book. Previously we used only transfers as baseline which would give a value of 128176.

src/MgvOfferTaking.sol
lnist: Resolving obsolete discussion.

src/MgvGovernable.sol
lnist: Resolving obsolete discussion.

src/MgvOfferTaking.sol
lnist: Done

script/periphery/CopyOpenSemibooks.s.sol
```diff
@@ -4,6 +4,7 @@ pragma solidity ^0.8.13;
 import "mgv_src/periphery/MgvReader.sol";
 import {Deployer} from "mgv_script/lib/Deployer.sol";
 import {MgvStructs, OLKey} from "mgv_src/MgvLib.sol";
+import "mgv_lib/Debug.sol";
```
espendk, line 7: Should this be deleted?

```suggestion
```

script/periphery/UpdateMarket.s.sol
```diff
@@ -4,6 +4,7 @@ pragma solidity ^0.8.13;
 import "mgv_src/periphery/MgvReader.sol";
 import {IERC20} from "mgv_src/IERC20.sol";
 import {Deployer} from "mgv_script/lib/Deployer.sol";
+import "mgv_lib/Debug.sol";
```
espendk, line 7: Delete?

```suggestion
```

src/MgvOfferTakingWithPermit.sol
```diff
@@ -5,7 +5,9 @@ import {HasMgvEvents, Tick, LogPriceLib, OLKey} from "./MgvLib.sol";
 
 import {MgvOfferTaking} from "./MgvOfferTaking.sol";
 import {TickLib} from "./../lib/TickLib.sol";
+import "mgv_lib/LogPriceConversionLib.sol";
 import {MgvStructs} from "./MgvLib.sol";
+import "mgv_lib/Debug.sol";
```
espendk, line 10: Delete?

```suggestion
```

espendk (commented):
I thought there would be changes to `MgvReader`? Cf this comment: https://github.com/mangrovedao/mangrove-core/pull/461#discussion_r1315884201

espendk:
> I thought there would be changes to `MgvReader`? Cf this comment: [#461 (comment)](https://github.com/mangrovedao/mangrove-core/pull/461#discussion_r1315884201)

Doh, there is! Somehow GitHub didn't list them in the changes since my last review...
Sorry!

script/periphery/CopyOpenSemibooks.s.sol
lnist: No - they implicitly got console from reader, and I removed the import in reader, making these scripts fail. We could have imported only console, but the debug toString stuff is nice to have.

src/AbstractMangrove.sol
```diff
@@ -27,6 +27,12 @@ abstract contract AbstractMangrove is MgvOfferTakingWithPermit, MgvOfferMaking {
       require(success, "mgv/ctor/gasprice");
       (success,) = APPENDIX.delegatecall(abi.encodeCall(MgvGovernable.setGasmax, (gasmax)));
       require(success, "mgv/ctor/gasmax");
+      /* Without optimizer enabled it fails above 79. Set default a bit lower to be safe. */
+      (success,) = APPENDIX.delegatecall(abi.encodeCall(MgvGovernable.setMaxRecursionDepth, (75)));
```
adhusson: I'm wondering if it shouldn't be an immutable. 
- I don't think different chains can yield different max recursion depths.
- Nor can an evolution of the gas pricing yield a different max recursion depths.
- So if the value is correct at deployment, it's correct forever (?)

Also, as long as there's a test that checks the max depth can be reached, having a wrong value there is wasteful at worst. And if we have one more test that checks (max depth + 1) cannot be reached, we canot make a wrong deployment :)

src/AbstractMangrove.sol
```diff
@@ -27,6 +27,12 @@ abstract contract AbstractMangrove is MgvOfferTakingWithPermit, MgvOfferMaking {
...
+      require(success, "mgv/ctor/maxRecursionDepth");
+      /* Arbitrary initial limit for number of failing offers - let on average 10 failing offers consume up to gasmax*/
+      (success,) = APPENDIX.delegatecall(abi.encodeCall(MgvGovernable.setMaxGasreqForFailingOffers, (gasmax * 10)));
```
adhusson: Same here I would have the `10` in Constants.sol

src/IMangrove.sol
```diff
@@ -79,6 +79,15 @@ interface IMangrove is HasMgvEvents {
     external
     returns (uint takerGot, uint takerGave, uint bounty, uint feePaid);
 
+  function marketOrderByLogPrice(
+    OLKey memory olKey,
+    int maxLogPrice,
+    uint fillVolume,
+    bool fillWants,
+    uint maxGasreqForFailingOffers,
+    uint maxRecursionDepth
```
adhusson: Following the above, I don't think `maxRecursionDepth` should be a param -- and even in a world where it's still alterable by governance, I don't see any good reason for callers to customize the value?

src/MgvOfferTaking.sol
```diff
@@ -89,6 +112,11 @@ abstract contract MgvOfferTaking is MgvHasOffers {
       sor.olKey = olKey;
       OfferList storage offerList;
       (sor.global, sor.local, offerList) = _config(olKey);
+      mor.maxRecursionDepth = maxRecursionDepth > 0 ? maxRecursionDepth : sor.global.maxRecursionDepth();
+      /* We have an upper limit on total gasreq for failing offers to avoid failing offers delivering nothing and exhausting gaslimit for the transaction. */
+      mor.maxGasreqForFailingOffers =
```
adhusson, line 160: I'm missing why we have to copy `maxGasreqForFailingOffers` from `sor` to `mor`

adhusson (changes requested):
I wouldn't want to be a plumber, this is hard :D 

On top of my comments, I think we should do the "move `gasreq` to `offer` so that the code for `internalMarketOrder` doesn't have to be restructured so much with an additional `endInternalMarketOrder` function." -- I think we would want to do it anyway but doing it now prevents another PR cycle where we swap the `internalMarketOrder` structure back to where it was. (I can add it as a commit if you want)

test/lib/MangroveTest.sol
lnist: Added FIXME in ActivateSemibook

src/AbstractMangrove.sol
lnist: As long as we hit the very worst code paths wrt stack depth, then yes it is a constant (modulo optimization). For the current tests, it fails above depth 80 with optimization and 200 runs. So setting to a constant of 75. This should leave enough room for some weird extra costly path. And in that case the path will most likely not be hit often, but if it is, we can pass in an overridden value via external calls.

src/IMangrove.sol
lnist: It depends on how cautious  we are. If we are confident we cannot hit the 75 limit, then we can remove the override. Remember the limit only really affects successful offers since failing offers can also be controlled by maxGasreqForFailingOffers.

src/MgvOfferTaking.sol
lnist: I don't understand? Maybe the code changed?

src/MgvGovernable.sol
lnist: It should only change with optimizations, so we decided to set it at an expected safe 75 constant value.

src/MgvOfferTaking.sol
lnist: ah, it is because I allow it to be overridden by parameter to function.

src/AbstractMangrove.sol
lnist: I have made it a deployment parameter (last commit - can revert if you disagree). Why? Because 10 was quite arbitrary. We could consider:
1) Let it be hardcoded, but with 10 in constants like you suggest, since it can be set via governance
2) Let it be a multiplier of gasmax and thus save a lot of bits in global struct.
wdyt?

adhusson: Hardcoded multiplier of gasmax seems like the easiest thing!

src/IMangrove.sol
adhusson: I think that even if the 75 limit is wrong, it's fine if governance changes it, right? (assuming we don't make it an immutable)

I think that `maxRecursionDepth` works differently from `maxGasreqForFailingOffers`, which has to be adjusted depending on type of clog. Whereas `maxRecursionDepth` ideally should just be set to the perfect value and never changed?

src/MgvOfferTaking.sol
adhusson: Ah yes makes sense :)

PR #469, opened 2023-08-28
Feat: add mgv config to handle addresses dependancies

peterMangrove:
This PR update how we configure and read addresses in mangrove-core.
It adds the file `mgvConfig.json` which holds the paths to all the dependancy addresses mangrove-core has. Mangrove-core is only dependant on its own addresses folder, it therefore only points to that.
In mangrove-strats we dependant on both its own addresses folder and mangrove-core's addresses folder, therefore the `mgvConfig.json` points to both in mangrove-strats. The update to mangrove-strats can be found here: https://github.com/mangrovedao/mangrove-strats/pull/28

The reason for this change, was that in the mangrove-strats, we depend on the addresses from mangrove-core. But in order to read them, i would (with the current solution), have to add the path to mangrove-core in the .env file. Meaning mangrove-strats would not work out of the box. With this new solution, it works out of the box and clearly shows that there is a dependancy to mangrove-core.


adhusson (dismissed):
looks good!

lnist (approved):

PR #483, opened 2023-08-30
tick math

adhusson:
Inspired by uniswap's approach, using price instead of sqrtPrice, and floating point instead of fixed point.

Precise conversion functions:
- `logPriceFromPrice: mantissa,exp -> logPrice`
- `priceFromLogPrice: logPrice -> mantissa,exp`

Also:
- `logPriceFromVolumes: inbound,outbound -> logPrice`
- `priceFromVolumes: inbound,outbound -> mantissa,exp`

Some properties:
* `logPrice` must be in the range`+/- 1_048_575` to be valid (that's `+/- 2^20-1`). With a 1bp tick spacing, that means prices between `2.9e-46` and `3.4e45` (more than enough).
* `logPriceFromPrice(price)` should return `max { logPrice | priceFromLogPrice(logPrice) <= price }`
* For a valid `logPrice`, `logPriceFromPrice(priceFromLogPrice(logPrice)) == logPrice`.
* `priceFromLogPrice(logPrice)` is sometimes a little below, sometimes a little above `1.0001^logPrice`. Precision should be within 0.0001%. There should be some off-by-1 results, that is, if `1.0001^logPrice < price < priceFromLogPrice(logPrice)` then  `logPriceFromPrice(price) = ⌊log_1.0001(price)⌋ - 1`, and if `priceFromLogPrice(logPrice) < price < 1.0001^logPrice` then `logPriceFromPrice(price) = ⌊log_1.0001(price)⌋ + 1`.
* Currently the tree data structure does not accomodate this range with a 1bp tick spacing (current bounds are `+/- 524287`).

### DIFF
--- a/lib/Constants.sol
+++ b/lib/Constants.sol
@@ -1,0 +1,50 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.17;
+
+uint constant ONES = type(uint).max;
+uint constant TOPBIT = 1 << 255;
+
+// MIN_TICK and MAX_TICK should be inside the addressable range defined by the sizes of LEAF, LEVEL0, LEVEL1, LEVEL2
+int constant MIN_TICK = -524287;
+int constant MAX_TICK = -MIN_TICK;
+
+// sizes must match field sizes in structs.ts where relevant
+uint constant TICK_BITS = 24;
+uint constant OFFER_BITS = 32;
+
+// only power-of-two sizes are supported for LEAF_SIZE and LEVEL*_SIZE
+uint constant LEAF_SIZE_BITS = 2; 
+uint constant LEVEL0_SIZE_BITS = 6;
+uint constant LEVEL1_SIZE_BITS = 6;
+uint constant LEVEL2_SIZE_BITS = 6;
+
+int constant LEAF_SIZE = int(2 ** (LEAF_SIZE_BITS));
+int constant LEVEL0_SIZE = int(2 ** (LEVEL0_SIZE_BITS));
+int constant LEVEL1_SIZE = int(2 ** (LEVEL1_SIZE_BITS));
+int constant LEVEL2_SIZE = int(2 ** (LEVEL2_SIZE_BITS));
+
+uint constant LEAF_SIZE_MASK = ~(ONES << LEAF_SIZE_BITS);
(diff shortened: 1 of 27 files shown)


lib/Constants.sol
```diff
@@ -0,0 +1,50 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.17;
+
+uint constant ONES = type(uint).max;
+uint constant TOPBIT = 1 << 255;
+
+// MIN_TICK and MAX_TICK should be inside the addressable range defined by the sizes of LEAF, LEVEL0, LEVEL1, LEVEL2
+int constant MIN_TICK = -524287;
+int constant MAX_TICK = -MIN_TICK;
```
espendk, line 9: `MIN_TICK` can be one smaller in 20 bits:
```suggestion
int constant MIN_TICK = -524288;
int constant MAX_TICK = -MIN_TICK - 1;
```

lib/Constants.sol
```diff
@@ -0,0 +1,50 @@
...
+int constant NUM_LEAFS = NUM_LEVEL0 * LEVEL0_SIZE;
+int constant NUM_TICKS = NUM_LEAFS * LEAF_SIZE;
+
+uint constant OFFER_MASK = ONES >> (256 - OFFER_BITS);
+
+
+
+// +/- 2**20-1 because only 20 bits are examined by the logPrice->price function
+int constant MIN_LOG_PRICE = -((1 << 20)-1);
+int constant MAX_LOG_PRICE = -MIN_LOG_PRICE;
```
espendk, line 42: If understand things correctly, the limitation here is not only the 20 bits, but also that we assume we can represent `abs(logPrice)` in 20 bits, correct? Otherwise, `MIN_LOG_PRICE` could be `-(1 << 20)`.

espendk (commented):
Added a few smaller comments.

I did not review the actual tick/logPrice/price math in any detail. Let's discuss if that's needed.

lib/Constants.sol
adhusson: yes, the function never inspects logPrice if it's negative.

espendk: Ok. I think it would be good to update the comment with this info

PR #484, opened 2023-08-31
fix(test): Set meaningful density

lnist:


test/lib/MangroveTest.sol
```diff
@@ -79,7 +79,7 @@ contract MangroveTest is Test2, HasMgvEvents {
     defaultTickscale: 1,
     gasprice: 40,
     gasbase: 50_000,
-    density: 10,
+    density: 2 ** 32,
```
espendk, line 82: I'm surprised this doesn't affect more tests!

espendk: Why/how was this density chosen?

espendk (commented):
The implications of the density change in `MangroveTest` is unclear to me. Can you say a little about how the density was chosen and how this will impact tests going forward?

test/lib/MangroveTest.sol
lnist: At random, bigger than what it was :)

lnist: The tokens have fictional values, so the importance is that asserts can be made regarding density, not the exact value. Previously it was effectively 0.

espendk: Ok :-) We can adjust later if needed

espendk (approved):
lgtm

PR #485, opened 2023-08-31
feat: emit offerSucces and offerFail during posthook end

peterMangrove:
This PR moves the events OfferSucces and OfferFail, so that they are not longer called at offer execution but at posthook end. The reason for moving the events, is to be able to add the exact penalty amount.

By moving the OfferFail event, it enables us to add the penalty amount to the event, without having to add another event. 
The taker has also been removed from the OfferSuccess and OfferFail events, as that information can be gathered ny the OrderComplete event. ( Later, in another PR, that will be moved to OrderStart)
This now solves the issue that we were not able deduce what offer triggered what penalty. This was becayse the OfferFail event was emitted at offer execution and the corresponding credit event (handling the penalty) was done in the posthook. We can now easily tell what offer triggered what penalty. And without adding any new events.

The consequences of this is:
1. OfferSucces and OfferFail are now emitted in reverse, compared to before. This is because they are emitted in the posthook.
2. When Cleaning offers, then they are not emitted in reverse, as each offer is taken one by one and not by a market order.

### DIFF
--- a/src/IMangrove.sol
+++ b/src/IMangrove.sol
@@ -24,9 +24,9 @@ interface IMangrove {
   event Debit(address indexed maker, uint amount);
   event Kill();
   event NewMgv();
-  event OfferFail(bytes32 indexed olKeyHash, uint id, address taker, uint takerWants, uint takerGives, bytes32 mgvData);
+  event OfferFail(bytes32 indexed olKeyHash, uint id, uint takerWants, uint takerGives, uint penalty, bytes32 mgvData);
   event OfferRetract(bytes32 indexed olKeyHash, uint id, bool deprovision);
-  event OfferSuccess(bytes32 indexed olKeyHash, uint id, address taker, uint takerWants, uint takerGives);
+  event OfferSuccess(bytes32 indexed olKeyHash, uint id, uint takerWants, uint takerGives);
   event OfferWrite(
     bytes32 indexed olKeyHash, address maker, int logPrice, uint gives, uint gasprice, uint gasreq, uint id
   );
--- a/src/MgvLib.sol
+++ b/src/MgvLib.sol
@@ -104,7 +104,6 @@ contract HasMgvEvents {
     bytes32 indexed olKeyHash,
     uint id,
     // `maker` is not logged because it can be retrieved from the state using `(outbound_tkn,inbound_tkn,id)`.
-    address indexed taker,
     uint takerWants,
     uint takerGives
   );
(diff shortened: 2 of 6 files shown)


adhusson:
Espen and I thought being out of order with the posthook events was a bridge too far in [this discussion](https://github.com/mangrovedao/mangrove-core/pull/459#discussion_r1299399058). However the only benefit we had in mind was that the penalty value could be included.

espendk (approved):
It's weird that `Offer{Fail,Success}` will come in the opposite order of how they happened during execution. But I like that `penalty` is tied to `OfferFail`. And the alternative of having a separate `Penalty` event (my favorite so far) would cost more gas, so I guess we'll have to accept the compromise.

For the record, I'm noting that for `clean`, the `Offer{Fail,Success}` events will come in the order they are executed, adding to the weirdness of this solution. And it'll probably result in extra complexity when indexing, since the event orders will differ:

```
marketOrder events:
  OrderStart
  Offer{Fail,Success}  for LAST executed offer
  …
  Offer{Fail,Success}  for FIRST executed offer
  OrderComplete

clean events:
  CleanStart
  Offer{Fail,Success}  for FIRST executed offer
  …
  Offer{Fail,Success}  for LAST executed offer
  CleanEnd
```

This is assuming that we wrap them in different events (which I think we have to, since clean doesn’t have the needed data for OrderStart).

peterMangrove:
> It's weird that `Offer{Fail,Success}` will come in the opposite order of how they happened during execution. But I like that `penalty` is tied to `OfferFail`. And the alternative of having a separate `Penalty` event (my favorite so far) would cost more gas, so I guess we'll have to accept the compromise.
> 
> For the record, I'm noting that for `clean`, the `Offer{Fail,Success}` events will come in the order they are executed, adding to the weirdness of this solution. And it'll probably result in extra complexity when indexing, since the event orders will differ:
> 
> ```
> marketOrder events:
>   OrderStart
>   Offer{Fail,Success}  for LAST executed offer
>   …
>   Offer{Fail,Success}  for FIRST executed offer
>   OrderComplete
> 
> clean events:
>   CleanStart
>   Offer{Fail,Success}  for FIRST executed offer
>   …
>   Offer{Fail,Success}  for LAST executed offer
>   CleanEnd
> ```
> 
> This is assuming that we wrap them in different events (which I think we have to, since clean doesn’t have the needed data for OrderStart).

Agreed that we need new events for Clean. This will be implemented in another PR.

peterMangrove:
A brief summary of the discussion on slack:

1. A similar approach was attempted earlier, see [here](https://github.com/mangrovedao/mangrove-core/pull/485#issuecomment-1701095069)
2. There was a lot of discussion on whether the reverse order of events when doing a MarketOrder would add to much complexity. Especially because it would not be reverse in the Clean scenario.
3. By moving the events to posthook, we are not only able to add the penalty, but we would also save gas, both with respect to less events but also e.g. Clean that would emit less events, in the SfferSuccess scenario.
4. From the indexer perspective, it has no real difference in what order the events are emitted, as they always come in a context of an offerlist and offer id.
5. We looked into having consistency between Clean and Market order, w.r.t. the sequence of events. But that is not doable with this approach. Instead a future PR will add new event for the Clean sequence, in order to better tell MarketOrder and Clean apart.
6. We ultimately agreed to this PRs approach, knowing that future interrations might be needed.

PR #486, opened 2023-08-31
Add more tests of the tick tree

espendk:
This PR implements tests of Mangrove's tick tree to ensure that the tree is always in a consistent state.

# TODO
- [x] Once tick math supports `MIN_TICK` the tests that use it should be enabled
- [x] Once the bug described in `TickTreeUpdateOffer` is fixed, the tests that trigger it should be enabled
- [x] Consider testing what the tick tree looks like when inside `makerExecute` and `makerPosthook`.
   - `makerExecute`: We'll add a read-lock so this isn't relevant: https://github.com/mangrovedao/mangrove-core/issues/503
   - `makerPosthook`: There's a separate issue on that: https://github.com/mangrovedao/mangrove-core/issues/506
- [x] Review the `FIXMEs` added in this PR; Many of them are comments and questions that should be discussed and then the FIXME should be removed.
   - Added questions on the `FIXME`s in this PR for @adhusson  :-)
- [x] Review whether all relevant tick scenarios are included for all test contracts

# High-level test approach
For optimization reasons, the tick tree data structure is not encapsulated but read and manipulated in a low-level way in all the operations that use it (and this is interleaved with other non-tree logic). This means that we cannot unit test the data structure, but instead have to make integration tests of all the functions that manipulate the tree.

The approach taken in these integration tests is to simulate what Mangrove should be doing to its tick tree in a parallel tick tree implementation which is simpler and easier to reason about. By comparing the resulting tick trees (MGV and the test implementation) we can check that they agree on the end state; If not, we can report exactly where the discrepancy is.

In overview, the tests are structured as follows:

1. Set up Mangrove's initial state, i.e, post offers at relevant ticks (and possibly retract some of them).
2. Take a snapshot of Mangrove's tick tree, i.e, a copy in the form of a test `TickTree`
3. Perform some operation on Mangrove (e.g, add or remove an offer)
4. Perform equivalent operation(s) on the test `TickTree`
5. Compare Mangrove's tick tree to the test `TickTree`.

# Structure of the test contracts
There are multiple test contracts named `TickTree*` all located in `test/core/ticktree`.

The base contract is `TickTreeTest` and implements the test `TickTree` struct and operations on it as well as shared functions for the test contracts.

For each Mangrove operation that modifies the tick tree, there is a separate contract which tests that operation on the tick tree. For example, `newOffer*` is tested in `TickTreeNewOfferTest`.

# Test scenarios
Mangrove's tick tree is quite complex due to the different levels each being stored separately as well as some things being stored in `local`. There are there many scenarios to test, e.g, updating an offer to another tick within the same `Leaf` is different than updating it to a tick below a different level1 position.

### Tick of interest, higher tick, and lower tick
For all the tests, there's a notion of "tick of interest", a higher tick, and a lower tick. This allows us to capture the different scenarios where ticks are in the same/different leaf, level0, level1, and level2. `TickTreeTest` contains functions for generating all relevant higher and lower ticks for a given tick of interest. Combinations of three such ticks (including the absence of higher/lower) is captured by the `TickScenario` struct.

For each test contract, these we define a `*Scenario` struct (eg `UpdateOfferScenario`) that defines what a scenario looks like and we describe how the `TickScenario` should be interpreted in these scenarios.

## Structure of test contracts
The tests contracts are therefore generally structured as follows:

1. An execute scenario function (e.g, `run_update_offer_scenario`)
   - This executes a single scenario
   - As this function may be called multiple times within a single `test_*` function, it does a `vm.snapshot()` initially and a `vm.revertTo(vmSnapshotId)` at the end. This is super fragile and if a scenario fails, you may see other, totally unrelated tests fail...
2. An execute scenarios for a specific tick (e.g, `run_update_offer_scenarios_for_tick`)
   - This runs all scenarios for a particular tick of interest
   - The reason for running multiple scenarios in one function is to avoid having to manually write many `test_*` functions that enumerate all the scenarios.
3. One or more `test_*` functions (e.g, `test_clean_offer_for_tick_0`)
   - This is the actual test function that `forge` will run.
   - The reason for having multiple `test*_` functions instead of running all scenarios in one test is that we run into limitations of `Foundry` when we do this.
   - In other words, there's a limit to how many scenarios we can run in one `test_*` function which depends on the memory used by the `run_*_scenario` function. 
   - We've tried to minimize the number of `test_*` functions, but for `updateOffer` we've had to make 4 `test_*` functions per tick we want to test :-/


adhusson:
Bug fixed in #494

test/core/ticktree/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+There are multiple test contracts named `TickTree*` all located in `test/core/ticktree`.
+
+The base contract is `TickTreeTest` and implements the test `TickTree` struct and operations on it as well as shared functions for the test contracts.
+
+For each Mangrove operation that modifies the tick tree, there is a separate contract which tests that operation on the tick tree. For example, `newOffer*` is tested in `TickTreeNewOfferTest`.
+
+# Test scenarios
+
+Mangrove's tick tree is quite complex due to the different levels each being stored separately as well as some things being stored in `local`. There are there many scenarios to test, e.g, updating an offer to another tick within the same `Leaf` is different than updating it to a tick below a different level1 position.
```
lnist:
```suggestion
Mangrove's tick tree is quite complex due to the different levels each being stored separately as well as some things being stored in `local`. There are there many scenarios to test, e.g, updating an offer to another tick within the same `Leaf` is different than updating it to a tick below a different `level1` position.
```

test/core/ticktree/TickTreeTestAssumptions.t.sol
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier:	AGPL-3.0
+
+pragma solidity ^0.8.18;
+
+import {TickTreeTest} from "./TickTreeTest.t.sol";
+import "mgv_src/MgvLib.sol";
+import "mgv_lib/Debug.sol";
+
+// Tests of the assumptions made in the TickTreeTests.
+// FIXME: Not sure this is needed anymore?
```
lnist: ?

test/core/ticktree/TickTreeTestAssumptions.t.sol
```diff
@@ -0,0 +1,38 @@
...
+contract TickTreeTestAssumptionsTest is TickTreeTest {
+  // Cheks that the ticks used in these tests have the expected locations at various levels.
```
lnist:
```suggestion
  // Checks that the ticks used in these tests have the expected locations at various levels.
```

test/core/ticktree/TickTreeTest.t.sol
```diff
@@ -0,0 +1,729 @@
...
+import "mgv_src/MgvLib.sol";
+import "mgv_lib/Debug.sol";
+
+// Base class for test of Mangrove's tick tree data structure
+//
+// Provides a simple tick tree data structure and operations on it that can be used to simulate Mangrove's tick tree
+// and then be compared to the actual tick tree.
+//
+// The test tick tree operations uses simpler (and less efficient) code to manipulate the tick tree, which should make
+// it clearer what is going on and easier to convince yourself that the tick tree is manipulated correctly.
```
lnist, line 28: reference readme.md here?

test/core/ticktree/TickTreeTest.t.sol
```diff
@@ -0,0 +1,729 @@
...
+    offerData.gasprice = gasprice;
+  }
+
+  // # Mangrove tick tree asserts
+
+  // Checks that the current Mangrove tick tree in olKey is consistent
+  function assertMgvTickTreeIsConsistent() internal {
+    Field level2 = mgv.level2(olKey);
+    for (uint level2Pos = 0; level2Pos <= MAX_LEVEL2_POSITION; ++level2Pos) {
+      if (!isBitSet(level2, level2Pos)) {
```
lnist: does it also assert that it should be set if it is not?

test/core/ticktree/TickTreeTest.t.sol
```diff
@@ -0,0 +1,729 @@
...
+        isBitSet(level2, level2Pos),
+        isBitSet(tickTree.local.level2(), level2Pos),
+        string.concat("level2 bit mismatch, pos: ", vm.toString(level2Pos))
+      );
+      if (!isBitSet(level2, level2Pos)) {
+        continue;
+      }
+
+      int level1Index = level1IndexFromLevel2Pos(level2Pos);
+      Field level1 = mgv.level1(olKey, level1Index);
```
lnist: If these loops OOM then maybe putting each level in its own function will help it release memory?

test/core/ticktree/TickTreeMarketOrder.t.sol
```diff
@@ -0,0 +1,243 @@
...
+  }
+
+  struct MarketOrderScenario {
+    TickListScenario lowerTick;
+    TickListScenario middleTick; // ignored if lowerTickListScenario.size == 2 (scenario 3. and 4.)
+    TickListScenario higherTick; // ignored if {lower,middle}TickListScenario.size == 2 (scenario 3. and 4.)
+  }
+
+  // (list size, offers to take)
+  uint[2][] tickListScenarios = [[0, 0], [1, 1], [2, 1], [2, 0]];
```
lnist: so we never take 2 offers?

test/core/ticktree/TickTreeMarketOrder.t.sol
```diff
@@ -0,0 +1,243 @@
...
+    (uint[] memory higherOfferIds, uint higherOffersGive) =
+      add_n_offers_to_tick(Tick.unwrap(scenario.higherTick.tick), scenario.higherTick.size);
+    uint fillVolume = lowerOffersGive * scenario.lowerTick.offersToTake
+      + middleOffersGive * scenario.middleTick.offersToTake + higherOffersGive * scenario.higherTick.offersToTake;
+
+    // 3. Snapshot tick tree
+    TickTree storage tickTree = snapshotTickTree();
+
+    // 4. Run the market order
+    mgv.marketOrderByLogPrice(olKey, MAX_LOG_PRICE, fillVolume, true);
```
lnist, line 233: Consider (but maybe separate PR) whether it is worth it to snapshot the resulting ticktree inside posthook of the last taken offer, so that we verify tree is right already when posthooks are executed. Not high priority.

lnist (commented):
Very thorough tests - good decision to make a simple implementation to compare against the other-code-interleaved-version.

test/core/ticktree/TickTreeMarketOrder.t.sol
espendk: Indeed; maybe we should?

Maybe a single test that ensures we can take two offers would be sufficient to test the tick list updates in market orders?

espendk: Tick list scenarios have been changed such that fully taken lists have length 2: 150c33c1bb456ff1f982118d155bdf2c65917f4a

test/core/ticktree/TickTreeTestAssumptions.t.sol
espendk: Addressed in bda293ace9b1eb6402d3aa65425b64f821414d99

(I decided to keep this contract as self-checking documentation of the assumptions)

test/core/ticktree/TickTreeTest.t.sol
espendk: Addressed in 917b25d6dc07e36529cff757f41114e03f2aa207

test/core/ticktree/TickTreeTest.t.sol
espendk: Good point. I added a check that the level below is empty if the bit isn't set: 7e8a3a31181a4d6f5bdbb527feec4d40fbd8ef9b

I also implemented a check that the entire subtree was empty, but that used too much gas, so I compromised on one level below.

test/core/ticktree/TickTreeTest.t.sol
espendk: Worth trying if we run into the issue in the future. For now, the checks succeed so I think we should park it here :-)

test/core/ticktree/TickTreeMarketOrder.t.sol
espendk: Good idea.  I've created an issue on it: https://github.com/mangrovedao/mangrove-core/issues/506

lib/Constants.sol
```diff
@@ -33,6 +33,18 @@ int constant NUM_LEVEL0 = NUM_LEVEL1 * LEVEL1_SIZE;
 int constant NUM_LEAFS = NUM_LEVEL0 * LEVEL0_SIZE;
 int constant NUM_TICKS = NUM_LEAFS * LEAF_SIZE;
 
+// FIXME: Should these be here or placed somewhere else? Besides being useful in tests, they serve as documentation for the datastructure
```
espendk: @adhusson Do you have an opinion on where these constants should be?

lib/TickLib.sol
```diff
@@ -209,7 +209,9 @@ library TickLib {
   // see note posIn*
   // note with int24 tick we only use 64 bits of level2 (64*256*256*4 is 2**24)
   // the goal is that have the bit positions in {} used:
-
+// FIXME: This is a non-standard way of writing bit positions, normally they're numbered right-to-left
+// FIXME: Also, the highest position in a field is 63 and is used for the highest tick in the level.
+//        This feels like it's conflicting with the stmt in the Excalidraw that says "ticks go right (cheap) to left (expensive)"
```
espendk, line 214: Maybe we can leave this comment here for when we update all comments?
@adhusson wdyt?

espendk:
Thanks for the review @lnist ! I addressed your comments/suggestions :-)

lib/Constants.sol
adhusson: If they're not used by production code I don't think they should be there. Documentation can help fill in the gaps.

lib/TickLib.sol
adhusson: Yes the diagram is in reverse, it should go the other way, so we can leave the comment there and reverse it later!

lib/preprocessed/ToString.post.sol
```diff
@@ -56,6 +56,10 @@ function toString(LocalUnpacked memory __unpacked) pure returns (string memory)
   return string.concat("Local{","active: ", vm.toString(__unpacked.active), ", ", "fee: ", vm.toString(__unpacked.fee), ", ", "density: ", toString(__unpacked.density), ", ", "tickPosInLeaf: ", vm.toString(__unpacked.tickPosInLeaf), ", ", "level0: ", toString(__unpacked.level0), ", ", "level1: ", toString(__unpacked.level1), ", ", "level2: ", toString(__unpacked.level2), ", ", "kilo_offer_gasbase: ", vm.toString(__unpacked.kilo_offer_gasbase), ", ", "lock: ", vm.toString(__unpacked.lock), ", ", "last: ", vm.toString(__unpacked.last),"}");
 }
 
+function tickBranchToString(Tick tick) pure returns (string memory) {
+  return string.concat(vm.toString(tick.posInLevel2()), "->", vm.toString(tick.posInLevel1()), "->", vm.toString(tick.posInLevel0()), "->", vm.toString(tick.posInLeaf()));
```
adhusson: I would go even further and give the indices of each level!

test/core/ticktree/TickTreeTest.t.sol
```diff
@@ -0,0 +1,765 @@
...
+      }
+    }
+  }
+
+  // # Test tick tree data structure and operations
+
+  struct Offer {
+    MgvStructs.OfferPacked offer;
+    MgvStructs.OfferDetailPacked detail;
+  }
```
adhusson: Why not reuse MgvRoot.OfferData? (maybe not important)

test/core/ticktree/TickTreeTest.t.sol
```diff
@@ -0,0 +1,765 @@
...
+
+  struct TickTree {
```
adhusson: I think this and everything below might deserve to live in a `test/lib/*.sol` file since it's pretty crucial tooling that is way too buried in a .t.sol. Or at least in a .sol in the `test/core/ticktree` dir.

test/core/ticktree/TickTreeUpdateOffer.t.sol
```diff
@@ -0,0 +1,347 @@
...
+  //         lowerTick: -16384,
+  //         lowerTickListSize: 0
+  //       }),
+  //       offerTickListSize: 1,
+  //       offerPos: 0,
+  //       newTick: -16384
+  //     }),
+  //     true
+  //   );
+  // }
```
adhusson: This can be uncommented now

test/core/ticktree/TickTreeUpdateOffer.t.sol
```diff
@@ -0,0 +1,347 @@
...
+        console.log("  higherTickListSize: %s", vm.toString(scenario.tickScenario.higherTickListSize));
+      }
+      if (scenario.tickScenario.hasLowerTick) {
+        console.log("  lowerTick: %s", toString(Tick.wrap(scenario.tickScenario.lowerTick)));
+        console.log("  lowerTickListSize: %s", vm.toString(scenario.tickScenario.lowerTickListSize));
+      }
+    }
+
+    // 1. Capture VM state before scenario so we can restore it after
+    uint vmSnapshotId = vm.snapshot();
```
adhusson, line 296: This is unreliable, right? Alternatives:
- Actually create a ton of tests (also alleviates the memory leak problems in loops)
- Still loop, but create a new Mangrove every time, so no need to reset a state

foundry.toml
```diff
@@ -12,6 +12,7 @@ fs_permissions = [{ access = "read-write", path = "./addresses/"}, { access = "r
 solc_version="0.8.21"
 ffi=true
 optimizer=false
+memory_limit = 100000000 # 33554432 is default limit
```
adhusson, line 15: Depending on how hard it is, adding manual memory management before merging could be nice. Otherwise this is a ticking time bomb, + it probably slows tests.

adhusson (changes requested):
Super important part of the test infrastructure, already caught a bug, just a few remarks bug apart from that let's go 🚀

test/core/ticktree/TickTreeUpdateOffer.t.sol
espendk: I think it is only unreliable when tests run out of gas/memory. When the tests work, they do so reliably.

I think we should keep it as-is for now and move on. If we need to update/change the tick tree tests at some point, we can look into improving this. Ok?

foundry.toml
espendk: I agree that this is not ideal. But as mentioned above, I think we should wrap up this test suite and park improvements for later, as we don't have much time and much to do.

test/core/ticktree/TickTreeUpdateOffer.t.sol
espendk: That specific scenario is already part of the `tick=0` tests :-) I've kept the commented out single-scenario tests as an easy way to run a single scenario.

But now that I think about it, it would ease maintenance if those tests are commented in. I'll do that!

test/core/ticktree/TickTreeTest.t.sol
espendk: Ok, I'll move it 👍

lib/Constants.sol
espendk: Addressed in 9f6a98e5c4f29483f8edb7959876ee640497e7ab

test/core/ticktree/TickTreeUpdateOffer.t.sol
espendk: Addressed in 794c06c0a341dd0d5757f7989e33eb67187d56bb

test/core/ticktree/TickTreeTest.t.sol
espendk: Good point. Originally the `Offer` struct was different from `MgvRoot.OfferData` but ended up being identical w/o me realizing 😅

Addressed in 5fd47a7ba1d39d980b20ebf27a2d70752dfe895c

lib/preprocessed/ToString.post.sol
espendk: Addressed in 69f8b815afb701420ae5537f9639656fca2a5bf7

test/core/ticktree/TickTreeTest.t.sol
espendk: Addressed in 8bdd91f4a084d87f0110bf15b57966515c88269e

adhusson (approved):
✅✓✔☑

PR #487, opened 2023-09-01
feat(IMangrove): Ensure IMangrove and Mangrove align on signatures

lnist:
See individual commits. The last one may be skipped without much hurt.

### DIFF
--- a/src/IMangrove.sol
+++ b/src/IMangrove.sol
@@ -8,69 +8,50 @@ pragma solidity >=0.7.0 <0.9.0;
 
 pragma experimental ABIEncoderV2;
 
-import {MgvLib, MgvStructs, IMaker, OLKey} from "./MgvLib.sol";
+import {MgvLib, MgvStructs, IMaker, OLKey, HasMgvEvents} from "./MgvLib.sol";
 import "./MgvLib.sol" as MgvLibWrapper;
 
-interface IMangrove {
-  event Approval(
-    address indexed outbound_tkn,
-    address indexed inbound_tkn,
-    uint indexed tickScale,
-    address owner,
-    address spender,
-    uint value
-  );
-  event Credit(address indexed maker, uint amount);
-  event Debit(address indexed maker, uint amount);
-  event Kill();
-  event NewMgv();
-  event OfferFail(bytes32 indexed olKeyHash, uint id, uint takerWants, uint takerGives, uint penalty, bytes32 mgvData);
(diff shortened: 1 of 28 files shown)


src/MgvRoot.sol
```diff
@@ -138,12 +138,12 @@ contract MgvRoot is HasMgvEvents {
   function configInfo(OLKey memory olKey)
     external
     view
-    returns (MgvStructs.GlobalUnpacked memory global, MgvStructs.LocalUnpacked memory local)
+    returns (MgvStructs.GlobalUnpacked memory _global, MgvStructs.LocalUnpacked memory _local)
```
espendk, line 141: This looks weird - why not avoid the `_` ?

src/IMangrove.sol
```diff
@@ -8,70 +8,51 @@ pragma solidity >=0.7.0 <0.9.0;
...
+  function config(OLKey memory olKey)
+    external
+    view
+    returns (MgvStructs.GlobalPacked _global, MgvStructs.LocalPacked _local);
 
   function configInfo(OLKey memory olKey)
     external
     view
-    returns (MgvStructs.GlobalUnpacked memory global, MgvStructs.LocalUnpacked memory local);
+    returns (MgvStructs.GlobalUnpacked memory _global, MgvStructs.LocalUnpacked memory _local);
```
espendk, line 41: Again, why the underscores?

espendk (commented):
Great!

One question: What do we win by writing tests against the interface when the ABI is checked?

src/MgvRoot.sol
lnist: We have _global in other function signatures as well. My editor doesn't like just "global" and I thought that was why it was introduced elsewhere, so I aligned it. It is not super important whether we go with or without _, but I think we should be consistent.

src/IMangrove.sol
lnist: This is the same signature as for your other comment. So they are just identical :)

"config" was the one with already existing _global.

src/MgvRoot.sol
espendk: Agree on the consistency. Unless there's a reason to have the `_` I think we should avoid it.

lnist: global is a reserved keyword. Agreed with @adhusson to keep _.

lnist:
Writing tests against IMangrove means being able to navigate to natspec.

PR #488, opened 2023-09-01
feat: remove posthookFail event

peterMangrove:
This PR removes the event PosthookFail. Instead OfferSuccess and OfferFail now has 2 versions, one with posthookData and one without, and we then emit the one with the posthook data, when the posthook fails.

The reason for this change, is to make the worst case more gas efficient. With this change we now emit only one event, either OfferSuccess or OfferFail, when a posthook fails. Where it used to emit both the PosthookFail event and the Offer{Success,Fail} event. We can therefore save a little gas and it intuitively makes more sense to have one, instead of two, when we can. 

Running a test where the posthook fails, with and without these changes, showed a small gas change of ~600, meaning we save ~600 gas in this case.


espendk (approved):
lgtm!

PR #496, opened 2023-09-02
fix: port permit_typehash to new arg names + test

adhusson:
permit_typehash was the hash with base,quote names for token args (instead of outbound_tkn,inbound_tkn)


espendk (approved):
nice catch!

PR #499, opened 2023-09-02
Reduce Mangrove contract size below the limit

### LINKED ISSUE #470: Reduce Mangrove contract size below limit
For now, Mangrove's contract size is too big for the ~24kB limit.

1. One way to reducing it would be to move all governance functions outside and just have a generic "modify config" function inside Mangrove.
2. Another would be to move all view functions outside and have a generic `execute arbitrary code` function inside Mangrove.

For the second point, the idea is: have Mangrove implement a function that runs arbitrary code but then reverts. So you can do anything, revert, and return the data you got. In pseudocode):

```
// in Mangrove
function run_code(address implem, bytes call_data) {
  (success, revert_data) = delegatecall(address(this),abi.encodeCall(this.run_code_and_revert,(implem,call_data)));
}

function run_code_and_revert(address implem, bytes call_data) {
  (success,return_data) = delegatecall(implem,data);
  revert(return_data);
}
```

Now if you have a MangroveViewFunctions contract with the same storage layout as Mangrove at address `0xViewFunctions`, you can do:

```
bytes memory view_call_data = abi.encodeCall(MangroveViewFunctions.viewSomething,(...args));
(success, view_data) = staticall(address(mgv),abi.encodeCall(Mangrove.run_code,(mgv_view_functions,view_call_data)));
```

adhusson:
⚠️ This PR mistakenly contains commit dff4515ba8a75d7deade41ee980997722f61a341 from #509. For future reference, you should look at [the fixed version of this branch](https://github.com/mangrovedao/mangrove-core/tree/feat/reduce-codesize-fixed).

Closes #470. Mangrove is now below the contract size limit with 200 runs on both the normal and via_ir pipeline.

This PR create a satellite `MgvAppendix` contract that contains all view&governance functions. `Mangrove` now has a `fallback` function that proxies to `MgvAppendix` code. The proxying functionality is in `AbstractMangrove`.

The change means we consolidate state variables + some functions in a new `MgvCommon` contract. It also contains functions that are used by both `Mangrove` and `MgvAppendix`.

`MgvRoot` has been removed: most of its content is now in `MgvCommon`, and the remaining functions are in `MgvHasOffers`.

`MgvGovernable` has also been removed, with its content moved to `MgvAppendix`.

So the contract structure is now:

```

                                   HasMgvEvents
                                        │
                                        │
                                        │
                                    MgvCommon
                                        │
                       ┌────────────────┴──────────┐
                       │                           │
                 MgvHasOffers                      │
                       │                           │
           ┌───────────┴──────────┐                │
           │                      │                │
    MgvOfferTaking         MgvOfferMaking          │
           │                      │                │
           │                      │                │
MgvOfferTakingWithPermit          │                │
           │                      │                │
           └───────────┬──────────┘                │
                       │                           │
                       │                           │
                AbstractMangrove                   │
                       │                           │
                       ▼                           ▼
                    Mangrove                  MgvAppendix
```

Note: For the `IMangrove`/`Mangrove` ABI comparison test to work, we create a dummy `MangroveMix` contract that inherits from both `Mangrove` and `MgvAppendix`, and match its ABI with the ABI of `IMangrove`.

TODO:
- [x] Remove `marketOrder[For]ByPrice`
- [ ] Consider merging part of `MgvReader` with `MgvAppendix` (all functions that do not depend on `MgvReader`'s extra state)


src/AbstractMangrove.sol
```diff
@@ -5,12 +5,41 @@ import {MgvLib} from "./MgvLib.sol";
...
 import {MgvOfferTakingWithPermit} from "./MgvOfferTakingWithPermit.sol";
-import {MgvGovernable} from "./MgvGovernable.sol";
+import {MgvAppendix} from "mgv_src/MgvAppendix.sol";
 
-/* `AbstractMangrove` inherits the three contracts that implement generic Mangrove functionality (`MgvGovernable`,`MgvOfferTakingWithPermit` and `MgvOfferMaking`) but does not implement the abstract functions. */
-abstract contract AbstractMangrove is MgvGovernable, MgvOfferTakingWithPermit, MgvOfferMaking {
-  constructor(address governance, uint gasprice, uint gasmax, string memory contractName)
+/* `AbstractMangrove` inherits the three contracts that implement generic Mangrove functionality (`MgvOfferTakingWithPermit` and `MgvOfferMaking`) but does not implement the abstract functions. */
+abstract contract AbstractMangrove is MgvOfferTakingWithPermit, MgvOfferMaking {
+  address immutable appendix;
```
jkrivine: appendix <- APPENDIX

src/MgvCommon.sol
```diff
@@ -0,0 +1,142 @@
...
+
+   The contract structure is as follows:
+   <img src="./modular_mangrove.svg" width="180%"> </img>
+ */
+
+pragma solidity ^0.8.10;
+
+import {MgvStructs, Field, OLKey, HasMgvEvents, Density, IMgvMonitor, IERC20, Leaf} from "./MgvLib.sol";
+
+/* `MgvRoot` contains state variables used everywhere in the operation of Mangrove and their related function. */
```
espendk, line 25: `MgvRoot` is gone, should be `MgvCommon` right?

src/MgvCommon.sol
```diff
@@ -0,0 +1,142 @@
...
+
+       Offers specify the amount of gas they require for successful execution ([`gasreq`](#structs.js/gasreq)). To minimize book spamming, market makers must provision a *penalty*, which depends on their `gasreq` and on the offerList's [`offer_gasbase`](#structs.js/gasbase). This provision is deducted from their `balanceOf`. If an offer fails, part of that provision is given to the taker, as retribution. The exact amount depends on the gas used by the offer before failing.
+
+       The Mangrove keeps track of their available balance in the `balanceOf` map, which is decremented every time a maker creates a new offer, and may be modified on offer updates/cancellations/takings.
+     */
+  mapping(address maker => uint balance) internal _balanceOf;
+
+  /* This code exists in 2 copies one in Mangrove and one in MgvViewFns.
+     We could have it only in Mangrove and get MgvViewFns to always call mgv.config but would be an extra call to mangrove for a lot of view functions, which can be avoided by just duplicating the code.
+  */
```
espendk, line 74: There's no `MgvViewFns` ?

src/MgvCommon.sol
lnist, line 25: MgvRoot?

src/MgvAppendix.sol
```diff
@@ -1,27 +1,131 @@
...
+  function offerDetails(OLKey memory olKey, uint offerId)
+    external
+    view
+    returns (MgvStructs.OfferDetailPacked offerDetail)
+  {
+    return offerLists[olKey.hash()].offerData[offerId].detail;
+  }
+
+  /* Returns information about an offer in ABI-compatible structs. Do not use internally, would be a huge memory-copying waste. Use `offerLists[outbound_tkn][inbound_tkn].offers` and `offerLists[outbound_tkn][inbound_tkn].offerDetails` instead. */
+  function offerInfo(OLKey memory olKey, uint offerId)
```
lnist: I assume all functions here are simply cut-pasted and thus have not reviewed them.
And all the auth stuff still applies the same due to delegate call.
Are there any special attack vectors we should think about with this contract-split that did not apply before?

src/MgvCommon.sol
```diff
@@ -0,0 +1,142 @@
...
+  }
+  // Also duplicated because MgvAppendix uses transferToken & we want to keep them together
+
+  /* # Token transfer functions */
+  /* `transferTokenFrom` is adapted from [existing code](https://soliditydeveloper.com/safe-erc20) and in particular avoids the
+    "no return value" bug. It never throws and returns true iff the transfer was successful according to `tokenAddress`.
+
+      Note that any spurious exception due to an error in Mangrove code will be falsely blamed on `from`.
+    */
+  function transferTokenFrom(address tokenAddress, address from, address to, uint value) internal returns (bool) {
```
lnist, line 109: Sorry if we have discussed before, but should we use TransferLib here?

test/core/IMangroveAbi.t.sol
```diff
@@ -3,23 +3,30 @@ pragma solidity ^0.8.10;
...
-    string memory mangrovePath = string.concat(root, "/out/Mangrove.sol/Mangrove.json");
+    string memory mangrovePath = string.concat(root, "/out/IMangroveAbi.t.sol/MangroveMix.json");
     string memory interfacePath = string.concat(root, "/out/IMangrove.sol/IMangrove.json");
     string memory mangroveJson = vm.readFile(mangrovePath);
     string memory interfaceJson = vm.readFile(interfacePath);
 
-    string[] memory mangroveAbiElementsWithConstructor = stdJson.readStringArray(mangroveJson, ".abi");
-    string[] memory mangroveAbiElements = new string[](mangroveAbiElementsWithConstructor.length-1);
+    string[] memory mangroveAbiElements = stdJson.readStringArray(mangroveJson, ".abi");
+    // string[] memory mangroveAbiElements = new string[](mangroveAbiElementsWithConstructor.length-1);
```
lnist: You did introduce MangroveMix, should the test be enabled again or does it need fixing?

src/AbstractMangrove.sol
```diff
@@ -5,12 +5,41 @@ import {MgvLib} from "./MgvLib.sol";
 
 import {MgvOfferMaking} from "./MgvOfferMaking.sol";
 import {MgvOfferTakingWithPermit} from "./MgvOfferTakingWithPermit.sol";
-import {MgvGovernable} from "./MgvGovernable.sol";
+import {MgvAppendix} from "mgv_src/MgvAppendix.sol";
 
-/* `AbstractMangrove` inherits the three contracts that implement generic Mangrove functionality (`MgvGovernable`,`MgvOfferTakingWithPermit` and `MgvOfferMaking`) but does not implement the abstract functions. */
-abstract contract AbstractMangrove is MgvGovernable, MgvOfferTakingWithPermit, MgvOfferMaking {
-  constructor(address governance, uint gasprice, uint gasmax, string memory contractName)
+/* `AbstractMangrove` inherits the three contracts that implement generic Mangrove functionality (`MgvOfferTakingWithPermit` and `MgvOfferMaking`) but does not implement the abstract functions. */
```
espendk:
```suggestion
/* `AbstractMangrove` inherits the two contracts that implement generic Mangrove functionality (`MgvOfferTakingWithPermit` and `MgvOfferMaking`) but does not implement the abstract functions. */
```

test/core/IMangroveAbi.t.sol
```diff
@@ -3,23 +3,30 @@ pragma solidity ^0.8.10;
...
     // Delete constructor definition as it is not present in interface
-    for (uint i; i < mangroveAbiElements.length; ++i) {
-      mangroveAbiElements[i] = mangroveAbiElementsWithConstructor[i + 1];
-    }
+    // for (uint i; i < mangroveAbiElements.length; ++i) {
+    //   mangroveAbiElements[i] = mangroveAbiElementsWithConstructor[i + 1];
+    // }
```
espendk: Delete?

lnist (commented):
It does make things more complicated, but not by too much, and the constraints have to be considered.

If we need further size reduction, then possibly cleanByImpersonation could be moved since it anyway does an external call internally.

espendk (commented):
It's a nifty solution, but also a bit weird. Not too bad though.

Good idea to mainly pay the extra gas off fallback-delegatecall in governance functions.

I wonder if it would be nicer if the read-functions and governance functions were in separate contracts like before. Something like:

```
                                   HasMgvEvents
                                        │
                                        │
                                        │
                                    MgvCommon
                                        │
                       ┌────────────────┴──────────┬─────────────┐
                       │                           │             │
                 MgvHasOffers                MgvGovernable   MgvView
                       │                           │             │
           ┌───────────┴──────────┐                │             │
           │                      │                │             │
    MgvOfferTaking         MgvOfferMaking          │             │
           │                      │                │             │
           │                      │                │             │
MgvOfferTakingWithPermit          │                │             │
           │                      │                │             │
           └───────────┬──────────┘                └──────┬──────┘
                       │                                  │
                       │                                  │
                AbstractMangrove                          │
                       │                                  │
                       ▼                                  ▼
                    Mangrove                         MgvAppendix
```

test/core/IMangroveAbi.t.sol
lnist: Hopefully not as that would disable the test :D

espendk: huh? Isn't it just the code that removes constructor that has been commented out?

src/MgvCommon.sol
adhusson: Yes, I updated comments haphazardly -- will fix all that in a comments PR

src/MgvAppendix.sol
adhusson: Main attack vectors I know are related to proxy upgrades but here APPENDIX is immutable. 

The state of the MgvAppendix contract can probably be changed by directly calling its functions but that state does not matter. One could add a `require(address(this) == MGV_ADDRESS)` everywhere but it seems like theater more than an actual measure.

src/MgvCommon.sol
adhusson: Not sure, are the functions different? Keeping those functions here because they served us well in the past that's all

test/core/IMangroveAbi.t.sol
adhusson: Test is enabled! I just commented the "remove the constructor" lines but should delete them (MangroveMix has no constructor)

test/core/IMangroveAbi.t.sol
adhusson: yes!

src/MgvCommon.sol
adhusson: Yes, I updated comments haphazardly -- will fix all that in a comments PR

src/AbstractMangrove.sol
adhusson: done in 6a5584b8e81b47b3d87ee08c586416d2cbfed36e

test/core/IMangroveAbi.t.sol
adhusson: cleaned in 99b42f795b73d25e13d77ba653e07b4fbdbd86de

adhusson:
Just noticed that this PR is only compatible with 0.8.21. We may compile with 0.8.21 (although I'd rather use 0.8.20) but anyway it's better to let people use older versions and still import Mangrove.

I pushed 556ccb43ed8926a6fa4ba8ee6376ea37512d42f4 which restores compatibility with previous versions at the cost of consistency (a state var that's not `APPENDIX` is now defined outside of `MgvCommon`)

adhusson:
@espendk I gave a shot at following your diagram in 6cb1e26aa1d7e2d00d68881fb60fac61fef7e2a2 let me know what you think. Only drawbacks are 
- more files 
- `AbstractMangrove` has to write `MgvGovernable.functionName` instead of `MgvAppendix.functionName` (because solidity does not inherit function selectors)

Let me know what you think!

espendk:
> @espendk I gave a shot at following your diagram in [6cb1e26](https://github.com/mangrovedao/mangrove-core/commit/6cb1e26aa1d7e2d00d68881fb60fac61fef7e2a2) let me know what you think. Only drawbacks are
> 
>     * more files
> 
>     * `AbstractMangrove` has to write `MgvGovernable.functionName` instead of `MgvAppendix.functionName` (because solidity does not inherit function selectors)
> 
> 
> Let me know what you think!

I think this structure is clearer. And more, simpler and well-scoped files and the `MgvGovernable.functionName` instead of `MgvAppendix.functionName` is an upside an IMO. 
But I don't feel strongly about it, so I'll let you decide :-)

espendk (approved):
lgtm!

lnist (approved):

PR #505, opened 2023-09-04
feat(MgvReader): Update market order simulation to match new tick-based algorithm

espendk:


adhusson (approved):
looks good to me.

I'm trying to think of a way to let Mangrove simulate that. The only part that needs changing is the "transfer tokens" part, right?

espendk:
> I'm trying to think of a way to let Mangrove simulate that. The only part that needs changing is the "transfer tokens" part, right?

Umm, not sure. Any state change needs to be avoided, including not updating the offer list :-)

Maybe we could execute the order and then revert with the result?

But I'm not sure this should be a priority - I'm unsure if this feature will end up being used in practice.

src/periphery/MgvReader.sol
```diff
@@ -256,50 +283,42 @@ contract MgvReader {
...
-    external
-    view
-    returns (VolumeData[] memory)
-  {
-    return marketOrder(olKey, takerWants, takerGives, fillWants, true);
-  }
-
-  function internalMarketOrder(MarketOrder memory mr, bool proceed) internal view {
+  function internalMarketOrder(MarketOrder memory mr) internal view {
+    console.log("internalMO");
```
lnist, line 298: are these logs intentional?

lnist (commented):
This saves a lot of gas compared to a call to the actual market order - and does not require tokens for transfers.  But it does come at the cost of logic duplication and imprecision due to not executing the real order.
I am not sure what scenario we see it used, but for now it is probably OK.

src/periphery/MgvReader.sol
espendk: doh! I'll remove them in another PR

espendk: PR to fix this here: https://github.com/mangrovedao/mangrove-core/pull/507

PR #509, opened 2023-09-04
Refactor/merge tree walkers

adhusson:
Attempt to merge `getNextBest` into `dislodgeOffer`.


src/MgvOfferTaking.sol
```diff
@@ -235,7 +188,8 @@ abstract contract MgvOfferTaking is MgvHasOffers {
           2. `sor.gives` was at most `mor.initialGives - mor.totalGave` from earlier step,
           3. `sor.gives` may have been clamped _down_ during `execute` (to "`offer.wants`" if the offer is entirely consumed, or to `makerWouldWant`, cf. code of `execute`).
         */
-        (sor.offerId, sor.local) = getNextBest(offerList, mor, sor.offer, sor.local, sor.olKey.tickScale);
+        (mor.leaf, sor.local) = dislodgeOffer(offerList, sor.olKey.tickScale, offer.prev(0), sor.local, true, mor.leaf);
+        sor.offerId = mor.leaf.getNextOfferId();
```
espendk, line 180: I like how this makes it clearer that we're removing `offer` and then getting the next `offerId`.

espendk (approved):
I think this is better than before, nice.

`dislodgeOffer` is not easier to follow now, but the code that used to call `getNextBest` is clearer IMO.

I'd like so see the tick tree tests validate this change, but that can happen after this is merged :-)

adhusson:
With classic pipeline and runs=200, I'm seeing a ~1.4k-per-offer gas cost increase with this merge, I'm looking into it

adhusson:
Closing (resulting single fn is too complex)

PR #514, opened 2023-09-06
Feat/cache best tick

adhusson:
It's unsafe to recompute bestTick every time in `dislodgeOffer` (and wasteful)

### DIFF
--- a/src/MgvHasOffers.sol
+++ b/src/MgvHasOffers.sol
@@ -67,6 +67,7 @@ contract MgvHasOffers is MgvCommon {
     uint tickScale,
     MgvStructs.OfferPacked offer,
     MgvStructs.LocalPacked local,
+    Tick bestTick,
     bool shouldUpdateBranch
   ) internal returns (MgvStructs.LocalPacked) {
     unchecked {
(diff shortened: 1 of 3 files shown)


src/MgvHasOffers.sol
```diff
@@ -67,6 +67,7 @@ contract MgvHasOffers is MgvCommon {
     uint tickScale,
     MgvStructs.OfferPacked offer,
     MgvStructs.LocalPacked local,
+    Tick bestTick,
```
espendk, line 70: Is it more efficient/less gas costly to take this as a separate parameter than to do `bestTick = local.bestTick()` in the beginning of the function?

In all calls to `dislodgeOffer` we have `bestTick == local.bestTick()`, so this parameter isn't strictly needed (but may save gas?).

src/MgvOfferMaking.sol
```diff
@@ -323,7 +323,8 @@ contract MgvOfferMaking is MgvHasOffers {
         // bool updateLocal = tick.strictlyBetter(ofp.local.bestTick().strictlyBetter(tick)
         bool shouldUpdateBranch = !insertionTick.strictlyBetter(ofp.local.bestTick());
 
-        ofp.local = dislodgeOffer(offerList, ofp.olKey.tickScale, ofp.oldOffer, ofp.local, shouldUpdateBranch);
+        ofp.local =
+          dislodgeOffer(offerList, ofp.olKey.tickScale, ofp.oldOffer, ofp.local, cachedLocalTick, shouldUpdateBranch);
```
espendk, line 327: At this point `cachedLocalTick == ofp.local.bestTick()`, so it's not strictly necessary to pass `bestTick` separately.

espendk (commented):
lgtm, just have a question about whether having `bestTick` as a parameter of `dislodgeOffer` is more efficient than caching it at the beginning of the function?

src/MgvHasOffers.sol
adhusson: It's already computed before one of the calls (in `writeOffer`) so it prevents re-computing it. I haven't measured though

src/MgvOfferMaking.sol
adhusson: I didn't understand that

espendk: I was just arguing that we don't __need__ to pass `bestTick` to `dislodgeOffer` as mentioned in the comment above :-)

espendk (approved):
I don't think we should spend time on the small question I asked, so let's merge this :-)

PR #515, opened 2023-09-06
Expand min/max tick POC 1: "Big Level 2"

adhusson:
Try "one big level2"

### `local` does not contain `level2` anymore
`local` contains:
  - `tickPosInLeaf`
  - `level0`
  - `level1`
  - `tickPosInLevel2`

### `level2` has its own storage slot
The extra read cost only occurs when it's necessary to read `level2`. That's when the price has deviated enough from the initial price to switch to a new level 1. On average that's when the price is multiplied/divided by 2. By expanding `level0` and `level1` we can raise that to ⨉÷6.

## 👍 Pros
_At this stage it's a POC and level0,1 are still 64 bits (they could be more) and level2 is still 64 (it has to be more)_

Code change is small. `level0` and `level1` can be expanded so `posInLevel2` (aka ~`level1Index`) changes less often.

## 👎 Cons
Every time there's an update to the single `level2` node, there is an additional SSTORE.

# A note on expanding min/max tick:
"Not expanding the ticktree" is a possible answer. 
Importantly, as soon as tickScale>1 we already have more ticks than necessary to cover all price points.

So the ticktree can be too small only for token pairs where tickScale=1 is the right fit. This concerns assets with very little expected price variation. Asset examples that would fit the bill:
* Asset1 is supposed to represent $100k, and is on 6 decimals. Asset2 is supposed to represent $0.00001, and is on 18 decimals.
* Asset1 and Asset2 are supposed to represent $1, one is on 23 decimals, the other is on 1 decimals.

Here are the main arguments for/against expanding the ticktree I have:

#### For
One sharp edge removed, aka lower cognitive complexity, aka we don’t have to think about this again

#### Against
+code complexity, +gas consumption 

# Other approaches 
_taking opinions on what's good here! Will do POCs for some_
- Add a `level3` inside `local`: it would be 2 or 4 bits wide. Code might be simple. #529
- Add a `level3` in a new slot: it would only be read in marketOrder when the price does ⨉÷50e22, but you still need to encode `tickPosInLevel3` somewhere in `local` and not how to do that cleanly. Code might be simpler than this PR though.
- Add a "level 2 window": `local` contains a window of level2 which is actually 256 bits. The window is shifted when needed. Code will probably be more complex than this PR.
- Add a `tickShift` param to `offerList`, means we don't have to touch tree-walking code, but it adds overhead spread over the lifetime of the governance of the project.


test/lib/TestTickTree.sol
```diff
@@ -45,9 +46,9 @@ contract TestTickTree is MangroveTest {
   // Creates a snapshot of the Mangrove tick tree
   function snapshotMgvTickTree() public {
     local = reader.local(olKey);
-    Field level2 = mgv.level2(olKey);
+    Field _level2 = mgv.level2(olKey);
```
espendk: I think this should be written to `level2` directly, otherwise level2 isn't included in the snapshot (which could explain why the tests fail):

```
level2 = mgv.level2(olKey);
```

The remaining changes in this function can be removed.

espendk: I pushed a fix and reenabled the tests: 50bfeb09f5416030e8f62faf33e8b881151d2005

adhusson:
Closed in favor or #529

Issue #471, opened 2023-08-28
Update MgvReader (multiple features)


adhusson:
MgvReader has obsolete reimplementations of `getNextBest` and `marketOrder`, and maybe other functions.

- [x] ~~`getNextBest` maybe be implemented using the method described in #470~~
	- abandoned, see #516 
- [x] `marketOrder` probably needs to be reimplemented because the MgvReader's version of the function does not actually try to transfer tokens

PR #516, opened 2023-09-07
refactor: Clearer split between `MgvReader` and `MgvView`

espendk:
# `MgvReader` VS `MgvView`
As part of the effort to deduplicate some tree-walking code in `MgvReader`, I tried moving all non-Market view functions from `MgvReader` to `MgvView`: [eff3ad5](https://github.com/mangrovedao/mangrove-core/commit/eff3ad5d5bdfb4ef9fefda34c5f5099455cf780e).

However, I don’t like the result: There’s now a big bunch of utility functions in `(I)Mangrove` that clutter the API.
Also, where do we draw the line for what to put in `MgvView`?

The scope of `MgvReader` and `MgvView` in that commit can be characterized as:
- `MgvView` : View and utility functions for Mangrove.
- `MgvReader`: Tracking of open markets.

Perhaps a more sensible split of scope could be based on whether a view function needs access to Mangrove’s internal state or functions:

- `MgvView`: View functions for Mangrove that require access to internal state or functions.
- `MgvReader`: View & utility functions that (1) do not require direct access to internal state or functions or (2) require extra state (such as tracking open markets).

An upside of this is also that we don't risk running into the `initcode` size limit: https://eips.ethereum.org/EIPS/eip-3860.

The downside of this is that `MgvReader` remains a necessity.

We decided to go for this second option.

# Misc
I took the chance to do a bit of cleanup:
- organize `IMangrove`, `MgvView`, and `MgvReader` into sections of related methods
- add `unchecked` to all methods in `MgvView`
- make all methods in `MgvView` `external`

# **[CANCELLED]** Deduplicating `MgvReader.nextOfferId` by calling `Mangrove.getNextBest` and reverting
The code in `MgvReader.nextOfferId` is nearly identical to `Mangrove.getNextBest` so we were hoping replace it with a call to Mangrove.

The approach was:

1. Move all non-Market functions from `MgvReader` to `MgvView`
2. Move `getNextBest()` to `MgvCommon` so `MgvView` can call it
3. Implement a `MgvView.nextOfferIdByRevert()` function which calls `getNextBest()` and then reverts with the result.

However, we ran into the following issues:
- Becuase `MgvView.nextOfferIdByRevert()` modifies state, the compiler will not accept a `view` modifier for it, despite it always reverting. This is because `view` functions are called with `STATICCALL` and the EVM reverts if they execute state modifying operations.
  - This propagates non-`view` to A LOT of functions that are view, so that’s not acceptable.
- Getting `getNextBest()` to return the next id starting from any offer seems tricky.
  - @adhusson thinks it's not that bad.
  - I wonder if the complexity we save by this deduplication scheme is higher than what we gain?
  - We don’t want to do anything that makes `getNextBest()`  use more gas, so our options are limited..

Based on this, we decided not to deduplicate this.

This closes https://github.com/mangrovedao/mangrove-core/issues/471

### DIFF
--- a/src/periphery/MgvReader.sol
+++ b/src/periphery/MgvReader.sol
@@ -17,6 +17,12 @@ struct MarketConfig {
   MgvStructs.LocalUnpacked config10;
 }
 
+struct VolumeData {
+  uint totalGot;
+  uint totalGave;
+  uint totalGasreq;
+}
+
 /// @notice We choose a canonical orientation for all markets based on the numerical values of their token addresses. That way we can uniquely identify a market with two addresses given in any order.
 /// @return address the lowest of the given arguments (numerically)
 /// @return address the highest of the given arguments (numerically)
@@ -119,8 +125,8 @@ contract MgvReader {
         markets[i] = Market({tkn0: tkn0, tkn1: tkn1, tickScale: tickScale});
 
         if (withConfig) {
-          configs[i].config01 = MGV.localUnpacked(toOLKey(markets[i]));
-          configs[i].config10 = MGV.localUnpacked(toOLKey(flipped(markets[i])));
+          configs[i].config01 = localUnpacked(toOLKey(markets[i]));
+          configs[i].config10 = localUnpacked(toOLKey(flipped(markets[i])));
         }
       }
     }
@@ -139,15 +145,15 @@ contract MgvReader {
   /// @return config The market configuration. config01 and config10 follow the order given in arguments, not the canonical order
   /// @dev This function queries Mangrove so all the returned info is up-to-date.
   function marketConfig(Market memory market) external view returns (MarketConfig memory config) {
-    config.config01 = MGV.localUnpacked(toOLKey(market));
-    config.config10 = MGV.localUnpacked(toOLKey(flipped(market)));
+    config.config01 = localUnpacked(toOLKey(market));
+    config.config10 = localUnpacked(toOLKey(flipped(market)));
   }
 
   /// @notice Permisionless update of _openMarkets array.
   /// @notice Will consider a market open iff either the offer lists tkn0/tkn1 or tkn1/tkn0 are open on Mangrove.
   function updateMarket(Market memory market) external {
     (market.tkn0, market.tkn1) = order(market.tkn0, market.tkn1);
-    bool openOnMangrove = MGV.local(toOLKey(market)).active() || MGV.local(toOLKey(flipped(market))).active();
+    bool openOnMangrove = local(toOLKey(market)).active() || local(toOLKey(flipped(market))).active();
     uint position = marketPositions[market.tkn0][market.tkn1][market.tickScale];
 
     if (openOnMangrove && position == 0) {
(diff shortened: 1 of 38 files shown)


adhusson:
Agree with not deduplicating! Also I had misunderstood about `getNextBest`, I think now without deduplication, the existing  `nextOfferId` gives us what we want, right?

espendk:
> I think now without deduplication, the existing `nextOfferId` gives us what we want, right?

I believe it does. It's used in a number of tests that pass, so I think it's good.

test/core/Gatekeeping.t.sol
```diff
@@ -703,6 +703,7 @@ contract GatekeepingTest is MangroveTest {
     assertTrue(mgv.locked(olKey), "market must be locked");
     vm.expectRevert("mgv/reentrancyLocked");
     mgv.config(olKey);
+    mgv.global();
```
adhusson: Leftover line?

test/core/Gatekeeping.t.sol
```diff
@@ -739,15 +738,15 @@ contract GatekeepingTest is MangroveTest {
   function olReadOK(OLKey memory olKey) external {
     assertTrue(!mgv.locked(olKey), "market must not be locked");
     mgv.config(olKey);
+    mgv.global();
```
adhusson, line 744: or if it's intentional I don't get it 😅

src/periphery/MgvReader.sol
```diff
@@ -87,6 +88,57 @@ contract MgvReader {
...
+      _global = __global.to_struct();
+      _local = __local.to_struct();
+    }
+  }
+
+  /* Sugar for getting only global/local config */
+  function local(OLKey memory olKey) public view returns (MgvStructs.LocalPacked) {
+    (, MgvStructs.LocalPacked _local) = MGV.config(olKey);
+    return _local;
+  }
```
adhusson: Since the access to packed global is in MgvView it increases surprise to have local here, is it OK to move it to MgvView?

adhusson (commented):
This is really nice :)

test/core/Gatekeeping.t.sol
espendk: It's intentional: I want to ensure that the read-lock doesn't affect `global()` :-)
Maybe deserves a comment?

The functions are called in the order they are listed in `MgvView` for easy comparison and maintenance.

src/periphery/MgvReader.sol
espendk: Agreed, will do.

I tried to minimize `MgvView`, but this one felt weird to when I moved it back to `MgvReader`, which is a sign that you're right :-)

espendk: Addressed in a24822f19c4b277facd56a3d25a6690a03723944

test/core/Gatekeeping.t.sol
espendk: Addressed in ffa58ba13c4d283e5f57df184a6dcc625b8b66e6

test/core/Gatekeeping.t.sol
espendk: nope, see below :-)

lnist (approved):

adhusson (approved):
All good :)

PR #521, opened 2023-09-08
Feat/cheaper first insert & single offer move

adhusson:
Reduce worst case cost by thousands of gas (2k - 8k) at the cost of ~200 gas increase with low optim

### DIFF
--- a/src/MgvOfferMaking.sol
+++ b/src/MgvOfferMaking.sol
@@ -304,7 +304,13 @@ contract MgvOfferMaking is MgvHasOffers {
       require(TickLib.inRange(insertionTick), "mgv/writeOffer/tick/outOfRange");
 
       // must cache tick because branch will be modified and tick information will be lost (in case an offer will be removed)
-      Tick cachedLocalTick = ofp.local.bestTick();
+      Tick cachedLocalTick;
+      // force control flow through gas saving path if offer list is empty
+      if (ofp.local.level0().isEmpty()) {
+        cachedLocalTick = insertionTick;
+      } else {
+        cachedLocalTick = ofp.local.bestTick();
+      }
 
       // remove offer from previous position
       if (ofp.oldOffer.isLive()) {
@@ -326,11 +332,17 @@ contract MgvOfferMaking is MgvHasOffers {
         ofp.local =
           dislodgeOffer(offerList, ofp.olKey.tickScale, ofp.oldOffer, ofp.local, cachedLocalTick, shouldUpdateBranch);
         // If !shouldUpdateBranch, then ofp.local.level0 and ofp.local.level1 reflect the removed tick's branch post-removal, so one cannot infer the tick by reading those fields. If shouldUpdateBranch, then the new tick must be inferred from the new info in local.
+        // FIXME check if shouldUpdateBranch was true but became false inside dislodgeOffer
         if (shouldUpdateBranch) {
-          cachedLocalTick = ofp.local.bestTick();
+          // force control flow through gas-saving path if retraction emptied the offer list
+          if (ofp.local.level0().isEmpty()) {
+            cachedLocalTick = insertionTick;
+          } else {
+            cachedLocalTick = ofp.local.bestTick();
+          }
         }
       }
-      if (insertionTick.strictlyBetter(cachedLocalTick)) {
+      if (!cachedLocalTick.strictlyBetter(insertionTick)) {
         ofp.local = ofp.local.tickPosInLeaf(insertionTick.posInLeaf());
       }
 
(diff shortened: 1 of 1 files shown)


src/MgvOfferMaking.sol
```diff
@@ -399,6 +411,7 @@ contract MgvOfferMaking is MgvHasOffers {
 
       // store offer at the end of the tick
       leaf = leaf.setTickLast(insertionTick, ofrId);
+      // FIXME check whether write necessary: when offer neither first nor last of tick
```
espendk: How can the offer be not last? In the line before, we set it to be last?

Maybe the idea is not to write if the offer was already last in that tick? Ie an update that doesn't change the tick.

src/MgvOfferMaking.sol
```diff
@@ -304,7 +304,13 @@ contract MgvOfferMaking is MgvHasOffers {
...
+      Tick cachedLocalTick;
+      // force control flow through gas saving path if offer list is empty
+      if (ofp.local.level0().isEmpty()) {
+        cachedLocalTick = insertionTick;
+      } else {
+        cachedLocalTick = ofp.local.bestTick();
+      }
 
       // remove offer from previous position
       if (ofp.oldOffer.isLive()) {
```
espendk: These if-stmts can be merged: If the offer list is empty, the offer cannot be live.

src/MgvOfferMaking.sol
```diff
@@ -304,7 +304,13 @@ contract MgvOfferMaking is MgvHasOffers {
       require(TickLib.inRange(insertionTick), "mgv/writeOffer/tick/outOfRange");
 
       // must cache tick because branch will be modified and tick information will be lost (in case an offer will be removed)
-      Tick cachedLocalTick = ofp.local.bestTick();
+      Tick cachedLocalTick;
+      // force control flow through gas saving path if offer list is empty
```
espendk, line 308: Would be nice with an explanation of how this works.

If I understand correctly, the trick is that `cachedLocalTick == insertionTick  =>  insertionIndex == currentIndex` which means we don't read&write level{0,1} in mappings below.

espendk: I think this also means that we don't need to check whether `level2` is empty in the code below (GH won't let me comment on the lines in question, it's L369 and L389).

Ie I think we can change
```
if (insertionIndex <= currentIndex || ofp.local.level2().isEmpty()) {
```

into 
```
if (insertionIndex <= currentIndex) {
```

since `ofp.local.level2().isEmpty()  =>  ofp.local.level0().isEmpty()  =>  cachedLocalTick == insertionTick  =>  insertionIndex == currentIndex`.

src/MgvOfferMaking.sol
```diff
@@ -326,11 +332,17 @@ contract MgvOfferMaking is MgvHasOffers {
...
+          // force control flow through gas-saving path if retraction emptied the offer list
+          if (ofp.local.level0().isEmpty()) {
+            cachedLocalTick = insertionTick;
+          } else {
+            cachedLocalTick = ofp.local.bestTick();
+          }
         }
       }
-      if (insertionTick.strictlyBetter(cachedLocalTick)) {
+      if (!cachedLocalTick.strictlyBetter(insertionTick)) {
```
lnist, line 344: why also update if they are equal?

src/MgvOfferMaking.sol
adhusson: The idea was « it’s 2am and I forgot we append to lists » 😄

src/MgvOfferMaking.sol
adhusson: But the implication is not true the other way? How would you write it

src/MgvOfferMaking.sol
adhusson: Yes I forgot to remove those! Will do when I’m back

src/MgvOfferMaking.sol
adhusson: We encode « empty book » with « insertion tick = current tick ». I tried maintaining the original condition and add « OR level0 is empty » and I got very mysterious gas increases from a couple hundred to thousand ds in some tests. I let it be but it’s a mystery open for any interested detective.

src/MgvOfferMaking.sol
espendk:
```suggestion
      if (ofp.local.level0().isEmpty()) {
        cachedLocalTick = insertionTick;
      } else
        cachedLocalTick = ofp.local.bestTick();
        // remove offer from previous position
        if (ofp.oldOffer.isLive()) {
```

adhusson: nice!

adhusson: nice! addressed in 006d6c3c31462e8ba582050c7baa34a69d9033dc

src/MgvOfferMaking.sol
adhusson: Addressed in d077234bb08e93993c783350582c05565c014869 :)

src/MgvOfferMaking.sol
espendk: This still needs to be fixed :-)

espendk (approved):
lgtm when the wrong FIXME is removed :-)

PR #526, opened 2023-09-08
Feat/return update branch status

adhusson:
for consistency, check that the tree branch was actually updated before recomputing

also, opportunistically: use cached local tick in one more place


adhusson:
it adds ~50 gas in many tests but saves ~600 when branch is not updated even though caller of `dislodgeOffer` thought it would be updated

espendk (commented):
lgtm!

PR #527, opened 2023-09-08
Feat/ol key inverse mapping

adhusson:
* Add a `_olKeys: olKeyHash -> olKey` mapping for markets that have been activated at least once
* Expand SetActive to add the contents of `olKey`
  * `SetActive(...,false)` for deactivation also contains the info. Otherwise suppose there is an event `SetInactive(hash)` with no other info. An indexer that wants to know the state of all offerLists that relate to a specific token would need to do 2 passes: 1) collect all markets ever activated, save their hashes, 2) list all `SetInactive` events with that hash. 
  
It will be later easy to add an inheriting contract `Layer2Mangrove` (or to just update Mangrove) so that it contains functions such as
  
```solidity
  function marketOrderByLogPriceForL2(bytes32 olKeyHash, ...) public returns (...) {
    return generalMarketOrder(_olKeys[olKeyHash], maxLogPrice, fillVolume, fillWants, msg.sender, maxGasreqForFailingOffers);
  }
```

### DIFF
--- a/src/IMangrove.sol
+++ b/src/IMangrove.sol
@@ -120,9 +120,9 @@ interface IMangrove is HasMgvEvents {
 
   function best(OLKey memory olKey) external view returns (uint offerId);
 
-  // # Offer view functions
-
   function olKeys(bytes32 olKeyHash) external view returns (OLKey memory olKey);
+
+  // # Offer view functions
 
   function offers(OLKey memory olKey, uint offerId) external view returns (MgvStructs.OfferPacked offer);
 
--- a/src/IMangrove.sol
+++ b/src/IMangrove.sol
@@ -122,6 +122,8 @@ interface IMangrove is HasMgvEvents {
 
   // # Offer view functions
 
+  function olKeys(bytes32 olKeyHash) external view returns (OLKey memory olKey);
+
   function offers(OLKey memory olKey, uint offerId) external view returns (MgvStructs.OfferPacked offer);
 
   function offerDetails(OLKey memory olKey, uint offerId)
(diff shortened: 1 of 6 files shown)


lnist (approved):

src/MgvView.sol
```diff
@@ -108,6 +108,13 @@ contract MgvView is MgvCommon {
 
   // # Offer view functions
 
+  /* Get the olKey that corresponds to a hash, only works for offerLists that have been activated > 0 times */
+  function olKeys(bytes32 olKeyHash) external view returns (OLKey memory olKey) {
+    unchecked {
+      olKey = _olKeys[olKeyHash];
+    }
+  }
+
```
espendk, line 115: I'd say this is an offer LIST view function, so should be listed with those ;-)

src/IMangrove.sol
```diff
@@ -122,6 +122,8 @@ interface IMangrove is HasMgvEvents {
 
   // # Offer view functions
 
+  function olKeys(bytes32 olKeyHash) external view returns (OLKey memory olKey);
+
```
espendk, line 124: Same here

espendk (changes requested):
nice, except for messing up my newly established order 😈😅

src/MgvView.sol
adhusson: fixed! 141636413fb7a17b04097c3cde56559e5c124dde

src/IMangrove.sol
adhusson: fixed! 141636413fb7a17b04097c3cde56559e5c124dde

espendk (approved):

PR #529, opened 2023-09-08
Expand min/max tick POC 2: "Inline Level 3"

adhusson:
### `local` contains a 2-bit `level3` in addition to the existing levels
This doubles the tick range available (which matches the logprice range).

_I haven't fully update ticktree tests since they depend on tick values that completely change now, so they are `.off` for now_

## 👍 Pros
Code change is copy/paste, so easy to check. Unless you start/end with an empty offer list or get out of the initial `level2`, no significant gas change.

## 👎 Cons
Gaswise, worst case is worse. Unlucky markets with price oscillating around 0 will cost a lot more [^1].

# A note on expanding min/max tick:
"Not expanding the ticktree" is a possible answer. 
Importantly, as soon as tickScale>1 we already have more ticks than necessary to cover all price points.

So the ticktree can be too small only for token pairs where tickScale=1 is the right fit. This concerns assets with very little expected price variation. Asset examples that would fit the bill:
* Asset1 is supposed to represent $100k, and is on 6 decimals. Asset2 is supposed to represent $0.00001, and is on 18 decimals.
* Asset1 and Asset2 are supposed to represent $1, one is on 23 decimals, the other is on 1 decimals.

Here are the main arguments for/against expanding the ticktree I have:

#### For
One sharp edge removed, aka lower cognitive complexity, aka we don’t have to think about this again

#### Against
+code complexity, +gas consumption 

# Other approaches 
_taking opinions on what's good here! Will do POCs for some_
- Have a "big level2": #515
- Add a `level3` in a new slot: it would only be read in marketOrder when the price does ⨉÷50e22, but you still need to encode `tickPosInLevel3` somewhere in `local` and not how to do that cleanly. Code might be simpler than this PR though.
- Add a "level 2 window": `local` contains a window of level2 which is actually 256 bits. The window is shifted when needed. Code will probably be more complex than this PR.
- Add a `tickShift` param to `offerList`, means we don't have to touch tree-walking code, but it adds overhead spread over the lifetime of the governance of the project.

[^1]:  When you go `tick 0` -> `tick -1` (or the other way around), you have to swap your `level0`, `level1` and `level2` in storage, so if you do that all the time you feel it.


lnist:
The extra cost of oscilating around tick 0 also affects maker operations like updateOffer, newOffer, right? If marketOrder is on path warming up tick 0, and they hit tick 1, then they have to account for this in gasreq. Smart makers will account for this in gasreq for offers near tick 0 / sparse markets.

espendk:
The problem of oscillating around tick 0 (ie price=1) was there before as well, but this exacerbates it.

I think this is problematic for stable coin markets with the same decimals such as USDC/USDT or DAI/TUSD. For stable pairs, we'd ideally have the mid price tick located in the middle of all the levels. Tick shifts could achieve this, but I recognize there are downsides to tick shifts as well.

espendk:
PS: Happy to help update the tick tree tests if we deem this approach worth pursuing further.

adhusson:
I have updated the commit to use #530, and to dirty both `level2`s when an offer list activated.

Resulting gas cost increases are more manageable, often ~4k, maxing at +9k for a test that goes across a level2 and takes > 5 offers.

This is a "minimal version". With a level3 in place, we could use those level sizes:

| level  | size |
|--------|------|
| level0 | 88   |
| level1 | 88   |
| level2 | 9    |
| level3 | 9    |

The bigger level0 and level1 mean fewer level switches. 

Then we shift everything so that tick 0 starts in the middle of level3 **and** in the middle of level2. 
* We'll still have all the price range we need (`4 *  88 *  88 * 9 * ⌊9/2⌋ > 2^20 - 1`). 
* Price would have to move by more than ⨉÷490 to reach another level2.

espendk:
> Price would have to move by more than ⨉÷490 to reach another level2.

Can you elaborate on this? I'm not familiar with that notation

src/MgvOfferMaking.sol
```diff
@@ -346,53 +343,89 @@ contract MgvOfferMaking is MgvHasOffers {
...
-          field = offerList.level0[insertionIndex];
+          field = offerList.level0[insertionIndex].clean();
         } else {
           field = ofp.local.level0();
         }
 
         // Save current level0
         if (insertionIndex < currentIndex) {
-          offerList.level0[currentIndex] = ofp.local.level0();
+          Field localLeve0 = ofp.local.level0();
```
espendk:
```suggestion
          Field localLevel0 = ofp.local.level0();
```

src/MgvOfferMaking.sol
```diff
@@ -346,53 +343,89 @@ contract MgvOfferMaking is MgvHasOffers {
...
+          bool shouldSaveLevel0 = !localLeve0.isEmpty();
+          if (!shouldSaveLevel0) {
+            shouldSaveLevel0 = !offerList.level0[currentIndex].eq(DirtyFieldLib.CLEAN_EMPTY);
+          }
+          if (shouldSaveLevel0) {
+            offerList.level0[currentIndex] = localLeve0.dirty();
+          }
```
espendk: Would an `else` be cheaper here?

```suggestion
          if (!shouldSaveLevel0) {
            shouldSaveLevel0 = !offerList.level0[currentIndex].eq(DirtyFieldLib.CLEAN_EMPTY);
          } else {
            offerList.level0[currentIndex] = localLeve0.dirty();
          }
```

src/MgvOfferMaking.sol
```diff
@@ -346,53 +343,89 @@ contract MgvOfferMaking is MgvHasOffers {
...
           if (insertionIndex < currentIndex) {
-            offerList.level1[currentIndex] = ofp.local.level1();
+            Field localLevel1 = ofp.local.level1();
+            bool shouldSaveLevel1 = !localLevel1.isEmpty();
+            if (!shouldSaveLevel1) {
+              shouldSaveLevel1 = !offerList.level1[currentIndex].eq(DirtyFieldLib.CLEAN_EMPTY);
+            }
+            if (shouldSaveLevel1) {
+              offerList.level1[currentIndex] = localLevel1.dirty();
+            }
```
espendk: `else` here as well?

adhusson:
> Can you elaborate on this? I'm not familiar with that notation

Sorry I meant that as an equivalent of "+/-". I meant that, on average, the price would have to be multiplied by more than 490 or divided by more than 490 before the tree switches to the next or previous level2.

src/MgvOfferMaking.sol
adhusson: I don't think this results in the same logic. Suppose that `shouldSaveLevel0=false` and that `offerList.level[currentIndex].eq(DirtyFieldLib.CLEAN_EMPTY)=false`.
* As-is, the first `if` will update `shouldSaveLevel0` to `true`. Then the second `is` will update the `level0` mapping.
* With the suggestion, the first branch of the only `if` will be taken, and then the mapping will not be updated.

espendk: doh! I don't know how I missed that 🤦‍♂️

src/MgvOfferMaking.sol
espendk: this was never here 👋

espendk (approved):
go go go!

PR #530, opened 2023-09-10
Save gas by not zeroing data structures

### LINKED ISSUE #522: Do not zero out data structures for ticks
We currently zero out data structures for ticks - this in theory will result in some discounts, but will then cost extra gas for the next offer hitting the same ticks as given by those structures.

Instead we should leave a bit set on each level, always.

adhusson:
Close #522.

**For the reviewer** The first commit d9509afe196a8fbdad1e03735ffd0a180c26e190 restructures `dislodgeOffer` to save stack space. The second commit d45bc97d9dcfbc7600bbd1836cd3fe4a9c1ebda0 is thus easier to read than the full diff.

# How it works

* Whenever a value `v:Field` is written to a mapping, its top bit is set to 1.
* Whenever a value `v:Leaf` is written to a mapping, if `v==0` then 1 is written. 1 is not a valid leaf value.

Fields with top bit at 1 or leafs equal to 1 are "dirty". 

The type system prevents dealing with dirty values.

**There are additional gas costs in some cases; is it worth it anyway?**

# Some gas costs increase

_(Example with Fields but a similar scenario occurs with Leaves)_

The top bit of `Field` values in mappings (level0, level 1) is set to 1 whenever 

Some gas tests are showing extra costs of ~40k gas (classic/200 runs). This increase is due to the following scenario:

Step 1: A never-before-used level0/1 equal to 0 is cached in `local` when it receives an offer (the current best)
Step 2: That level0 or level1 becomes empty and is evicted from `local` and flushed to storage

This happens when a level0/11 become empty during a market order or a clean.
This includes the case of the entire book becoming empty and the last level0 and level1 are flushed (as empty levels) to storage.

Without this PR, the scenario above would be much cheaper. But by definition it can occur at most one per level0/1. 

Without this PR, however, the following scenario can repeat indefinitely:

Step 1: An _empty_ level0/1 equal to 0 is cached in `local` when it receives an offer `o1` (the current best)
Step 2: Another level0/1 receives an even better offer `o2`. The level0/1 mentioned in Step 1. is flushed to storage (~20k gas)
Step 3: A market order consumes both `o1` and `o2`. The level0/1 mentioned in Step 1 is _empty_ and written to storage.

### DIFF
--- a/lib/TickLib.sol
+++ b/lib/TickLib.sol
@@ -409,7 +409,7 @@ library FieldLib {
   // Will throw if field is empty
   function lastOnePosition(Field field) internal pure returns (uint) {
     // FIXME stop checking for 0 or integrate it into ctz function in assembly
-    require(!field.isEmpty() ,"field is 0");
+    require(!field.isEmpty(), "field is 0");
     return BitLib.fls(Field.unwrap(field));
   }
 
--- a/lib/TickLib.sol
+++ b/lib/TickLib.sol
@@ -292,6 +292,9 @@ library TickLib {
 
 // We use TOPBIT as the optimized in-storage value since 1 is a valid Field value
 library DirtyFieldLib {
+  DirtyField constant DIRTY_EMPTY = DirtyField.wrap(TOPBIT);
+  DirtyField constant CLEAN_EMPTY = DirtyField.wrap(0);
+
   // Return clean field with topbit set to 0
   function clean(DirtyField field) internal pure returns (Field) {
     unchecked {
--- a/lib/TickLib.sol
+++ b/lib/TickLib.sol
@@ -6,26 +6,68 @@ import {BitLib} from "mgv_lib/BitLib.sol";
 import {console2 as csf} from "forge-std/console2.sol";
 
 type Leaf is uint;
+type DirtyLeaf is uint;
 
 using LeafLib for Leaf global;
+using DirtyLeafLib for DirtyLeaf global;
 
 type Field is uint;
+type DirtyField is uint;
 
 using FieldLib for Field global;
+using DirtyFieldLib for DirtyField global;
 
 type Tick is int;
 
 using TickLib for Tick global;
 
+// Leafs are of the ford [id,id][id,id][id,id][id,id]
+// With the property that within a [id1,id2] pair, id1==0 iff id2==0
+// So 1 is not a valid leaf
+// We use that (for storage gas savings) yet the field is still considered empty
+// 1 is chosen as a special invalid leaf value to make dirty/clean more gas efficient
+library DirtyLeafLib {
+  // Return 0 if leaf is 1, leaf otherwise
+  function clean(DirtyLeaf leaf) internal pure returns (Leaf) {
+    unchecked {
+      assembly ("memory-safe") {
+        leaf := xor(eq(leaf,ONE),leaf)
+      }
+      return Leaf.wrap(DirtyLeaf.unwrap(leaf));
+    }
+  }
+  function isDirty(DirtyLeaf leaf) internal pure returns (bool) {
+    unchecked {
+      return DirtyLeaf.unwrap(leaf) == ONE;
+    }
+  }
+  function eq(DirtyLeaf leaf1, DirtyLeaf leaf2) internal pure returns (bool) {
+    unchecked {
+      return DirtyLeaf.unwrap(leaf1) == DirtyLeaf.unwrap(leaf2);
+    }
+  }
+}
+
 library LeafLib {
   Leaf constant EMPTY = Leaf.wrap(uint(0));
+
+  // Return 1 if leaf is 0, leaf otherwise
+  function dirty(Leaf leaf) internal pure returns (DirtyLeaf) {
+    unchecked {
+      assembly ("memory-safe") {
+        leaf := or(iszero(leaf),leaf)
+      }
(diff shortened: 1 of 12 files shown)


lnist:
This gas test and its sibling for new offer: "Worst case scenario if strat updates an offer on a different offer list which has become empty." are the ones I expected savings for since they now update non-zero structures.
I had not anticipated the 40k extra in some cases. Could we with little complexity avoid writing it if never written - i.e. when it becomes empty, we first read to check for 0 and then only write if not 0? But - are you saying that the extra cost is for the market order and not insert/updates? Looking at gas-tests it looks like it - and then I think it is OK. 40k is a lot, but not too much compared to total cost for a market order, but it would be a large increase for inserts/updates - and actually insert/updates are what become cheaper here.
There is an outlier: ExternalUpdateOfferOtherOfferList_Gasreq but that one is just more expensive due to updating the only offer away from "best" and doing it outside a market order. Compared, then PosthookSuccessUpdateOfferSameList does not have the extra cost, and I anticipate that is because the market order takes care of the extra writes before the posthook.
This does mean an increase in gasbase.

lnist (commented):
I wrote a lengthy comment - besides that I have not looked at it in detail. I do wonder if all this dirty-ness is needed - I had envisioned simply ignoring that single bit and whether a level is 0 or 1. Leafs would still have to be handled specially though, I see that now.

espendk:
I'm a bit surprised at the complexity here. 

For `Field`, I had expected the top bit to always be one (also in `Field.EMPTY`) such that methods like `Field.eq()` and `Field.isEmpty()` would work as is. When reading a field from a mapping or `local`, we set top bit, so that we always work with Fields that have top-bit set.

I realize that when finding the next non-zero position we would have to ignore the top-bit. Not sure how difficult/expensive that would be.


For `Leaf`, I see the problem is different. Amending reading methods to handle the special `EMPTY` value seems doable and not that costly? And perhaps a `LeafLib.encodeIfEmpty(Leaf l)` method called whenever we write to `offerList.leafs` could ensure we never write an empty leaf?

adhusson:
## Fields

My reasoning re: Fields are complicated, either we 1) always add a dirty bit to fields extracted from local; 2) or we add a disjunction to the isEmpty test (I think that’s enough); or 3) we strictly separate clean/dirty. 

(Note on the FieldLib functions: they should all work as is with dirty fields, with the proviso you noted that anyway ctz should never be called on an « empty » field)

1) Seemed more complex than 3) but I did not try. 2) meant external users would sometimes get dirty sometimes non dirty values and I did not like it but maybe it’s a small price to pay!

## Leafs

@espendk i don’t think I get your proposal it looks more complicated at first glance but I may be missing something!

## Extra gas cost

Reading before writing would mean an extra gas cost of ~100 every single time you write a level. I thought over the long run it would be less gas costly to pay the 20k once per level, but I had not thought about the sum of extra costs created by the increased gasbase… so not sure

espendk:
> Reading before writing

Was this in response to my suggestion on Slack to not write levels and leafs when they're emptied?

We'd only need to read extra levels for `{new,update}Offer` where we don't navigate through the tree from a known up-to-date state. So I think the overhead is fairly limited,

adhusson:
Ahhh no it was in response to @lnist's

> when it becomes empty, we first read to check for 0 and then only write if not 0?

If I understand your suggestion it is to leave "stale data" at a level but then every time you read a nonempty level you go up the tree to check that it actually is nonempty? At first glance that seems like quite a lot of code complexity + gas use to defend against something that happens at most once per level?

espendk:
> If I understand your suggestion it is to leave "stale data" at a level but then every time you read a nonempty level you go up the tree to check that it actually is nonempty?

yes

> At first glance that seems like quite a lot of code complexity + gas use to defend against something that happens at most once per level?

It would only be when we read from mappings and the check is trivial, so I think the code and complexity is manageable and very localized.

But is a trade-off:
- Pros:
  - avoiding an SSTORE for market orders and retracts whenever a leaf or level becomes empty
- Cons:
  - two extra mapping reads in `{new,update}Offer`

adhusson:
I've been working on implementing "only write dirty_empty field when destination is not clean_empty", I like the gas diff it yields :) (see last 2 commits)

adhusson:
I took at look at doing Field&Leaf differently:

- For Leaf, adding if statements (to detect the 'dirty empty leaf') in every Leaf-related function would be worse both gas-wise and code-complexity-wise. The current version has a very low gas overhead and gets safety from the typing which is nice.

- For Fields: I tried allowing mixed fields in #534. The code is easier to read vs. we lose the type safety giving us a guarantee that we never zero out storage slots. But also very weirdly: gas costs _increase_, often by a few hundred, up to 1.3k. I don't get why.

lib/TickLib.sol
```diff
@@ -321,10 +405,10 @@ library FieldLib {
     }
   }
 
-  // Will throw with "field is 0" if field is empty
+  // Will throw if field is empty
   function lastOnePosition(Field field) internal pure returns (uint) {
     // FIXME stop checking for 0 or integrate it into ctz function in assembly
-    require(!field.isEmpty(),"field is 0");
+    require(!field.isEmpty() ,"field is 0");
```
espendk:
```suggestion
    require(!field.isEmpty(), "field is 0");
```

PR #531, opened 2023-09-11
perf: Use vars already on stack

espendk:
Saves 13 gas in some cases 😎

Edit: Found one more that save 9 gas per offer taken in a market order 😎


adhusson (approved):
😎

PR #532, opened 2023-09-11
POC: tick shift

espendk:
## Summary
This is a POC of expanding the `logPrice` range that can be mapped into the tick tree without expanding the tree: By shifting the location of ticks in the tick tree by a per-offer list constant (`tickShift`) we shift the `logPrice` range for that offer list.

Gas measurement diff: [gas-measurement-diff.csv](https://github.com/mangrovedao/mangrove-core/files/12586145/gas-measurement-diff.csv)

### Todo

- [ ] Address FIXMEs

## 👍 Pros

- Simple code

- Choice of optimal mid-price tick
	- Besides extending, or rather *shifting*, the logPrice range, tick shifts allows placing the mid price of stable markets at a location in the tick tree where price movements will be unlikely to move outside of `level0` and `level1` thereby reducing gas costs for both makers and takers.

- Eases smaller "base tick size" than 1 bps
	- If the "base tick size" were to be made smaller than 1bps (eg 0.1bps), the combination of `tickScale` and `tickShift` should allow the existing tick tree to be used for both stable markets with small tick sizes (due to a small base tick size), volatile markets with large tick sizes (due to `tickScale`), and for tokens with a 0 to 18 difference in the number of decimals (due to `tickShift`).

## 👎 Cons

- `tickShift` is a purely technical parameter that leaks into the API. One could argue that this is similar to the `tickScale` parameter, but tick scale isn't purely technical: One could imagine having parallel markets with different tick sizes and `tickScale` enables this.

# How it works
Offer list keys are extended with a `tickShift` parameter which shifts the location of ticks in the tick tree:

```
tick        = logPrice / tickScale
shiftedTick = tick - tickShift = logPrice / tickScale - tickShift
logPrice    = tick * tickScale = (shiftedTick + tickShift) * tickScale
```

This in turn shifts the ranges of ticks and logPrices that fit in the tick tree as follows:

```
Tick tree (holds shifted ticks):
	MIN_SHIFTED_TICK = -524,288
	MAX_SHIFTED_TICK = -MIN_TICK - 1

Unshifted ticks:
	MIN_TICK = -524,288 + tickShift
	MAX_TICK = -MIN_TICK - 1 + tickShift

Log prices:
	MIN_LOG_PRICE = (-524,288 + tickShift) * tickScale
	MAX_LOG_PRICE = (-MIN_TICK - 1 + tickShift) * tickScale
```

## Range of `tickShift`

The logPrice range supported by `LogPriceLib` is restricted to
```
MIN_LOG_PRICE = -((1 << 20)-1) = -1,048,575
MAX_LOG_PRICE = -MIN_LOG_PRICE =  1,048,575;
```

For `tickScale = 1` this means tick shifts should be within this range to avoid having ticks in the tree that correspond to unsupported logPrices:

```
MIN_LOG_PRICE  =       (-524,288 + MIN_TICK_SHIFT) * tickScale  =  -1,048,575
     <=>                       (-524,288 + MIN_TICK_SHIFT) * 1  =  -1,048,575
     <=>                                        MIN_TICK_SHIFT  =  -1,048,575 + 524,288  =  -524,287

MAX_LOG_PRICE  =  (-MIN_TICK - 1 + MAX_TICK_SHIFT) * tickScale  =   1,048,575
     <=>                        (524,287 + MAX_TICK_SHIFT) * 1  =   1,048,575
     <=>                                        MAX_TICK_SHIFT  =   1,048,575 - 524,287  =   524,288
```

Note that for `tickScale > 1`, the supported `tickShift` range becomes smaller IF all ticks in the tick tree should correspond to a supported logPrice.

However, even without `tickShift`, `tickScale > 1` means not all ticks in the tick tree correspond to a supported logPrice. Already at `tickScale = 2`, `MIN_TICK` gives a logPrice that is outside the supported range:

```
logPrice(MIN_TICK)  =  MIN_TICK * tickScale  =  -524,288 * 2  =  -1,048,576  
                                                              <  -1,048,575  =  MIN_LOG_PRICE
```

# Notes

## Terminology
I think the terminology we use around ticks is a bit confusing and possibly ambiguous. A glossary would be useful and I think we should try to align our terminology with common definitions (to the extent possible); Perhaps we can check with Hamza.

Some thoughts on terminology:

- "tick": *a minimal movement from a given price*
	- in some sources, *tick* appears to be a synonym for *tick size*
	- in others, a *tick* sounds like it's a minimal movement from a given price
		- https://www.investopedia.com/ask/answers/032615/what-difference-between-pips-points-and-ticks.asp
	- in Mangrove, a *tick* is currently an index into the lowest level of the tick tree and corresponds to a certain `logPrice`.
		- perhaps a more apt name could be "tickIndex"?

- "tick size": *the minimum, relative price movement in a market*
	- adapted from https://www.investopedia.com/terms/t/tick-size.asp
	- in Mangrove, this is currently $1.0001^{tickScale} - 1$ for a given `tickScale > 0`.

- "tick scale": ?
	- I couldn't find clearly relevant definitions of this. Google mostly gave results related to charting ticks while ChatGPT said "tick scale" and "tick size" are mostly synonymous.
	- in Mangrove, it is the number of "base ticks"/"basis point ticks" between two ticks
		- perhaps a more apt name could be "baseTicksPerTick"?

- "tick shift": ?
	- This is a purely technical term
	- in Mangrove, "tick shift" describes a shift of the *tick (index) range*
		- perhaps a more apt name could be "tickIndexShift"

### DIFF
--- a/lib/LogPriceLib.sol
+++ b/lib/LogPriceLib.sol
@@ -12,8 +12,8 @@ library LogPriceLib {
   }
   // 
   // FIXME ensure that you only called tick*tickScale if you previously stored the tick as a result of logPrice/tickScale. Otherwise you may go beyond the MAX/MIN.
-  function fromTick(Tick tick, uint tickScale) internal pure returns (int) {
-    return Tick.unwrap(tick) * int(tickScale);
+  function fromTick(Tick tick, uint tickScale, int tickShift) internal pure returns (int) {
+    return (Tick.unwrap(tick) + tickShift) * int(tickScale);
   }
 
   // tick underestimates the price, so we underestimate  inbound here, i.e. the inbound/outbound price will again be underestimated
--- a/lib/TickLib.sol
+++ b/lib/TickLib.sol
@@ -137,10 +137,10 @@ library TickLib {
     return Tick.unwrap(tick) >= MIN_TICK && Tick.unwrap(tick) <= MAX_TICK;
   }
 
-  function fromLogPrice(int logPrice, uint tickScale) internal pure returns (Tick) {
+  function fromLogPrice(int logPrice, uint tickScale, int tickShift) internal pure returns (Tick) {
     // Do not force logPrices to fit the tickScale (aka logPrice%tickScale==0)
     // Round all prices down (aka cheaper for taker)
-    int tick = logPrice / int(tickScale);
+    int tick = logPrice / int(tickScale) - tickShift;
     if (logPrice < 0 && logPrice % int(tickScale) != 0) {
       tick = tick - 1;
     }
--- a/lib/preprocessed/ToString.post.sol
+++ b/lib/preprocessed/ToString.post.sol
@@ -62,10 +62,10 @@ function tickBranchToString(Tick tick) pure returns (string memory) {
 
 function toString(Tick tick) pure returns (string memory ret) {
   string memory suffix;
-  if (MIN_TICK > Tick.unwrap(tick) || Tick.unwrap(tick) > MAX_TICK) {
-    suffix = "out of range";
+  if (tick.inRange()) {
+    suffix = logPriceToString(LogPriceLib.fromTick(tick,1,0));
   } else {
-    suffix = logPriceToString(LogPriceLib.fromTick(tick,1));
+    suffix = "out of range";
   }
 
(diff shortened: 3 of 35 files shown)


espendk:
Closing this as we decided to add third level instead: https://github.com/mangrovedao/mangrove-core/pull/529

PR #534, opened 2023-09-12
No zeroing data structures: allow mixed fields

adhusson:


adhusson:
closing due to slight gas use increase and loss of type-provided guarantee that written fields will always be dirty

PR #535, opened 2023-09-13
dont check offer.wants, add a test for safety

adhusson:


src/MgvOfferTaking.sol
```diff
@@ -516,8 +516,13 @@ abstract contract MgvOfferTaking is MgvHasOffers {
         }
 
         /* We update the totals in the multiorder based on the adjusted `sor.wants`/`sor.gives`. */
-        /* overflow: sor.{wants,gives} are on 96bits, sor.total{Got,Gave} are on 256 bits. */
+        /* no overflow: sor.wants is on 96 bits */
```
lnist: how is sor.wants 96 bits? If fillWants is false, and an offer wants a high amount (say close to close to 256 bits), then in `execute` we have `sor.wants = LogPriceLib.outboundFromInbound(sor.offer.logPrice(), fillVolume);` - fillVolume is bounded by MAX_SAFE_VOLUME which is >>96 bits. What am I missing?

lnist: Just below you say sor.wants can be 248 bits? :)

test/core/TakerOperations.t.sol
```diff
@@ -1152,6 +1152,21 @@ contract TakerOperationsTest is MangroveTest {
...
     assertTrue(emptyLeaf.isEmpty(), "leaf should not have other tick used");
   }
+
+  /* 
+  An attempt to check for overflow when accumulating sor.gives into mor.totalGave.
+  I have not found a way to actually trigger it by mutating state somewhere.
+  This test just considers as many offers as possible that each have a maximal `wants` and makes sure the error will be about stack overflow, not uint overflow. 
+  */
+  function test_maximal_wants_is_ok() public {
+    uint maxOfferWants = LogPriceLib.inboundFromOutboundUp(MAX_LOG_PRICE, type(uint96).max);
```
lnist, line 1162: isn't this only covering the fillWants=true scenario? which actually sets sor.gives and not sor.wants?

lnist: Ah, when fillWants=false then fillVolume will be the max we give, and it can obviously not be above uint256.max?

src/MgvOfferTaking.sol
adhusson: Thanks fixed in 0aba68666b1773bcb14343cf2dd5af88936ebd96 (refixed in bc4f34adaa66daf5d3b2219b4452081054311cdd)! (I meant sor.gives can be 248 bits, and actually sor.wants can be 104 bits)

sor.wants is 96 bits most of the time because the line
```
sor.wants = LogPriceLib.outboundFromInbound(sor.offer.logPrice(), fillVolume)
```

basically means `taker.wants = fillVolume/offer.price`, and we know that point that `fillVolume < offer.wants`. Since `offer.wants = offer.gives * offer.price`, we have `taker.wants <= offer.gives` and `offer.gives` is on 96 bits.

As I mention in the comment below, it can be 104 bits once.

test/core/TakerOperations.t.sol
adhusson: > isn't this only covering the fillWants=true scenario? which actually sets sor.gives and not sor.wants?

In both cases (fillWants = true or not), the 2nd line here:
https://github.com/mangrovedao/mangrove-core/blob/0aba68666b1773bcb14343cf2dd5af88936ebd96/src/MgvOfferTaking.sol#L471-L472
can cause the overflow.

In addition, when `fillWants=false` this can also set `sor.wants` to a number that is too close to uint.max:
https://github.com/mangrovedao/mangrove-core/blob/0aba68666b1773bcb14343cf2dd5af88936ebd96/src/MgvOfferTaking.sol#L479

But I think both cases are covered by multiplying `maxOfferWants` with `maxRecursionDepth`.

> Ah, when fillWants=false then fillVolume will be the max we give, and it can obviously not be above uint256.max?
Ahh OK so `fillVolume` is currently limited to 104 bits so when `fillWants=true` the general case is you set `sor.wants` to a <= 96 bits number N times then you set it to a <= 104 bits number one time. So I'll update my comment again, it's 104 not 96 :)

lnist (approved):

PR #543, opened 2023-09-14
Fix/max safe volume and roundup

adhusson:
- MAX_SAFE_VOLUME was off by x2
- inboundFromOutboundUp and outboundFromInboundUp now correctly round up when exp of floating-point price is > 255


lib/BitLib.sol
```diff
@@ -2,21 +2,33 @@
...
+        // isolate lsb
+        x := and(x, add(not(x), 1))
+
+        // 6th bit
+        c := or(c,shl(5, lt(0xffffffff, x)))
+
+        // debruijn lookup
+        c := or(c, byte(shr(251, mul(shr(c, x), shl(224, 0x077cb531))), 
+            0x00011c021d0e18031e16140f191104081f1b0d17151310071a0c12060b050a09))
+      }
```
espendk, line 27: I haven't checked this, let me know if I should.
Is it an optimized version of the previous implementation for 64 bits?

lib/LogPriceLib.sol
```diff
@@ -52,11 +53,28 @@ library LogPriceLib {
...
+}
 
+// Return a/2**e rounded up
+function divExpUp(uint a, uint e) pure returns (uint) {
+  unchecked {
+    uint carry;
+    /* 
+    Would like to write
 
+      carry = (e<<1 == 0 || prod % (e<<1) != 0) ? 1 : 0
```
espendk: `prod` -> `a`
```suggestion
      carry = (e<<1 == 0 || a % (e<<1) != 0) ? 1 : 0
```

espendk: Why left-shift when checking for zero?
Can't we just do:

```
carry = (e == 0 || a % (e<<1) != 0) ? 1 : 0
```

espendk: In the event that `e == 0  =>  2**e == 1` shouldn't `carry == 1` iff `a % 2 == 1`?

I think the current formula adds 1 when `e == 0`  and  `a % 2 == 0`, but I don't think it should?

espendk: Oh, should the expression `e<<1`have been `1<<e`?
That would correspond to $2^e$ which would make sense here :-)

(the fact that Yul's `shl(a,b)` corresponds to `b<<a` is confusing)

lib/TickLib.sol
```diff
@@ -62,110 +63,109 @@ library LeafLib {
...
-        } else {
-          offerId = Leaf.unwrap(leaf) << (OFFER_BITS * 6) >> (OFFER_BITS * 7);
-          if (offerId != 0) {
-            ret = 3;
-          }
-        }
+    unchecked {
+      assembly("memory-safe") {
+        ret := shl(1,gt(0xffffffffffffffffffffffffffffffff,leaf))
+        ret := or(ret,gt(0xffffffffffffffff,shr(shl(7,iszero(ret)),leaf)))
```
espendk, line 161: I have not checked this. Let me know if I should

lib/TickLib.sol
```diff
@@ -174,33 +174,49 @@ library LeafLib {
...
-          (BitLib.ctz(Field.unwrap(level1)) |
-            (BitLib.ctz(Field.unwrap(level2)) |
-              uint((int(BitLib.ctz(Field.unwrap(level3)))-LEVEL3_SIZE/2) << LEVEL2_SIZE_BITS)) 
+      LocalPacked local;
+      local = local.tickPosInLeaf(tickPosInLeaf).level0(level0).level1(level1).level2(level2).level3(level3);
+      return tickFromLocal(local);
+    }
+  }
+
+  function tickFromLocal(LocalPacked local) internal pure returns (Tick) {
```
espendk: These return the *best* tick on that branch, right? If so, would be good if the name reflected this, eg, `bestTickFromBranch`.

test/core/Leaf.t.sol
```diff
@@ -286,12 +293,18 @@ contract TickTest is Test {
...
     Field level2 = Field.wrap(bound(_level2, 1, uint(LEVEL2_SIZE) - 1));
     Field level3 = Field.wrap(bound(_level3, 1, uint(LEVEL3_SIZE) - 1));
-    Tick tick = TickLib.tickFromBranch(tickPosInLeaf, level0, level1, level2, level3);
+    MgvStructs.LocalPacked local;
+    local = local.tickPosInLeaf(tickPosInLeaf);
+    local = local.level0(level0);
+    local = local.level1(level1);
+    local = local.level2(level2);
+    local = local.level3(level3);
+    Tick tick = TickLib.tickFromLocal(local);
```
espendk, line 302: Test name still refers to `ticksFromBranch`: `test_tickFromBranch_matches_positions_accessor`

test/core/Leaf.t.sol
```diff
@@ -351,7 +364,11 @@ contract FieldTest is Test {
     for (; i < 256; i++) {
       if (uint(b >> i) % 2 == 1) break;
     }
-    assertFirstOnePosition(uint(b), i);
+    assertFirstOnePosition(uint(b), i > MAX_LEVEL_SIZE ? MAX_LEVEL_SIZE : i);
+  }
+
+  function test_regress() public {
+    test_firstOnePosition_auto(14530771982722103506997891882992560810543470234613570268015259683728373317632);
```
espendk, line 371: Should this stay here?

lib/LogPriceLib.sol
```diff
@@ -52,11 +53,28 @@ library LogPriceLib {
...
 
+    but solidity reverts on a%0 even inside unchecked blocks
+    */
+    assembly("memory-safe") {
+      let den := shl(e,1)
+      carry := gt(or(mod(a,den),iszero(den)),0)
```
espendk: When can `iszero(den) == iszero(shl(e,1))` be true? When `e > 255`?

espendk: If so, shouldn't there only be a carry if `a > 0`?

test/core/Leaf.t.sol
```diff
@@ -463,4 +480,53 @@ contract FieldTest is Test {
...
+    //inboundFromOutboundUp
+    (sig, exp) = LogPriceConversionLib.nonNormalizedPriceFromLogPrice(logPrice);
+    assertEq(LogPriceLib.inboundFromOutboundUp(logPrice, amt), divExpUp_spec(sig * amt, exp));
+
+    //outboundFromInboundUp
+    (sig, exp) = LogPriceConversionLib.nonNormalizedPriceFromLogPrice(-logPrice);
+    assertEq(LogPriceLib.outboundFromInboundUp(logPrice, amt), divExpUp_spec(sig * amt, exp));
+  }
+
+  function test_outuboundFromInboundUp(int logPrice, uint amt) public {
```
espendk:
```suggestion
  function test_outboundFromInboundUp(int logPrice, uint amt) public {
```

test/core/Leaf.t.sol
```diff
@@ -463,4 +480,53 @@ contract FieldTest is Test {
...
+    amt = bound(amt, 0, MAX_SAFE_VOLUME);
+    logPrice = bound(logPrice, MIN_LOG_PRICE, MAX_LOG_PRICE);
+
+    (uint sig, uint exp) = LogPriceConversionLib.nonNormalizedPriceFromLogPrice(logPrice);
+    assertEq(LogPriceLib.inboundFromOutboundUp(logPrice, amt), divExpUp_spec(sig * amt, exp));
```
espendk: Test name indicates this should test `outboundFromInboundUp`?

test/core/Leaf.t.sol
espendk: Does the previous test cover the same?

test/core/gas/MarketOrderOtherOfferList.t.sol
```diff
@@ -234,6 +234,8 @@ contract ExternalMarketOrderOtherOfferList_WithMultipleOffersAtManyTicks is Tick
     mgv.marketOrderByLogPrice(_olKey, _logPrice, 2 ** 96, false);
     gas_();
     (, MgvStructs.LocalPacked local) = mgv.config(_olKey);
-    assertLt(_logPrice, LogPriceLib.fromTick(local.bestTick(), _olKey.tickScale));
+    if (!local.level3().isEmpty()) {
```
espendk, line 237: How come this is necessary now? Does it mean some assumption in the tests no longer holds?

espendk (commented):
Gnarly stuff 😅 And more than expected for this fix. I added some comments and questions.

I did not review the optimizations, let me know if I should. Would appreciate an intro in that case :-)

lib/BitLib.sol
adhusson: yes it's an optimized version -- we can check this later :)

lib/TickLib.sol
adhusson: see #544

lib/TickLib.sol
adhusson: see #544

test/core/Leaf.t.sol
adhusson: see #544

test/core/Leaf.t.sol
adhusson: see #544

test/core/gas/MarketOrderOtherOfferList.t.sol
adhusson: see #544

lib/LogPriceLib.sol
adhusson: sorry this was stale, i'm pushing a cleaner version that addresses your comments

test/core/Leaf.t.sol
adhusson: test removed

test/core/Leaf.t.sol
adhusson: test removed

lib/LogPriceLib.sol
adhusson: pushed a fix :)

adhusson:
closed for #553

test/core/gas/MarketOrderOtherOfferList.t.sol
espendk: #554 I believe

PR #551, opened 2023-09-16
feat: Post offers at nearest, higher tick for unaligned logPrices

espendk:
The maker API admits (log)Prices that do not correspond exactly to a tick. Frontends and other intermediaries should ensure offers aren't posted at such prices, but we cannot ensure this. So either the maker API should revert if given such an unaligned logPrice or the offer should be posted at either the nearest higher or lower tick.

We think that some integrations will be easier if the API is permissive and allows unaligned logPrices, so we opt for choosing one of the nearest ticks.

Previously, the nearest, lower tick was chosen when adding the offer to the book. This meant that makers would sell at a lower price than what they specified which seems wrong.

This PR therefore changes the maker API such that offers are posted at the nearest, higher tick when given an unaligned logPrice.

Also, this PR optimizes the conversion of offer logPrices to ticks, exploiting that offers are always stored with a tick-aligned logPrice.

### DIFF
--- a/lib/TickLib.sol
+++ b/lib/TickLib.sol
@@ -181,15 +181,21 @@ library TickLib {
     return Tick.unwrap(tick) >= MIN_TICK && Tick.unwrap(tick) <= MAX_TICK;
   }
 
-  // Returns the closest, lower tick to the given logPrice at the given tickScale
-  function closestLowerTickToLogPrice(int logPrice, uint tickScale) internal pure returns (Tick) {
+  // Returns the nearest, higher tick to the given logPrice at the given tickScale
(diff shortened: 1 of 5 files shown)


adhusson (approved):
Looks good, just wondering should we have tests for the aligned version to either check the (trivial) implementation + that offers are always indeed aligned?

espendk:
> just wondering should we have tests for the aligned version to either check the (trivial) implementation

Good point, addressed in a3c2c24d7b326bdd0f65a1cf5e8b059e67c7ed09.

> + that offers are always indeed aligned?

The existing `test_insertionLogPrice_normalization` already checks this :-)

PR #553, opened 2023-09-17
Fix MAX_SAFE_VOLUME and roundup

adhusson:
proper PR for fixing roundup + max_safe_volume

### DIFF
--- a/lib/LogPriceLib.sol
+++ b/lib/LogPriceLib.sol
@@ -27,9 +27,10 @@ library LogPriceLib {
   // no overflow if outboundAmt is on 104 bits
   // rounds up
   function inboundFromOutboundUp(int logPrice, uint outboundAmt) internal pure returns (uint) {
-    (uint sig, uint exp) = LogPriceConversionLib.nonNormalizedPriceFromLogPrice(logPrice);
-    uint prod = sig*outboundAmt;
-    return (prod>>exp) + (prod%(1<<exp)==0 ? 0 : 1);
+    unchecked {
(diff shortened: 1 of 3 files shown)


lib/LogPriceLib.sol
```diff
@@ -52,11 +53,24 @@ library LogPriceLib {
...
+function divExpUp(uint a, uint e) pure returns (uint) {
+  unchecked {
+    uint carry;
+    /* 
+    Easier to represent carry in assembly.
+    carry is 1 if a&mask > 0, where mask is (1<<e)-1, and carry is 0 otherwise
+    */
+    assembly("memory-safe") {
+      carry := gt(and(a,sub(shl(e,1),1)),0)
+    }
```
espendk: I'm convinced, but I needed the explanation you gave me on Slack, so maybe include it here?

Copying it here for convenience :-)

> if a == 0then a & x == 0 for all x, so carry is 0
> otherwise, if e > 255 then (1<<e)-1 == type(uint).max so a & ((1<<e)-1) == a > 0, so carry is 1 (as it should, since 0 < a < 2^e)
> otherwise we’re in the normal case since a & ((1<<e)-1) == a % (1<<e)

espendk (approved):
Great! I think the reasoning for why it works deserves to be retained :-)

lib/LogPriceLib.sol
adhusson: done in https://github.com/mangrovedao/mangrove-core/pull/553/commits/32333eaabfaae5a7f65ed6867aec611f88701c41

PR #554, opened 2023-09-17
optimize ctz

adhusson:
- uses less gas
- specialized for 64bit uints
- easier to audit
- also renames `tickFromX` to `bestTickFromX`

In `ExternalMarketOrderOtherOfferList_WithMultipleOffersAtManyTicks`, some tests would leave the OB empty, others not. `tickFromBranch` and `tickFromLocal` must never be called when one of the levels is empty -- the result is garbage -- but that's what `ExternalMarketOrderOtherOfferList_WithMultipleOffersAtManyTicks` did. Now it checks for empty OB before reading the tick from `local`.

### DIFF
--- a/lib/BitLib.sol
+++ b/lib/BitLib.sol
@@ -2,21 +2,33 @@
 pragma solidity ^0.8.17;
 
 library BitLib {
-  // Returns the number of zeroes in x that do not have a 1 to their right, eg ctz(0)=256, ctz(1)=0
-  function ctz(uint x) internal pure returns (uint c) {
-    assembly ("memory-safe") {
-      // Isolate the least significant bit
-      x := and(x, add(not(x), 1))
+  // - if x is a nonzero uint64: 
+  //   return number of zeroes in x that do not have a 1 to their right
+  // - otherwise:
+  //    return 64
+  function ctz64(uint x) internal pure returns (uint256 c) {
+    unchecked {
+      assembly ("memory-safe") {
+        // clean
+        x:= and(x,0xffffffffffffffff)
 
-      // Get first 3 bits of c, this is the unusual part
-      c := shl(5,shr(252,shl(shl(2,shr(250,mul(x, 0xb6db6db6ddddddddd34d34d349249249210842108c6318c639ce739cffffffff))),0x8040405543005266443200005020610674053026020000107506200176117077)))
+        // 7th bit
+        c:= shl(6,iszero(x))
 
-      // Get last 5 bits of c
-      c := or(c, byte(shr(251, mul(shr(c, x), shl(224, 0x077cb531))), 
-          0x00011c021d0e18031e16140f191104081f1b0d17151310071a0c12060b050a09))
+        // isolate lsb
+        x := and(x, add(not(x), 1))
+
+        // 6th bit
+        c := or(c,shl(5, lt(0xffffffff, x)))
+
+        // debruijn lookup
+        c := or(c, byte(shr(251, mul(shr(c, x), shl(224, 0x077cb531))), 
+            0x00011c021d0e18031e16140f191104081f1b0d17151310071a0c12060b050a09))
+      }
     }
   }
 
+  // The fls function below is general-purpose and used by tests but not by Mangrove itself
   // Function fls is MIT License. Copyright (c) 2022 Solady.
 /// @dev find last set.
     /// Returns the index of the most significant bit of `x`,
--- a/lib/TickLib.sol
+++ b/lib/TickLib.sol
@@ -4,6 +4,7 @@ pragma solidity ^0.8.17;
 import "mgv_lib/Constants.sol";
 import {BitLib} from "mgv_lib/BitLib.sol";
 import {console2 as csf} from "forge-std/console2.sol";
+import {LocalPacked} from "mgv_src/preprocessed/MgvLocal.post.sol";
 
 type Leaf is uint;
 type DirtyLeaf is uint;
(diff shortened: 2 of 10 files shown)


lib/BitLib.sol
```diff
@@ -2,21 +2,33 @@
...
+        // isolate lsb
+        x := and(x, add(not(x), 1))
+
+        // 6th bit
+        c := or(c,shl(5, lt(0xffffffff, x)))
+
+        // debruijn lookup
+        c := or(c, byte(shr(251, mul(shr(c, x), shl(224, 0x077cb531))), 
+            0x00011c021d0e18031e16140f191104081f1b0d17151310071a0c12060b050a09))
+      }
```
espendk, line 27: I haven't checked this, let me know if I should.

lib/TickLib.sol
```diff
@@ -128,9 +129,8 @@ library LeafLib {
     return uint(raw << (OFFER_BITS * ((index * 2) + 1)) >> (256 - OFFER_BITS));
   }
 
-  // TODO optimize with a hashmap, or:
-  // if > half: +=1; if += quarter: +=1, or:
-  // or nested if dichotomy
+  // Will check for the first position (0,1,2 or 3) that has a nonzero first-of-tick or a nonzero last-of-tick offer. Leafs where only one of those is nonzero are invalid anyway.
+  // Offers are ordered msb to lsb
```
espendk, line 133: This is unintuitive, since it's the opposite for levels 😅

test/core/Leaf.t.sol
```diff
@@ -286,12 +286,18 @@ contract TickTest is Test {
...
     Field level2 = Field.wrap(bound(_level2, 1, uint(LEVEL2_SIZE) - 1));
     Field level3 = Field.wrap(bound(_level3, 1, uint(LEVEL3_SIZE) - 1));
-    Tick tick = TickLib.tickFromBranch(tickPosInLeaf, level0, level1, level2, level3);
+    MgvStructs.LocalPacked local;
+    local = local.tickPosInLeaf(tickPosInLeaf);
+    local = local.level0(level0);
+    local = local.level1(level1);
+    local = local.level2(level2);
+    local = local.level3(level3);
+    Tick tick = TickLib.bestTickFromLocal(local);
```
espendk, line 322: The test name still refers to the old `tickFromBranch` name

test/core/Leaf.t.sol
```diff
@@ -286,12 +286,18 @@ contract TickTest is Test {
     Field level1 = Field.wrap(bound(_level1, 1, uint(LEVEL1_SIZE) - 1));
     Field level2 = Field.wrap(bound(_level2, 1, uint(LEVEL2_SIZE) - 1));
     Field level3 = Field.wrap(bound(_level3, 1, uint(LEVEL3_SIZE) - 1));
-    Tick tick = TickLib.tickFromBranch(tickPosInLeaf, level0, level1, level2, level3);
+    MgvStructs.LocalPacked local;
+    local = local.tickPosInLeaf(tickPosInLeaf);
+    local = local.level0(level0);
+    local = local.level1(level1);
+    local = local.level2(level2);
+    local = local.level3(level3);
```
espendk, line 321: This is the simplest case where only one bit is set in each level. Would be good to also test this where some higher bits are also set.

test/core/Leaf.t.sol
```diff
@@ -351,7 +357,7 @@ contract FieldTest is Test {
     for (; i < 256; i++) {
       if (uint(b >> i) % 2 == 1) break;
     }
-    assertFirstOnePosition(uint(b), i);
+    assertFirstOnePosition(uint(b), i > MAX_LEVEL_SIZE ? MAX_LEVEL_SIZE : i);
```
espendk, line 387: I found this surprising. Would it perhaps be more natural to split this into two assertions? Or am I missing something?

```suggestion
    assertTrue(i < MAX_LEVEL_SIZE);
    assertFirstOnePosition(uint(b), i);
```

test/core/Leaf.t.sol
```diff
@@ -463,4 +469,12 @@ contract FieldTest is Test {
   function field_isDirty(DirtyField field) public {
     assertEq(field.isDirty(), DirtyField.unwrap(field) & TOPBIT == TOPBIT);
   }
+
+  // Some BitLib.ctz64 relies on specific level sizes
+  function test_level_sizes() public {
+    assertLe(LEVEL3_SIZE, int(MAX_LEVEL_SIZE), "bad level3 size");
+    assertLe(LEVEL2_SIZE, int(MAX_LEVEL_SIZE), "bad level2 size");
+    assertLe(LEVEL1_SIZE, int(MAX_LEVEL_SIZE), "bad level1 size");
+    assertLe(LEVEL0_SIZE, int(MAX_LEVEL_SIZE), "bad level0 size");
```
espendk: Since these are `int` should we check that they are also > 0?

test/core/gas/MarketOrderOtherOfferList.t.sol
```diff
@@ -234,6 +234,9 @@ contract ExternalMarketOrderOtherOfferList_WithMultipleOffersAtManyTicks is Tick
     mgv.marketOrderByLogPrice(_olKey, _logPrice, 2 ** 96, false);
     gas_();
     (, MgvStructs.LocalPacked local) = mgv.config(_olKey);
-    assertLt(_logPrice, LogPriceLib.fromTick(local.bestTick(), _olKey.tickScale));
+    // In some tests the market order takes all offers, in others not
```
espendk: Not clear why this is relevant, so perhaps state that `local.bestTick()` only makes sense on a non-empty book. Something like:

```suggestion
    // In some tests the market order takes all offers, in others not. `local.bestTick()` must only be called when the book is non-empty
```

espendk (commented):
Cool! Added some smaller improvement suggestions

lib/TickLib.sol
adhusson: agreed but I couldn't figure out how to easily do it the other way around

test/core/Leaf.t.sol
adhusson: fixed in adad68d184b477bc2c72959e208ac17ad791baa5

test/core/Leaf.t.sol
adhusson: Doesn't this allow any number of bits to be set except 0, up to the size of the level?
```
    Field level0 = Field.wrap(bound(_level0, 1, uint(LEVEL0_SIZE) - 1));
    Field level1 = Field.wrap(bound(_level1, 1, uint(LEVEL1_SIZE) - 1));
    Field level2 = Field.wrap(bound(_level2, 1, uint(LEVEL2_SIZE) - 1));
    Field level3 = Field.wrap(bound(_level3, 1, uint(LEVEL3_SIZE) - 1));
 ```

File: test/core/Leaf.t.sol (hunk 5 above)
Comment (adhusson) [reply test/core/Leaf.t.sol:387]: So before `ctz` would return the number of trailing zeroes for any word. Now it maxes out at 64. But the loop just before the assertions may push `i` to 255, so `assertTrue(i < MAX_LEVEL_SIZE);` would fail. Would this be OK?

```
if (i > MAX_LEVEL_SIZE) {
  assertFirstOnePosition(uint(b),MAX_LEVEL_SIZE);
} else {
  assertFirstOnePosition(uint(b), i);
}
```

test/core/Leaf.t.sol
adhusson: Fixed in d8bb32b2b4b8dd8e9aa2e0e51c9ca33d4ab9fdbe

lib/BitLib.sol
adhusson: we can schedule this for later!

test/core/Leaf.t.sol
espendk: doh, you're right, my bad - sorry!

test/core/Leaf.t.sol
espendk: I think that commit changes too much 😅 Adding `best` is IMO only needed and relevant when the arguments could identify multiple ticks; When the arguments determine one tick, then I don't think `best` is relevant and actually somewhat misleading since it indicates we choose between multiple possible ticks.

test/core/Leaf.t.sol
espendk: Ah, now I get it. I think it's fine as is 👍

test/core/Leaf.t.sol
adhusson: fixed with revert + proper commit

espendk: I'm still seeing this in TestTickTree?

`function bestTickFromBranch(uint posInLevel3, uint posInLevel2, uint posInLevel1, uint posInLevel0, uint posInLeaf)`

espendk (approved):

PR #555, opened 2023-09-17
optimize leaf.firstOfferPosition (changes spec for invalid leaves)

adhusson:
Uses simple assembly instead of nested if-then-else.


espendk (approved):
fancy bit-shuffling 😎

PR #556, opened 2023-09-17
Avoid useless JUMP

adhusson:
Avoids useless jumps (2nd branch of upper `if` implies `if` below cannot be taken)


espendk (approved):
neat!

PR #557, opened 2023-09-17
consistent unchecked (after #554 and #555)

adhusson:
Depends on #554 and #555 (ie should be merged after)


espendk (approved):
cool. Should we do the same for the other libs?

adhusson:
yes I think so after checking that its safe of course

PR #558, opened 2023-09-17
index->pos (depends on #557)

adhusson:
Depends on #557


espendk (approved):
Good catch, makes things clearer 👌

PR #559, opened 2023-09-18
remove setDensity, self-documenting setDensity fn names

adhusson:
- remove `setDensity` 
- rename setters/events to `setDensity96X32` so format is immediately known
- fix MgvOracle error message


espendk (commented):
I'm surprised that there aren't more changes in this PR. What about all the calls to `mgv.activate` in tests and scripts? Have they already been migrated to the new density format?

espendk:
I just did a few random samples:
- I see `ActivateSemibook.s.sol`  was updated when the new density format was introduced in 317310b
   - It looks like things were mostly migrated to the new format in that commit :+1:
- `MangroveTes` has `MangroveTestOptions.density` which is passed unchanged to `activate` - maybe that should be renamed to `density96x32` or converted before being passed to `activate`?

espendk (approved):
🧹🌪️👌

PR #563, opened 2023-09-18
Remove FIXMEs

adhusson:
- Did not remove 2 FIXMEs related to ticks in `GasTestBase.t.sol`, @lnist is it hard to adapt?


lib/LogPriceLib.sol
```diff
@@ -11,7 +11,6 @@ library LogPriceLib {
     return logPrice >= MIN_LOG_PRICE && logPrice <= MAX_LOG_PRICE;
   }
   //
```
espendk:
```suggestion

```

lib/TickLib.sol
```diff
@@ -483,7 +474,6 @@ library FieldLib {
 
   // Will throw if field is empty
   function firstOnePosition(Field field) internal pure returns (uint) {
-    // FIXME stop checking for 0 or integrate it into ctz function in assembly
     unchecked {
       require(!field.isEmpty(),"field is 0");
```
espendk: Can we remove the require? I'd expect that we should be able to

lib/TickLib.sol
```diff
@@ -493,7 +483,6 @@ library FieldLib {
   // Will throw if field is empty
   function lastOnePosition(Field field) internal pure returns (uint) {
     unchecked {
-      // FIXME stop checking for 0 or integrate it into ctz function in assembly
       require(!field.isEmpty(), "field is 0");
```
espendk: same here

src/MgvOfferTaking.sol
```diff
@@ -778,7 +777,7 @@ abstract contract MgvOfferTaking is MgvHasOffers {
         mor.feePaid = concreteFee;
       }
       if (mor.totalGot > 0) {
-        /* It should be statically provable that this transfer cannot return false under well-behaved ERC20s and a non-blacklisted, non-0 target, if governance does not call withdrawERC20 during order execution. */
+        /* It should be statically provable that this transfer cannot return false under well-behaved ERC20s and a non-blacklisted, non-0 target, if governance does not call withdrawERC20 during order execution, unless the caller a set a gas limit which precisely makes `transferToken` go OOG but retains enough gas to revert here. */
```
```suggestion
        /* It should be statically provable that this transfer cannot return false under well-behaved ERC20s and a non-blacklisted, non-0 target, if governance does not call withdrawERC20 during order execution, unless the caller set a gas limit which precisely makes `transferToken` go OOG but retains enough gas to revert here. */
```

test/core/Constant.t.sol
```diff
@@ -0,0 +1,45 @@
+// SPDX-License-Identifier:	AGPL-3.0
+
+// those tests should be run with -vv so correct gas estimates are shown
```
espendk: There are no gas estimates in these tests? Copy-paste leftover?

test/core/Constant.t.sol
```diff
@@ -0,0 +1,45 @@
...
+  function test_constants_min_max_price() public {
+    (uint man, uint exp) = LogPriceConversionLib.priceFromLogPrice(MIN_LOG_PRICE);
+    assertEq(man, MIN_PRICE_MANTISSA);
+    assertEq(int(exp), MIN_PRICE_EXP);
+    (man, exp) = LogPriceConversionLib.priceFromLogPrice(MAX_LOG_PRICE);
+    assertEq(man, MAX_PRICE_MANTISSA);
+    assertEq(int(exp), MAX_PRICE_EXP);
+  }
+
+  // Since "Only direct number constants and references to such constants are supported by inline assembly", NOT_TOPBIT is not defined in terms of TOPBIT. Here we check that its definition is correct.
```
espendk, line 20: ffs solcc 🤬

test/core/Leaf.t.sol
espendk, line 1: This file also has an initial comment about gas estimates - is it relevant?

test/core/MakerOperations.t.sol
```diff
@@ -1050,4 +1047,11 @@ contract MakerOperationsTest is MangroveTest, IMaker {
     (, local) = mgv.config(olKey);
     assertEq(local.tickPosInLeaf(), 2);
   }
+
+  function test_leaf_update_both_first_and_last() public {
+    uint ofr0 = mgv.newOfferByVolume(olKey, 0.01 ether, 1 ether, 1000000, 0);
```
espendk: I think we should avoid adding more ByVolume in tests. I find they are more difficult to reason about because the price (and thus tick) isn't apparent.

espendk (commented):
🧹

lib/TickLib.sol
adhusson: fixed in 962779950c4407e50d1a2c9882e0f1ff6bfb4d5

however all code that calls `[first|last]OnePosition` should be wary about this

lib/TickLib.sol
adhusson: fixed in 962779950c4407e50d1a2c9882e0f1ff6bfb4d5

however all code that calls `[first|last]OnePosition` should be wary about this

test/core/Constant.t.sol
adhusson: fixed in 962779950c4407e50d1a2c9882e0f1ff6bfb4d56

test/core/Leaf.t.sol
adhusson: fixed in 962779950c4407e50d1a2c9882e0f1ff6bfb4d56

test/core/MakerOperations.t.sol
adhusson: fixed in 962779950c4407e50d1a2c9882e0f1ff6bfb4d56

espendk (approved):

PR #565, opened 2023-09-19
Fix/remove spurious todo

adhusson:
I don't think this TODO is right, but making a PR just in case


espendk (approved):
good for me
@lnist should probably also check

PR #566, opened 2023-09-19
restore full event checking in tests now that penalties are stable

adhusson:


espendk (approved):
poor future us when we modify the code 😅

PR #567, opened 2023-09-19
fix: properly update gas test to level3

adhusson:
In #564 I wrote as the last component of the sum defining `MIDDLE_LOG_PRICE` in gas tests:

```
LEAF_SIZE * LEVEL0_SIZE * LEVEL1_SIZE + (LEVEL2_SIZE / 2);
```

it should have been

```
LEAF_SIZE * LEVEL0_SIZE * LEVEL1_SIZE * (LEVEL2_SIZE / 2);
```

but this makes other ticks go beyond the limit, so I use

```
LEAF_SIZE * LEVEL0_SIZE * LEVEL1_SIZE * (LEVEL2_SIZE / 3);
```

The rest of the commit is adapting volumes to these high ticks: due to the max market order volume, with very high prices it becomes impossible to consume many offers, unless each offer has a negligible volume.

### DIFF
--- a/src/MgvOfferTaking.sol
+++ b/src/MgvOfferTaking.sol
@@ -21,7 +21,6 @@ import {
 import {MgvHasOffers} from "./MgvHasOffers.sol";
 import {TickLib} from "mgv_lib/TickLib.sol";
 import "mgv_lib/LogPriceConversionLib.sol";
-import "mgv_lib/Debug.sol";
 
 abstract contract MgvOfferTaking is MgvHasOffers {
   /* # MultiOrder struct */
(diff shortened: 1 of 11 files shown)


src/MgvOfferTaking.sol
```diff
@@ -21,6 +21,7 @@ import {
 import {MgvHasOffers} from "./MgvHasOffers.sol";
 import {TickLib} from "mgv_lib/TickLib.sol";
 import "mgv_lib/LogPriceConversionLib.sol";
+import "mgv_lib/Debug.sol";
```
lnist:
```suggestion
```

lnist (approved):

Issue #562, opened 2023-09-18
Big ticks renaming


adhusson:
Heavily paraphrasing @espendk 's writeup:

* `price` becomes `ratio`: a ratio [inbound]/[outbound] at which a taker/maker is willing to trade
* `tick` becomes `tickTreeIndex`
* `logRatio` becomes `tick`: an index into the discrete range  ]0; ∞[ of logs of ratios that contains 1 and with `ratio_tick / ratio_{tick-1} = 1 + tickSize`

  * This still keeps market terminology distinct from offer list terminology. Tick is a known term, both from trading generally and from Uniswap

* `ratio = (1+tickSize)^tick`
* `tick = log_{1+tickSize}(ratio)`  when `log_{1+tickSize}(ratio) % 1 == 0`
* `tickSize` is 1bps (0.0001): the minimum, relative ratio movement in an offer list
* `tickSpacing` : an offer list parameter that defines the spacing between valid ticks for that offer list: `tick % tickSpacing == 0`.
* `tickTreeIndex`: a number that uniquely identifies the the position in the tick tree where the offers for a tick is stored for a given tickSpacing
* `tickTreeIndex = tick/tickSpacing`

Slack discussion link https://mangroveitops.slack.com/archives/C05LDSH7FHD/p1694526478940409

adhusson:
Closed by efcf853f7486a3f713c19bf15369665ea4794167

espendk:
For the record and easy later reference, we ended up with this slightly modified terminology:

* `price` becomes `ratio`: a ratio [inbound]/[outbound] at which a taker/maker is willing to trade
* `tick` becomes `bin`
* `logRatio` becomes `tick`: an index into the discrete range  ]0; ∞[ of logs of ratios that contains 1 and with `ratio_tick / ratio_{tick-1} = 1 + tickSize`

  * This still keeps market terminology distinct from offer list terminology. Tick is a known term, both from trading generally and from Uniswap

* `ratio = (1+tickSize)^tick`
* `tick = log_{1+tickSize}(ratio)`  when `log_{1+tickSize}(ratio) % 1 == 0`
* `tickSize` is 1bps (0.0001): the minimum, relative ratio movement in an offer list
* `tickSpacing` : an offer list parameter that defines the spacing between valid ticks for that offer list: `tick % tickSpacing == 0`.
* `bin`: a number that uniquely identifies the position in the tick tree where the offers for a tick is stored for a given tickSpacing
[...]

Issue #475, opened 2023-08-28
Improve ease-of-life when writing tests and developing against Mangrove


adhusson:
- [x] Reorganise MgvLib.sol so all types can be imported&used without prefix. 
  * It is always possible to `import * as Prefix from "file";`
    * A solc bug prevents doing that in multiple files with the same namespace or a spurious "identifier already defined" error will appear.
  * Either import everything flat, or maintain a hierarchy which can be collapsed if needed
    * For instance `import "MgvLib.Main.sol"` gives `Structs.Offer` and `import MgvLib.Structs.sol"` gives `Offer`.
- [x] Change interface types and use user-defined types (e.g. `Tick` instead of `int`, OLKeyHash instead of bytes32 in mappings, etc) 
- [x] Give `LogPriceLib` its own user-defined type. / make LogPriceLib more comfortable to use (should logPrice be a type)
- [x] Rename `logPrice` -> `tick` and `tick` -> `??` (tickIndex?)
- [x] Check all mentioned solidity versions in the contracts
  - Everything looks reasonable

adhusson:
Closed by the PRs mentioned above

PR #568, opened 2023-09-20
Simplify constants

adhusson:
In preparation to #562, and related to #475.

- Replace constants such as `LEVEL0_SIZE`, `LEVEL1_SIZE`, etc with just `LEVEL_SIZE`
- Rename `LEVEL3` to `ROOT`
- Rename `MAX_LEVEL_SIZE` to `MAX_FIELD_SIZE`

Should make code easier to read through and closer to what it would have been without all the exploration re: how many levels to have, what size they should be, etc.

### DIFF
--- a/test/core/Constant.t.sol
+++ b/test/core/Constant.t.sol
@@ -24,14 +24,9 @@ contract ConstantsTest is Test2 {
 
   // Some BitLib.ctz64 relies on specific level sizes
   function test_level_sizes() public {
-    assertLe(LEVEL3_SIZE, int(MAX_LEVEL_SIZE), "bad level3 size");
-    assertGt(LEVEL3_SIZE, 0);
-    assertLe(LEVEL2_SIZE, int(MAX_LEVEL_SIZE), "bad level2 size");
-    assertGt(LEVEL2_SIZE, 0);
-    assertLe(LEVEL1_SIZE, int(MAX_LEVEL_SIZE), "bad level1 size");
-    assertGt(LEVEL1_SIZE, 0);
-    assertLe(LEVEL0_SIZE, int(MAX_LEVEL_SIZE), "bad level0 size");
-    assertGt(LEVEL0_SIZE, 0);
+    assertLe(ROOT_SIZE, int(MAX_FIELD_SIZE), "bad root size");
+    assertGt(ROOT_SIZE, 0);
+    assertLe(LEVEL_SIZE, int(MAX_FIELD_SIZE), "bad level size");
   }
 
   // checks that there is no overflow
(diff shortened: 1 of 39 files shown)


test/core/Constant.t.sol
```diff
@@ -24,14 +24,9 @@ contract ConstantsTest is Test2 {
...
-    assertGt(LEVEL3_SIZE, 0);
-    assertLe(LEVEL2_SIZE, int(MAX_LEVEL_SIZE), "bad level2 size");
-    assertGt(LEVEL2_SIZE, 0);
-    assertLe(LEVEL1_SIZE, int(MAX_LEVEL_SIZE), "bad level1 size");
-    assertGt(LEVEL1_SIZE, 0);
-    assertLe(LEVEL0_SIZE, int(MAX_LEVEL_SIZE), "bad level0 size");
-    assertGt(LEVEL0_SIZE, 0);
+    assertLe(ROOT_SIZE, int(MAX_FIELD_SIZE), "bad root size");
+    assertGt(ROOT_SIZE, 0);
+    assertLe(LEVEL_SIZE, int(MAX_FIELD_SIZE), "bad level size");
```
espendk: LEVEL_SIZE > 0 is still relevant I think?

test/core/MakerPosthook.t.sol
```diff
@@ -300,9 +300,9 @@ contract MakerPosthookTest is MangroveTest, IMaker {
     expectFrom($(mgv));
     emit OfferRetract(olKey.hash(), $(this), ofr, true);
     expectFrom($(mgv));
-    emit Credit($(this), 10699440000000000);
+    emit Credit($(this), 10697680000000000);
     expectFrom($(mgv));
-    emit OfferFail(olKey.hash(), $(tkr), ofr, 1 ether, 1 ether, 8500560000000000, "mgv/makerRevert");
+    emit OfferFail(olKey.hash(), $(tkr), ofr, 1 ether, 1 ether, 8502320000000000, "mgv/makerRevert");
```
espendk, line 305: I'm surprised that the penalty is higher now?
Must mean an increase in gas, but I thought this was merely a refactor of constants.

It's negligable, so not important

espendk (approved):
Nice simplification :-)

test/core/Constant.t.sol
adhusson: fixed in e71e93ca8fb9562683be441dfa60be87c26f3420

test/core/MakerPosthook.t.sol
adhusson: I thought I knew why (change in error message length), but your comment prompted me to check and I was wrong. I drilled it down to the function name change in `MgvView` from `level3` to `root`, which I assume modified the shape of the binary tree solidity descends when it matches a selector against a bytecode entry point.

espendk: 🤯

PR #572, opened 2023-09-23
Tick library

adhusson:
- TickTree tests use bin instead of tick
- Make a Tick type
- Use library dot syntax (`firstArg.fn(...)`) where possible
- External APIs expose `Tick` arguments instead of `int` arguments
- Utility `olKey.tick(bin)` and `olKey.bin(tick)` to make tests more readable

### DIFF
--- a/lib/BinLib.sol
+++ b/lib/BinLib.sol
@@ -100,14 +100,14 @@ library LeafLib {
 
   function setBinFirst(Leaf leaf, Bin bin, uint id) internal pure returns (Leaf) {
     unchecked {
-      uint posInLeaf = BinLib.posInLeaf(bin);
+      uint posInLeaf = bin.posInLeaf();
       return setPosFirstOrLast(leaf, posInLeaf, id, false);
     }
   }
 
   function setBinLast(Leaf leaf, Bin bin, uint id) internal pure returns (Leaf) {
     unchecked {
-      uint posInLeaf = BinLib.posInLeaf(bin);
+      uint posInLeaf = bin.posInLeaf();
       return setPosFirstOrLast(leaf, posInLeaf, id, true);
     }
   }
@@ -130,13 +130,13 @@ library LeafLib {
 
   function firstOfBin(Leaf leaf, Bin bin) internal pure returns (uint) {
     unchecked {
-      return firstOfPos(leaf, BinLib.posInLeaf(bin));
+      return firstOfPos(leaf, bin.posInLeaf());
     }
   }
 
   function lastOfBin(Leaf leaf, Bin bin) internal pure returns (uint) {
     unchecked {
-      return lastOfPos(leaf, BinLib.posInLeaf(bin));
+      return lastOfPos(leaf, bin.posInLeaf());
     }
   }
 
@@ -183,12 +183,6 @@ library BinLib {
     unchecked {
       return Bin.unwrap(bin) >= MIN_BIN && Bin.unwrap(bin) <= MAX_BIN;
     }
-  }
-
-  // Optimized conversion for ticks that are known to map exactly to a bin at the given tickSpacing,
-  // eg for offers in the offer list which are always written with a tick-aligned tick
-  function fromBinAlignedTick(Tick tick, uint tickSpacing) internal pure returns (Bin) {
-    return Bin.wrap(Tick.unwrap(tick) / int(tickSpacing));
   }
 
   // Utility for tests&unpacked structs, less gas-optimal
--- a/lib/BinLib.sol
+++ b/lib/BinLib.sol
@@ -195,7 +195,7 @@ library BinLib {
     }
   }
 
-  function toNearestTick(Bin bin, uint tickSpacing) internal pure returns (Tick) {
+  function tick(Bin bin, uint tickSpacing) internal pure returns (Tick) {
     return Tick.wrap(Bin.unwrap(bin) * int(tickSpacing));
   }
 
--- a/lib/BinLib.sol
+++ b/lib/BinLib.sol
@@ -2,6 +2,7 @@
 pragma solidity ^0.8.17;
 
 import "mgv_lib/Constants.sol";
+import {Tick} from "mgv_lib/TickLib.sol";
 import {BitLib} from "mgv_lib/BitLib.sol";
 import {console2 as csf} from "forge-std/console2.sol";
 import {LocalPacked} from "mgv_src/preprocessed/MgvLocal.post.sol";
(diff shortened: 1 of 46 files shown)


lib/BinLib.sol
```diff
@@ -213,6 +195,10 @@ library BinLib {
     }
   }
 
+  function toNearestTick(Bin bin, uint tickSpacing) internal pure returns (Tick) {
```
espendk: There's no approximation happening, so `Nearest` seems misleading.
```suggestion
  function toTick(Bin bin, uint tickSpacing) internal pure returns (Tick) {
```

lib/TickConversionLib.sol
espendk, line 1: This file had tick represented by both `int` and `int24`, though I don't know why.
Now it's all `int`, but calculations don't seem to have changed. Was the old `int24` restriction unnecessary?

lib/preprocessed/ToString.post.sol
```diff
@@ -60,24 +60,24 @@ function binBranchToString(Bin tick) pure returns (string memory) {
...
-  if (MIN_BIN > Bin.unwrap(tick) || Bin.unwrap(tick) > MAX_BIN) {
+  if (MIN_BIN > Bin.unwrap(bin) || Bin.unwrap(bin) > MAX_BIN) {
     suffix = "out of range";
-  } else if (MIN_BIN_ALLOWED > Bin.unwrap(tick) || Bin.unwrap(tick) > MAX_BIN_ALLOWED) {
-    suffix = "out of tick range";
+  } else if (MIN_BIN_ALLOWED > Bin.unwrap(bin) || Bin.unwrap(bin) > MAX_BIN_ALLOWED) {
+    suffix = "out of bin range";
   } else {
-    suffix = tickToString(TickLib.fromBin(tick,1));
+    suffix = toString(BinLib.toNearestTick(bin,1));
```
espendk: Shouldn't this use the `bin.method` syntax?
```suggestion
    suffix = toString(bin.toNearestTick(1));
```

preprocessing/structs.ts
```diff
@@ -160,7 +160,7 @@ library OfferUnpackedExtra {
   }
   function bin(OfferUnpacked memory offer, uint tickSpacing) internal pure returns (Bin) {
     // Offers are always stored with a tick that corresponds exactly to a tick
-    return BinLib.fromBinAlignedTick(offer.tick, tickSpacing);
+    return offer.tick.alignedToNearestBin(tickSpacing);
```
espendk: This reads a bit weird. Not sure how to improve it though

src/MgvLib.sol
```diff
@@ -34,6 +34,16 @@ library OLLib {
   function flipped(OLKey memory olKey) internal pure returns (OLKey memory) {
     return OLKey(olKey.inbound, olKey.outbound, olKey.tickSpacing);
   }
+
+  // Convert tick to bin according to olKey's tickSpacing
+  function bin(OLKey memory olKey, Tick _tick) internal pure returns (Bin) {
```
espendk: Unsurprisingly, I'm not a fan of hiding that this is a lossy conversion.

If do you choose to go that route, it should be consistent, so `toNearestBin` should be `toBin`. But again, I don't think we should do that 😅

src/MgvOfferMaking.sol
```diff
@@ -242,13 +242,15 @@ contract MgvOfferMaking is MgvHasOffers {
 
       uint tickSpacing = ofp.olKey.tickSpacing;
       // normalize tick to tickSpacing
-      Bin insertionBin = BinLib.tickToNearestHigherBin(insertionTick, tickSpacing);
-      insertionTick = TickLib.fromBin(insertionBin, tickSpacing);
-      require(TickLib.inRange(insertionTick), "mgv/writeOffer/tick/outOfRange");
+      Bin insertionBin = insertionTick.toNearestBin(tickSpacing);
+      insertionTick = insertionBin.toNearestTick(tickSpacing);
```
espendk: Reading this is weird as it looks like it's lossy in both directions. (already commented on the `toNearestTick` function, just thought I'd point out this example of why I think it should be renamed)

lib/TickLib.sol
```diff
@@ -5,26 +5,52 @@ import {Bin} from "mgv_lib/BinLib.sol";
...
+        bin = bin + 1;
+      }
+      return Bin.wrap(bin);
+    }
+  }
+
+
+  // Optimized conversion for ticks that are known to map exactly to a bin at the given tickSpacing,
+  // eg for offers in the offer list which are always written with a tick-aligned tick
+  function alignedToNearestBin(Tick tick, uint tickSpacing) internal pure returns (Bin) {
```
espendk: As noted elsewhere, this reads weird when used: `tick.alignedToNearestBin(tickSpacing)`. Also, since the `tick` must be aligned with a bin, the conversion is exact and `nearest` is misleading.

Maybe `tick.toAlignedBin()`?

src/MgvOfferMaking.sol
```diff
@@ -52,7 +52,7 @@ contract MgvOfferMaking is MgvHasOffers {
     }
   }
 
-  function newOfferByTick(OLKey memory olKey, int tick, uint gives, uint gasreq, uint gasprice)
+  function newOfferByTick(OLKey memory olKey, Tick tick, uint gives, uint gasreq, uint gasprice)
```
espendk, line 55: Is this going to annoying for callers? Or will it be natural for them to construct a `Tick` using `TickConversionLib`? How about off-chain, eg will ethers.js handle this nicely?

espendk (changes requested):
> TickTree tests use bin instead of tick

I thinks it's the gas tests that have been changed, no?

lib/TickConversionLib.sol
adhusson: yes they were unnecessary, I added to a test on this in b5669da443432ba707ff11b0a00b095a7921b6a2

src/MgvOfferMaking.sol
adhusson: ethers.js handles custom types well (converts to underlying), and I think for contract writers the exported type safety is worth it; normally you only input literals when writing tests and real contracts use conversion functions.

lib/BinLib.sol
adhusson: Introduced bin-tick conversion asymmetry + uniform with olKey methods in b5669da443432ba707ff11b0a00b095a7921b6a2

lib/preprocessed/ToString.post.sol
adhusson: fixed in b5669da443432ba707ff11b0a00b095a7921b6a2

src/MgvLib.sol
adhusson: fixed in b5669da443432ba707ff11b0a00b095a7921b6a2

src/MgvOfferMaking.sol
adhusson: fixed in b5669da443432ba707ff11b0a00b095a7921b6a2

preprocessing/structs.ts
adhusson: In 941ab6bc22a8e35859e659ba0dcb92f698da88d5 I replace both tick->bin functions with a single optimized assembly function. The move to assembly negates any loss caused by the modulo check and in gas tests reduces costs by ~140 gas most of the time, up to ~1000 gas (no gas increase observed).

lib/TickLib.sol
adhusson: removed in 941ab6bc22a8e35859e659ba0dcb92f698da88d5

espendk (approved):

PR #573, opened 2023-09-23
21 bit tick + constants tests

adhusson:
Depends on #572.

- Only use necessary space for tick
- Add tests to check that constants in `Constants.sol` match the sizes used in `structs.ts`.

### DIFF
--- a/lib/BinLib.sol
+++ b/lib/BinLib.sol
@@ -278,7 +278,7 @@ library BinLib {
   }
 
   // see note posIn*
-  // note with int20 bin we only use 2 bits in root
+  // note with int21 bin we only use 2 bits in root
   //   root single node
   // <--------------------->
   //  1                  0
--- a/lib/BinLib.sol
+++ b/lib/BinLib.sol
@@ -224,7 +224,7 @@ library BinLib {
     }
   }
 
-  // ok because 2^BIN_BITS%BINS_PER_LEAF=0
+  // ok because 2^TICK_BITS%BINS_PER_LEAF=0
   // note "posIn*"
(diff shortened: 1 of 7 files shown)


test/core/Leaf.t.sol
```diff
@@ -272,7 +285,7 @@ contract TickAndBinTest is Test {
     showTickApprox(1 ether, 1 ether);
   }
 
-  // int constant min_tick_abs = int(2**(BIN_BITS-1));
+  // int constant min_tick_abs = int(2**(TICK_BITS-1));
```
espendk: This comment is a bit weird - there's no mention of ticks in this test?

lib/BinLib.sol
```diff
@@ -278,7 +278,7 @@ library BinLib {
   }
 
   // see note posIn*
-  // note with int24 bin we only use 2 bits in root
+  // note with int20 bin we only use 2 bits in root
```
espendk: Tick is 21 bits, so I guess bin must be the same and cannot fit in `int20`?

adhusson: fixed in 1a4f5ad2db572b5b7ca732e7585ab93c0bc6c5a7

test/core/Leaf.t.sol
adhusson: fixed in 1a4f5ad2db572b5b7ca732e7585ab93c0bc6c5a7

espendk (approved):

adhusson:
same commits, just avoiding a merge by resetting&cherry-picking

PR #575, opened 2023-09-24
literal constants

adhusson:


test/core/Constant.t.sol
```diff
@@ -54,4 +53,27 @@ contract ConstantsTest is MangroveTest {
   function test_offer_bits() public {
     assertEq(OFFER_BITS, last_bits);
   }
+
+  // Since constant expressions are (as of solidity 0.8.21) not evaluated at compile time, we write all constants in Constants.sol as literals and test their values here:
+  function test_constant_expressions() public {
+    assertEq(MAX_BIN, -MIN_BIN - 1, "MAX_BIN");
```
lnist, line 59: Consider defining the constant expressions in the Constant.sol alongside the constants but named, e.g., MAX_BIN_EXPRESSION and then the test is simply asserting expression==constant.
This makes it easier to see the calculations in the actual code and not tests. But this is only a consider, it is fine to merge as is.

lnist (approved):

test/core/Constant.t.sol
adhusson: Will keep as-is for now but can revisit

PR #577, opened 2023-09-24
Add Tick.negate

lnist:


lnist:
Inlining does not work here, so we close it.

PR #579, opened 2023-09-24
Test/no snapshots

adhusson:
- Remove `vm.snapshot` in TickTree tests, re-run `setUp()` instead
- Must remove `pauseGasMetering` because the creation of a new `MgvAppendix` instance makes the `Mangrove` constructor fail at the second invocation of `setUp()` when gas metering is pause. I don't know why.

### DIFF
--- a/test/core/ticktree/TickTreeUpdateOffer.t.sol
+++ b/test/core/ticktree/TickTreeUpdateOffer.t.sol
@@ -185,16 +185,18 @@ contract TickTreeUpdateOfferTest is TickTreeTest {
     uint[] storage higherBinListSizeScenarios,
     uint[] storage lowerBinListSizeScenarios
   ) internal {
-    vm.pauseGasMetering();
+    // vm.pauseGasMetering();
     runBinScenarios(bin, higherBinListSizeScenarios, lowerBinListSizeScenarios);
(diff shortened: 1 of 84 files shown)


test/core/ticktree/TickTreeUpdateOffer.t.sol
```diff
@@ -185,16 +185,18 @@ contract TickTreeUpdateOfferTest is TickTreeTest {
     uint[] storage higherBinListSizeScenarios,
     uint[] storage lowerBinListSizeScenarios
   ) internal {
-    vm.pauseGasMetering();
+    // vm.pauseGasMetering();
     runBinScenarios(bin, higherBinListSizeScenarios, lowerBinListSizeScenarios);
-    vm.resumeGasMetering();
+    // vm.resumeGasMetering();
```
espendk: These commented out lines should be deleted, no?

espendk (approved):

test/core/ticktree/TickTreeUpdateOffer.t.sol
adhusson: yes! fixed in https://github.com/mangrovedao/mangrove-core/pull/579

PR #583, opened 2023-09-26
Feat/better offer volumes

adhusson:
tldr:
* `offer.gives` goes up to 160 bits
* marketOrder's `fillVolume` as well
* volume-based marketOrder & offer creation/updated are limited to 128 `[maker|taker][Wants|Gives]` (nontrivial & no time to allow the full price range with the API, which I'd say is fine since it exists mostly for convenience)

For `ratioFromVolumes`, use Remco Bloemen's [muldiv](https://xn--2-umb.com/21/muldiv/). For `inboundFromOutbound` and similar functions, use a new `mulDivPow2` that should be carefully reviewed.

**Why `mulDivPow2`**: `mulDiv(a,b,d)` (and `mulDivRoundUp`) return `a*b/d`. It throws if the result does not fit in 256 bits or if `d=0`. This is useful to compute a ratio from `gives` and `wants`. 

But to compute a `wants` from `gives` and `tick`, you start by converting a tick to a ratio represented as a `mantissa,exponent` pair, and `wants = gives*mantissa / 2^exponent`. The exponent may be > 255, so we can't use `mulDiv`. `mulDivPow2(a,b,e)` returns `a*b/(2<<e)`. It throws if the result does not fit in 256 bits. It also takes an extra `roundUp` argument.

In reviewing `mulDivPow2` you can assume that the first assembly block correctly returns the full `a*b` product, split into two words `prod1` and `prod0`.

### DIFF
--- a/lib/Constants.sol
+++ b/lib/Constants.sol
@@ -54,8 +54,8 @@ uint constant MAX_SAFE_GIVES = 33645117837968510856504196905370;
 // Without optimizer enabled it fails above 79. With optimizer and 200 runs it fails above 80. Set default a bit lower to be safe.
 uint constant INITIAL_MAX_RECURSION_DEPTH = 75;
 uint constant INITIAL_MAX_GASREQ_FOR_FAILING_OFFERS_MULTIPLIER = 3;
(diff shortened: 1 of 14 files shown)


adhusson:
Closing in favor of #597

PR #585, opened 2023-09-27
Tests for tick conversion function

adhusson:
Exhaustive tick math tests
- Check all tick->ratio conversions against a reference library
- Check that for all valid ticks `t`, `tickFromRatio(ratioFromTick(t)) == t`


test-ratios/generate_ratios_worker.js
```diff
@@ -0,0 +1,71 @@
...
+    `Worker ${worker_index}:\n  - Tick range [${start_tick};${end_tick}[:\n  - Writing to file ${FILE}\n  - Starting with price ${current_price}, tick ${current_tick}`,
+  );
+
+  while (current_tick < end_tick) {
+    const { sig, exp } = priceToFloat(current_price);
+    const formatted_sig = M.format(sig, { notation: "fixed" });
+    const res = { tick: current_tick, sig: formatted_sig, exp: exp };
+    fs.appendFileSync(FILE, JSON.stringify(res) + "\n");
+
+    current_price = M.multiply(current_price, M.bignumber("1.0001"));
```
espendk, line 63: Is nodejs clever enough to hoist the 1.0001 constant or are we parsing that every time? Maybe not terribly important, I'm just getting paranoid after seeing how stupid solc is 😅

test-ratios/TickRatioConversion.sol
```diff
@@ -0,0 +1,106 @@
...
+      }
+
+      // Check that tickFromRatio(ratioFromTick(tick)) == tick
+      Tick new_tick = TickLib.tickFromRatio(cur_sig, int(cur_exp));
+      assertEq(Tick.unwrap(new_tick), ref.tick, "tickFromRatio not right inverse of ratioFromTick");
+
+      // Memory management
+      assembly ("memory-safe") {
+        mstore(0x40, free_mem)
+      }
```
espendk, line 82: Sad that this is needed :-/

espendk (approved):
Very nice! Really increases confidence in the tick/ratio conversion :-)

test-ratios/generate_ratios_worker.js
adhusson: I don't think it does but there is no impact on runtime

test-ratios/TickRatioConversion.sol
adhusson: Agreed!

PR #588, opened 2023-09-27
Document ticks

adhusson:
## TODO after this PR
(but not as part of this PR, so that changes are easier to understand)
- Rename `Constant.t.sol` to `Constants.t.sol`
- Rename `BinLin.sol` to `TickTreeLib.sol`
- Rename Mgv*.post.sol files to *.post.sol
- Reorganize files so that the "files to be audited" set is easy to define. For instance, move `src/<relevant files>` to `src/core/` and `lib/<relevant files>` to `lib/core`.
  - Files to audit:
    - src/Mangrove.sol
    - src/Mgv*.sol
    - lib/BinLib.sol
    - lib/BitLib.sol
    - lib/Constants.sol
    - lib/DensityLib.sol
    - lib/TickLib.sol
    - lib/*Extra.sol

## List of changes 
- Document everything
- Add graphics
- Move solidity code out of `structs.ts`
- In tickmath, use closer approximations of log2(1.0001) and the min/max errors
  - Link to [note](https://hackmd.io/@mangrovedao/HJvl21zla) on how the min/max tick errors are obtained

### Renamings:
- pluralize `levelN` mappings/getters

| before | after |
|-|-|
| `leaf.eraseFromBin(Bin)` | `leaf.eraseAbove(Bin)` |
| `leaf.eraseToBin(Bin)` | `leaf.eraseBelow(Bin)` |
| `field.eraseLevelNFrom(Bin)` | `field.eraseAboveInLevelN(Bin)` |
| `field.eraseLevelNTo(Bin)` | `field.eraseBelowInLevelN(Bin)` |
| `unlockedMarketOnly()` | `unlockedOfferListOnly()` |
| `liveMarketOnly()` | `liveOfferListOnly()` |
| `leaf.firstOfferPosition()` | `leaf.bestNonEmptyBinPos()` |
| `leaf.getNextOfferId()` | `leaf.bestOfferId()` |
| `marketOrderByTick(..., uint maxGasreqForFailingOffers)` | `marketOrderByTickCustom(..., uint maxGasreqForFailingOffers)` |
| internal `mgvData` | `internalMgvData` |


doc/tick_tree_structure.png
lnist, line 1: "bet bin's first offer has id 32" -> "best"

lnist: Should we mention top-most bit is always 1 for once-used bins?

lib/BinLib.sol
```diff
@@ -7,53 +7,35 @@ import {BitLib} from "mgv_lib/BitLib.sol";
...
-type Leaf is uint;
-type DirtyLeaf is uint;
 
-using LeafLib for Leaf global;
-using DirtyLeafLib for DirtyLeaf global;
+/* # Libraries for tick tree manipulation 
 
-type Field is uint;
-type DirtyField is uint;
+Offers in Mangrove are structured in a tree so that offer insertion can happen in constant time.
```
lnist: insertion, removal, and update

lib/BinLib.sol
```diff
@@ -98,79 +79,111 @@ library LeafLib {
...
+  function bestOfferId(Leaf leaf) internal pure returns (uint offerId) {
     unchecked {
-      return firstOfPos(leaf,firstOfferPosition(leaf));
+      return firstOfPos(leaf,bestNonEmptyBinPos(leaf));
     }
   }
 }
 
 
+/* Bins are numbered from MIN_BIN to MAX_BIN. Each bin contains the offers at a given price. For a given `tickSpacing`, bins represent the following prices (centered on the central bin):
```
lnist: inclusive?

doc/local_config.png
lnist, line 1: It is not clear what are binaries and what are base10. The green best bin position is base10, right? is root base10?

lib/BinLib.sol
```diff
@@ -392,132 +380,219 @@ library FieldLib {
...
     unchecked {
       return level3Index * LEVEL_SIZE + int(level3.lastOnePosition());
     }
   }
 
  }
+
+/* ## Clean/Dirty Fields and Leaves
+
+To save gas, leaves and fields at never zeroed out but written with a dirty bit. This is especially helpful when the price oscillates quickly between two nodes.
```
lnist, line 533: "nodes": is that the terminology we use elsewhere?

lib/TickLib.sol
```diff
@@ -5,6 +5,18 @@ import {Bin} from "mgv_lib/BinLib.sol";
 import "mgv_lib/BitLib.sol";
 import "mgv_lib/Constants.sol";
 
+/* This file is inspired by Uniswap's approach to ticks, with the following notable changes:
```
lnist, line 8: Should we add a rationale for this difference?

lib/TickLib.sol
```diff
@@ -64,9 +87,14 @@ library TickLib {
     }
   }
 
-  // returns a normalized ratio within the max/min ratio range
-  // returns max_ratio if at least outboundAmt==0
-  // returns min_ratio if only inboundAmt==0
+  /* ### (outbound,inbound) → ratio */
+
+  /* `outboundFromInbound[Up]` converts an inbound amount (i.e. an `offer.wants` or a `takerGives`), to an outbound amount, following the price induced by `tick`. There's a rounding-up and a rounding-down variant.
```
lnist: this looks like copy-paste from above?

lib/TickLib.sol
```diff
@@ -89,20 +117,30 @@ library TickLib {
...
-  // return greatest tick t such that ratio(tick) <= input ratio
-  // does not expect a normalized ratio float
+  /* ### low-level ratio → tick */
+  /* Given `ratio`, return greatest tick `t` such that `ratioFromTick(t) <= ratio`. 
+  * Input ratio must be within the maximum and minimum ratios returned by the available ticks. 
+  * Does _not_ expected a normalized float.
+  
+  The function works as follows:
+  * Approximate log2(ratio) to the 13th fractional digit.
+  * Following <a href="https://hackmd.io/@mangrovedao/HJvl21zla">https://hackmd.io/@mangrovedao/HJvl21zla</a>, obtain `tickLow` and `tickHigh` such that log_1.0001(ratio) is between them
```
lnist: log_1.0001(ratio) => $log_{1.0001}(ratio)$ ?

preprocessing/structs.ts
```diff
@@ -1,15 +1,82 @@
...
+
+   The granularity of available price points in an offer list is controlled by the `tickSpacing` parameter. With a high `tickSpacing`, fewer price points are available, but the gas cost of market orders is lower since a smaller part of the tick tree has to be explored.
+
+   The available prices in an offer list are `1.0001^i` for all `MIN_TICK <= i <= MAX_TICK` such that `i % tickSpacing = 0`.
+
+   ## Tree storage
+
+   The exposed structure of offers is a list, but offers are actually stored in a tree we call a "tick tree". 
+
+   Thanks to the tree structure, offer insertion takes constant time (the height of the tree is fixed).
```
lnist: insertion,update.removal

preprocessing/structs.ts
```diff
@@ -247,39 +251,43 @@ library OfferDetailUnpackedExtra {
...
       { name: "level1", bits: 64, type: "Field", underlyingType: "uint" },
       { name: "root", bits: 2, type: "Field", underlyingType: "uint" },
-      /* * `offer_gasbase` is an overapproximation of the gas overhead associated with processing one offer. The Mangrove considers that a failed offer has used at least `offer_gasbase` gas. The actual field name is `kilo_offer_gasbase` and the accessor `offer_gasbase` returns `kilo_offer_gasbase*1e3`. Local to a pair, because the costs of calling `outbound_tkn` and `inbound_tkn`'s `transferFrom` are part of `offer_gasbase`. Should only be updated when ERC20 contracts change or when opcode prices change. */
+      /* * `offer_gasbase` is an overapproximation of the gas overhead associated with processing one offer. The Mangrove considers that a failed offer has used at least `offer_gasbase` gas. The actual field name is `kilo_offer_gasbase` and the accessor `offer_gasbase` returns `kilo_offer_gasbase*1e3`. Local to an offerList, because the costs of calling `outbound_tkn` and `inbound_tkn`'s `transferFrom` are part of `offer_gasbase`. Should only be updated when ERC20 contracts change or when opcode prices change. */
```
lnist: Could we be a bit more specific - processing one offer sounds like if a market order has 10 offers this should count what one can consume. However, to compensate a taker of a single large failing offer it needs to compensate for the entire market order TX - or to incentivize cleaners, to compensate for at least a complete clean TX.

lnist (commented):
First rounds of comments, continuing.

src/AUDIT_TRAIL_2.md
```diff
@@ -0,0 +1,49 @@
...
+
+The volume-based version of the market order also interprets the price induced by `takerWants,takerGives` as a true limit price (not as a limit average price).
+
+## Density is stored as a float and presented as a fixed-point number
+
+To reduce gas use, we cache more data than before in the `local` storage slot of each market. To make some room, the `density` parameter is no longer a 112 bit integer but a 9-bit floating-point number. It has a 2-bit mantissa and a 7-bit exponent. The minimum increment between two densities is 25%, but we do not need more precision than that.
+
+That change in representation also extends the available `density` range to fractional values (there are plausible scenarios where `density` should be < 1).
+
+The external representation of density is not a float but a 96.32 fixed-point number. It is easier to manipulate that way.
```
lnist, line 41: should we mention other bitsize changes? e.g. gasprice? gasbase?

src/MgvHasOffers.sol
```diff
@@ -28,7 +27,13 @@ contract MgvHasOffers is MgvCommon {
   /* # Misc. low-level functions */
   /* ## Offer deletion */
 
-  /* When an offer is deleted, it is marked as such by setting `gives` to 0. Note that provision accounting in Mangrove aims to minimize writes. Each maker `fund`s Mangrove to increase its balance. When an offer is created/updated, we compute how much should be reserved to pay for possible penalties. That amount can always be recomputed with `offerDetail.gasprice * (offerDetail.gasreq + offerDetail.offer_gasbase)`. The balance is updated to reflect the remaining available ethers.
+  /* When an offer is deleted, it is marked as such by setting `gives` to 0. Note that provision accounting in Mangrove aims to minimize writes. Each maker `fund`s Mangrove to increase its balance. When an offer is created/updated, we compute how much should be reserved to pay for possible penalties. That amount can always be recomputed with
```
lnist:
```suggestion
  /* When an offer is deleted, it is marked as such by setting `gives` to 0 and leaving other fields intact. Note that provision accounting in Mangrove aims to minimize writes. Each maker `fund`s Mangrove to increase its balance. When an offer is created/updated, we compute how much should be reserved to pay for possible penalties. That amount can always be recomputed with
```

lnist: We use that fact in strats, so I think its worth calling it out.

src/AUDIT_TRAIL_2.md
```diff
@@ -0,0 +1,49 @@
+# Audit trail for Mangrove v2
```
lnist, line 1: mention bit-masking for makers along with read-lock?

src/MgvOfferTaking.sol
```diff
@@ -73,11 +73,11 @@ abstract contract MgvOfferTaking is MgvHasOffers {
   {
     uint fillVolume = fillWants ? takerWants : takerGives;
     Tick maxTick = TickLib.tickFromVolumes(takerGives, takerWants);
-    return marketOrderByTick(olKey, maxTick, fillVolume, fillWants);
+    return generalMarketOrder(olKey, maxTick, fillVolume, fillWants, msg.sender, 0);
   }
 
   /* If the offer list is filled with failing offers such that the default `maxGasreqForFailingOffers` is inadequate, this version of the market order lets the taker specify an upper bound on the gas they are ready to spend on failing offers. */
-  function marketOrderByTick(
+  function marketOrderByTickCustom(
```
espendk, line 82: This name is a bit weird... I don't have a better idea thoug.
I guess it's been renamed to avoid the annoying syntax for calling overloaded methods in ethers.js?

src/MgvLib.sol
```diff
@@ -53,30 +56,32 @@ library MgvLib {
...
   //+clear+
 
-  /* `SingleOrder` holds data about an order-offer match in a struct. Used by `marketOrder` (and some of its nested functions) to avoid stack too deep errors. */
+  /* `SingleOrder` holds data about an order-offer match in a struct. Used by `marketOrder` (and some of its nested functions) to avoid stack too deep errors. It is used in market order and cleaning. */
   struct SingleOrder {
     OLKey olKey;
     uint offerId;
+    /* The `offer` given to the maker will be cleaned of `prev`/`next` pointers. */
     Offer offer;
     /* `wants`/`gives` mutate over execution. Initially the `wants`/`gives` from the taker's pov, then actual `wants`/`gives` adjusted by offer's ratio and volume. */
```
lnist: Is this true? 1) it is now called takerWants and not just wants (same for gives) and 2) the SingleOrder.takerWants/gives is the same as offer.gives/wants except for partial fill, so not from the takers pov.

src/AUDIT_TRAIL_2.md
```diff
@@ -0,0 +1,49 @@
...
+
+The goal of Mangrove v2 is constant-time offer insertion.
```
espendk: and offer update. Should also be updated in the following text.

src/MgvOfferMaking.sol
```diff
@@ -27,42 +27,32 @@ contract MgvOfferMaking is MgvHasOffers {
...
 
-     `gasprice` indicates an upper bound on the gasprice at which the maker is ready to be penalised if their offer fails. Any value below Mangrove's internal `gasprice` configuration value will be ignored.
+     `gasprice` indicates an upper bound on the gasprice (in mwei) at which the maker is ready to be penalised if their offer fails. Any value below Mangrove's internal `gasprice` configuration value will be ignored.
 
     `gasreq`, together with `gasprice`, will contribute to determining the penalty provision set aside by Mangrove from the market maker's `balanceOf` balance.
 
   An offer cannot be inserted in a closed market, nor when a reentrancy lock for `outbound_tkn`,`inbound_tkn` is on.
 
-  No more than $2^{32}-1$ offers can ever be created for one `outbound_tkn`,`inbound_tkn`, `tickSpacing` offerList.
+  No more than $2^{32}-1$ offers can ever be created for one (`outbound_tkn`,`inbound_tkn`, `tickSpacing`) offerList.
```
lnist:
```suggestion
  No more than $2^{32}-1$ offers can ever be created for one (`outbound`,`inbound`, `tickSpacing`) offerList.
```

src/MgvOfferMaking.sol
```diff
@@ -27,42 +27,32 @@ contract MgvOfferMaking is MgvHasOffers {
...
+  The actual contents of the function is in `writeOffer`, which is called by both `newOffer` and `updateOffer`. */
 
   function newOfferByTick(OLKey memory olKey, Tick tick, uint gives, uint gasreq, uint gasprice)
     public
     payable
     returns (uint offerId)
   {
     unchecked {
-      /* In preparation for calling `writeOffer`, we read the `outbound_tkn`,`inbound_tkn`, `tickSpacing` offerList configuration, check for reentrancy and market liveness, fill the `OfferPack` struct and increment the offerList's `last`. */
+      /* In preparation for calling `writeOffer`, we read the `outbound_tkn`,`inbound_tkn`, `tickSpacing` offerList configuration, check for reentrancy and offer list liveness, fill the `OfferPack` struct and increment the offerList's `last`. */
```
lnist, line 50: again outbound_tkn -> outbound? or will it clash with the meaning of "gives"?

espendk:
I've been through the code changes and all looks good!

Commit [`20c85e6` (#588)](https://github.com/mangrovedao/mangrove-core/pull/588/commits/20c85e698bb92222b267e2c3aefbe8e24971fed5) is a bit weird: It contains none of the mentioned test penalty changes. I noticed them in another commit (which I can't seem to find again rn), so probably just a rebase blunder.
Not important though, just writing it here in case anyone else wonders.

Will look at documentation next.

Edit: I pushed a small change to IMangrove as I noticed one of the taker functions was listed with the maker functions. Unrelated to this PR, but didn't want to create merge issues and felt safe and insignificant.

src/AUDIT_TRAIL_2.md
espendk, line 1: We should probably also mention the new read-lock.
From the ticks PR:


## Prevent makers from reading offer list during `makerExecute`
Makers can no longer read the offer list during `makerExecute` for the following reasons:

1. To prevent `makerExecute` from detecting cleaning
2. To avoid the gas cost of keeping the offer list up-to-date for `makerExecute`. A number of optimizations are in place which mean that `local`, leafs, and levels may not be up-to-date.

Two measures have been added to achieve this:
- The existing offer list lock is now enforced as a read-lock as well in all the view functions.
- All fields in the `SingleOrder` struct that pertain to the offer list/tick tree structure are cleared (set to zero) before it is passed to `makerExecute`.

espendk (commented):
Reviewed the new AUDIT_TRAIL_2.md.

Looks good, added a few notes.

src/MgvOfferTaking.sol
```diff
@@ -21,55 +21,63 @@ import {MgvHasOffers} from "./MgvHasOffers.sol";
...
-  * If `fillWants` is false, the taker is filling `gives` instead: the market order stops when `takerGives` units of `inbound_tkn` have been sold. With `fillWants` set to false, to sell a specific volume of `inbound_tkn` at any ratio, set `takerGives` to the amount desired and `takerWants` to $0$. */
-  function marketOrderByVolume(OLKey memory olKey, uint takerWants, uint takerGives, bool fillWants)
+  The market order stops when the price exceeds (an approximation of) 1.0001^`maxTick`, or when the end of the book has been reached, or:
+  * If `fillWants` is true, the market order stops when `fillVolume` units of `olKey.outbound` have been obtained. To buy a specific volume of `olKey.outbound` at any price, set `fillWants` to true, set `fillVolume` to volume you want to buy, and set `maxTick` to the `MAX_TICK` constant.
+  * If `fillWants` is false, the market order stops when `fillVolume` units of `olKey.inbound` have been sold. To sell a specific volume of `olKey.inbound` at any price, set `fillWants` to false, set `fillVolume` to the volume you want to sell, and set `maxTick` to the `MAX_TICK` constant. */
```
lnist: would it make sense to here state how much of the _other_ token you can end up getting/giving based on the maxTick? For fillWants=true it is important the user knows how much takerGives to approve.

src/MgvOfferTaking.sol
lnist: marketOrderByTickAndMaxGasreqForFailingOffers?

lnist (commented):
Second batch

espendk:
The `contactMap.png` is outdated. It doesn't seem to be a DrawIO file, so not sure how to update it?

billede

preprocessing/structs.ts
```diff
@@ -1,15 +1,82 @@
...
+   * When an offer is executed, Mangrove does the following:
      1. Flashloan some `inbound_tkn` to the offer's `maker`.
      2. Call an arbitrary `execute` function on that address.
      3. Transfer back some `outbound_tkn`.
      4. Call back the `maker` so they can update their offers.
     
    **Let the devs know about any error, typo etc, by contacting 	devs@mangrove.exchange**
+
+
+  There is one Mangrove contract that manages all tradeable offerLists. This reduces deployment costs for new offerLists and lets market makers have all their provision for all offerLists in the same place.
```
espendk: It seems most of the doc uses "offerList(s)". Line 2 uses "offer lists" though and I think it reads better. But I don't feel strongly about it, it should just be consistent :-)

preprocessing/structs.ts
```diff
@@ -1,15 +1,82 @@
...
+
+   The interaction map between the different actors is as follows:
+   <img src="./contactMap.png" width="190%"></img>
```
espendk, line 20: The contactMap.png is outdated. It doesn't seem to be a DrawIO file, so not sure how to update it?

billede

preprocessing/structs.ts
espendk: Hmm, we also expose the tick tree (eg levels), so I think this should be rephrased.

Maybe simply:
```suggestion
   Offers are stored in a tree we call a "tick tree". Thanks to this tree structure, offer operations (insert, update, and retract) take constant time (the height of the tree is fixed).
```

src/MgvOfferTaking.sol
```diff
@@ -183,22 +236,23 @@ abstract contract MgvOfferTaking is MgvHasOffers {
...
-      unlockedMarketOnly(sor.local);
+      /* For the market order to start, the offer list needs to be both active, and not currently protected from reentrancy. */
+      activeOfferListOnly(sor.global, sor.local);
+      unlockedOfferListOnly(sor.local);
 
       /* ### Initialization */
       /* The market order will operate as follows : it will go through offers from best to worse, starting from `offerId`, and: */
-      /* * will maintain remaining `takerWants` and `takerGives` values. The initial `takerGives/takerWants` ratio is the average ratio the taker will accept. Better ratios may be found early in the book, and worse ones later.
-       * will not set `prev`/`next` pointers to their correct locations at each offer taken (this is an optimization enabled by forbidding reentrancy).
+      /* keep an up-to-date `fillWants`.
```
lnist: fillWants->fillVolume ?

src/MgvOfferTaking.sol
```diff
@@ -183,22 +236,23 @@ abstract contract MgvOfferTaking is MgvHasOffers {
...
+       * not set `prev`/`next` pointers to their correct locations at each offer taken (this is an optimization enabled by forbidding reentrancy).
        * after consuming a segment of offers, will update the current `best` offer to be the best remaining offer on the book. */
 
-      /* We start be enabling the reentrancy lock for this (`outbound_tkn`,`inbound_tkn`) offerList. */
+      /* We start be enabling the reentrancy lock for this (`olKey.outbound`,`olKey.inbound`) offer list. */
```
lnist:
```suggestion
      /* We start be enabling the reentrancy lock for this (`olKey.outbound`,`olKey.inbound`, `olKey.tickSpacing`) offer list. */
```

preprocessing/structs.ts
```diff
@@ -1,15 +1,82 @@
...
+
+   The exposed structure of offers is a list, but offers are actually stored in a tree we call a "tick tree". 
+
+   Thanks to the tree structure, offer insertion takes constant time (the height of the tree is fixed).
+
+   ### Bins
+
+   <img src="./bin.png" width="80%"></img>
+
+   Below the bottom of the tree are _bins_. A bin is a doubly-linked list of offers. All offers in a bin have the same tick. During a market order, offers in a bin are executed in order, from the first to the last. Inserted offers are always appended at the end of a bin.
```
espendk: We should use either "doubly-linked" or "doubly linked" consistently.

preprocessing/structs.ts
```diff
@@ -1,15 +1,82 @@
...
+
+   The available prices in an offer list are `1.0001^i` for all `MIN_TICK <= i <= MAX_TICK` such that `i % tickSpacing = 0`.
+
+   ## Tree storage
+
+   The exposed structure of offers is a list, but offers are actually stored in a tree we call a "tick tree". 
+
+   Thanks to the tree structure, offer insertion takes constant time (the height of the tree is fixed).
+
+   ### Bins
```
espendk, line 37: I think it would read better if the just-mentioned tick tree was explained first.

src/MgvOfferTaking.sol
```diff
@@ -262,21 +318,17 @@ abstract contract MgvOfferTaking is MgvHasOffers {
...
+          mor.fillVolume -= mor.fillWants ? takerWants : takerGives;
         }
 
         /* We move `sor` to the next offer. Note that the current state is inconsistent, since we have not yet updated `sor.offerDetails`. */
-        /* It is known statically that `mor.initialGives - mor.totalGave` does not underflow since
-          1. `mor.totalGave` was increased by `sor.takerGives` during `execute`,
-          2. `sor.takerGives` was at most `mor.initialGives - mor.totalGave` from earlier step,
-          3. `sor.takerGives` may have been clamped _down_ during `execute` (to "`offer.wants`" if the offer is entirely consumed, or to `makerWouldWant`, cf. code of `execute`).
-        */
         (sor.offerId, sor.local) = getNextBest(offerList, mor, sor.offer, sor.local, sor.olKey.tickSpacing);
```
lnist:
```suggestion
        (sor.offerId, sor.local) = getNextBest(offerList, mor, offer, sor.local, sor.olKey.tickSpacing);
```

preprocessing/structs.ts
```diff
@@ -1,15 +1,82 @@
...
+   Below the bottom of the tree are _bins_. A bin is a doubly-linked list of offers. All offers in a bin have the same tick. During a market order, offers in a bin are executed in order, from the first to the last. Inserted offers are always appended at the end of a bin. 
+
+   Bins are laid in sequence. If all the offers in a bin have tick `t`, all the offers in the following bin have tick `t+tickSpacing`.
+
+   Bins are identified by their index in the bin sequence, from the first (`MIN_BIN`) to the last (`MAX_BIN`). The number of available bins is larger than the number of available ticks, so some bins will never be used.
+
+   ### Tree structure
+
+   The structure of the tick tree is as follows:
+   <img src="./tick_tree_structure.png" width="190%"></img>
```
espendk, line 50: There's a typo: "bet bin's" -> "best bin's".

Also, I think it would be good to show the offers of a bin in the overview diagram, eg something like this:

billede

preprocessing/structs.ts
```diff
@@ -1,15 +1,82 @@
...
+
+   Thanks to the tree structure, offer insertion takes constant time (the height of the tree is fixed).
+
+   ### Bins
+
+   <img src="./bin.png" width="80%"></img>
+
+   Below the bottom of the tree are _bins_. A bin is a doubly-linked list of offers. All offers in a bin have the same tick. During a market order, offers in a bin are executed in order, from the first to the last. Inserted offers are always appended at the end of a bin. 
+
+   Bins are laid in sequence. If all the offers in a bin have tick `t`, all the offers in the following bin have tick `t+tickSpacing`.
```
espendk: In the code, bins have a tick and hold a linked list of offers with that tick. I think this explanation would be simpler if we didn't have to say "the tick of the offers in a bin" but just could say "the bin's tick".

lnist (commented):
Final set of comments

preprocessing/structs.ts
```diff
@@ -1,15 +1,82 @@
...
+
+   Bins are identified by their index in the bin sequence, from the first (`MIN_BIN`) to the last (`MAX_BIN`). The number of available bins is larger than the number of available ticks, so some bins will never be used.
+
```
espendk: These paragraphs seem surprisingly convoluted. I we broaden the "bin" concept to include the corresponding tick and index, I think the text could be simplified.

Eg, in the code, bins are treated as if they "have a tick" (ie it's an attribute of the bin) and hold a linked list of offers with that tick. I think this explanation would be simpler if we didn't have to say "the tick of the offers in a bin" but just could say "the bin's tick".

Similarly for the index of a bin.

preprocessing/structs.ts
```diff
@@ -1,15 +1,82 @@
...
+   ### Tree structure
+
+   The structure of the tick tree is as follows:
+   <img src="./tick_tree_structure.png" width="190%"></img>
+
+   At the bottom of the tree, leaves contain information about 4 bins: their first and last offer. Offer ids use 32 bits, so leaves use 256 bits.
+
+   When a market order runs, execution starts with the first offer of the smallest nonempty bin and continues from there.
```
espendk: "smallest" feels ambiguous (are we talking about the size of the bin?). Perhaps "lowest-index bin" ?

Also "nonempty" -> "non-empty"

preprocessing/structs.ts
```diff
@@ -1,15 +1,82 @@
...
    * Each offer promises `outbound_tkn` and requests `inbound_tkn`.
    * Each offer has an attached `maker` address.
-   * In the normal operation mode (called Mangrove for Maker Mangrove), when an offer is executed, we:
+   * When an offer is executed, Mangrove does the following:
      1. Flashloan some `inbound_tkn` to the offer's `maker`.
      2. Call an arbitrary `execute` function on that address.
      3. Transfer back some `outbound_tkn`.
      4. Call back the `maker` so they can update their offers.
     
    **Let the devs know about any error, typo etc, by contacting 	devs@mangrove.exchange**
```
lnist, line 12: should we still write an explicit email here?

preprocessing/structs.ts
lnist: Actually, the description in offer_Detail is better:
```
offer_gasbase represents the gas overhead used by processing the offer inside Mangrove + the overhead of initiating an entire order, in 1k gas increments.
```

preprocessing/structs.ts
```diff
@@ -1,15 +1,82 @@
...
+   The structure of the tick tree is as follows:
+   <img src="./tick_tree_structure.png" width="190%"></img>
+
+   At the bottom of the tree, leaves contain information about 4 bins: their first and last offer. Offer ids use 32 bits, so leaves use 256 bits.
+
+   When a market order runs, execution starts with the first offer of the smallest nonempty bin and continues from there.
+
+   Once all the offers in the smallest bin have been executed, the next nonempty bin is found. If the leaf of the current bin now only has empty bins, the node above the current leaf contains a bitfield with information about all its children: if the *i*th bit of the node is set, its *i*th child has at least one nonempty bin. Otherwise, its *i*th child only has empty bins.
+
+   To find the next nonempty bin, it may be necessary to keep going up the tree until the root is reached. At each level above, the bitfield rule applies similarly: the *i*th bit of a node is set iff its *i*th child has at least one set bit.
```
espendk: > If the leaf of the current bin now only has empty bins, the node above the current leaf contains a bitfield with information about all its children

The node above always contains this bitfield :-)

Maybe something like this instead:

```suggestion
   Once all the offers in the smallest bin have been executed, the next non-empty bin is found. If the leaf of the current bin now only has empty bins, the tree must be searched for the next non-empty bin, starting at the node above the leaf:
   
   A non-leaf node contains a bit field with information about all its children: if the *i*th bit of the node is set, its *i*th child has at least one non-empty bin. Otherwise, its *i*th child only has empty bins.

   To find the next non-empty bin, it may be necessary to keep going up the tree until the root is reached. At each level above, the bit field rule applies similarly: the *i*th bit of a node is set iff its *i*th child has at least one set bit.
```

preprocessing/structs.ts
```diff
@@ -247,39 +251,43 @@ library OfferDetailUnpackedExtra {
...
-      /* * `offer_gasbase` is an overapproximation of the gas overhead associated with processing one offer. The Mangrove considers that a failed offer has used at least `offer_gasbase` gas. The actual field name is `kilo_offer_gasbase` and the accessor `offer_gasbase` returns `kilo_offer_gasbase*1e3`. Local to a pair, because the costs of calling `outbound_tkn` and `inbound_tkn`'s `transferFrom` are part of `offer_gasbase`. Should only be updated when ERC20 contracts change or when opcode prices change. */
+      /* * `offer_gasbase` is an overapproximation of the gas overhead associated with processing one offer. The Mangrove considers that a failed offer has used at least `offer_gasbase` gas. The actual field name is `kilo_offer_gasbase` and the accessor `offer_gasbase` returns `kilo_offer_gasbase*1e3`. Local to an offerList, because the costs of calling `outbound_tkn` and `inbound_tkn`'s `transferFrom` are part of `offer_gasbase`. Should only be updated when ERC20 contracts change or when opcode prices change. */
       fields.kilo_offer_gasbase,
       /* * If `lock` is true, orders may not be added nor executed.
```
lnist: , nor the offer list read

preprocessing/structs.ts
```diff
@@ -1,15 +1,82 @@
...
+
+   Once all the offers in the smallest bin have been executed, the next nonempty bin is found. If the leaf of the current bin now only has empty bins, the node above the current leaf contains a bitfield with information about all its children: if the *i*th bit of the node is set, its *i*th child has at least one nonempty bin. Otherwise, its *i*th child only has empty bins.
+
+   To find the next nonempty bin, it may be necessary to keep going up the tree until the root is reached. At each level above, the bitfield rule applies similarly: the *i*th bit of a node is set iff its *i*th child has at least one set bit.
+
+   Once a node with a set bit is found, its rightmost nonempty child is examined, and so on until the next nonempty bin is reached, and the first offer of that bin gets executed.
+
+   ## Caching
+
+   At any time, if there is at least one offer in the offer list, the best offer is the first offer of the bin containing the cheapest offers; and that bin is the best bin. Its parent is the best `level3`, whose parent is the best `level2`, and whose parent is the best `level3` (whose parent is the `root`).
```
```suggestion
   At any time, if there is at least one offer in the offer list, the best offer is the first offer of the bin containing the cheapest offers; and that bin is the best bin. Its parent is the best `level3`, whose parent is the best `level2`, and whose parent is the best `level1` (whose parent is the `root`). 
```

preprocessing/structs.ts
```diff
@@ -1,15 +1,82 @@
...
+
+   To find the next nonempty bin, it may be necessary to keep going up the tree until the root is reached. At each level above, the bitfield rule applies similarly: the *i*th bit of a node is set iff its *i*th child has at least one set bit.
+
+   Once a node with a set bit is found, its rightmost nonempty child is examined, and so on until the next nonempty bin is reached, and the first offer of that bin gets executed.
+
+   ## Caching
+
+   At any time, if there is at least one offer in the offer list, the best offer is the first offer of the bin containing the cheapest offers; and that bin is the best bin. Its parent is the best `level3`, whose parent is the best `level2`, and whose parent is the best `level3` (whose parent is the `root`). 
+
+   The `root`, the best `level1`, the  best`level2`, and the best `level3` are always stored in `local`. The position of the best bin in the best leaf is also stored in `local`.
```
```suggestion
   The `root`, the best `level1`, the  best `level2`, and the best `level3` are always stored in `local`. The position of the best bin in the best leaf is also stored in `local`. 
```

src/MgvOfferMaking.sol
lnist: I am not sure when we reference using outbound_tkn or when we use outbound :) If we want to change this we should probably start searching for outbound_tkn as it is used in many places.

preprocessing/structs.ts
```diff
@@ -1,15 +1,82 @@
...
+   The `root`, the best `level1`, the  best`level2`, and the best `level3` are always stored in `local`. The position of the best bin in the best leaf is also stored in `local`. 
+
+   This data is useful for two things:
+   - Read and modify the tree branch of the best offer without additional storage reads and writes (except for modifying the best leaf).
+   - Know the price of the best offer without an additional storage read (it can be computed using the set bit positions in each level, the position of the best bin in the best leaf, and the value of `tickSpacing`).
+
+   The structure of the local config is as follows:
+   <img src="./local_config.png" width="140%"></img>
+
+  This caching means that as the price oscillates within a more restricted range, the fewer additional storage read/writes have to be performed (as most of the information is available in `local`) when there is a market order or an offer insertion/update.
```
```suggestion
  This caching means that as the price oscillates within a more restricted range, fewer additional storage read/writes have to be performed (as most of the information is available in `local`) when there is a market order or an offer insertion/update.
```

preprocessing/structs.ts
```diff
@@ -1,15 +1,82 @@
...
+
+  ## Some numbers
+  Here are some useful numbers, for reference:
+   <img src="./numbers.png" width="70%"></img>
```
espendk, line 84: Nit: Spacing could be improved:

billede

`min_price` could also be interesting to have

preprocessing/structs.ts
```diff
@@ -1,15 +1,82 @@
...
+
  */
 //+clear+
```
espendk, line 88: Applies to the following lines, but I can't comment on those...

The formatting is lost in the html:

billede

lib/BinLib.sol
adhusson: Yes, hopefully consistently -- nodes is leaves U fields

espendk (commented):
checkpoint :-)

src/MgvLib.sol
```diff
@@ -53,30 +56,32 @@ library MgvLib {
...
     bytes32 makerData;
     /* `mgvData` is an [internal Mangrove status code](#MgvOfferTaking/statusCodes) code. */
     bytes32 mgvData;
   }
 
-  /* `CleanTarget` holds data about an offer that should be cleaned, i.e. made to fail by executing it with the specified volume. */
+  /* `CleanTarget` holds data about an offer that should be cleaned, i.e. made to fail by executing it with the specified volume. It is used in `MgvOfferTaking.cleanByImpersonation`. */
   struct CleanTarget {
     uint offerId;
     Tick tick;
```
lnist, line 87: "The events emitted for use by bots are listed here:" -> "The events emitted are listed here:"

lib/BinLib.sol
adhusson: yes -- I hoped the name gave it away :p

preprocessing/structs.ts
adhusson: I didn't understand that?

preprocessing/structs.ts
adhusson: why not?

preprocessing/structs.ts
adhusson: adapted "smallest", but I like nonempty! Cambridge Dictionary accepts it :p https://dictionary.cambridge.org/fr/dictionnaire/anglais/non-empty

image

preprocessing/structs.ts
adhusson: I agree that saying "bin's tick" is good (but this only makes sense in the context of a tickSpacing). However I didn't understand how I should change the paragraph here?

src/MgvOfferMaking.sol
```diff
@@ -128,41 +120,49 @@ contract MgvOfferMaking is MgvHasOffers {
...
       Offer offer = offerData.offer;
       OfferDetail offerDetail = offerData.detail;
       require(msg.sender == offerDetail.maker(), "mgv/retractOffer/unauthorized");
 
-      /* Here, we are about to un-live an offer, so we start by taking it out of the book by stitching together its previous and next offers. Note that unconditionally calling `stitchOffers` would break the book since it would connect offers that may have since moved. */
+      /* Here, we are about to un-live an offer, so we start by taking it out of the tick tree. Note that unconditionally calling `stitchOffers` even if the offer is not `live` would break the offer list since it would connect offers that may have since moved. */
       if (offer.isLive()) {
         Local oldLocal = local;
         (local,) = dislodgeOffer(offerList, olKey.tickSpacing, offer, local, local.bestBin(), true);
-        /* If calling `stitchOffers` has changed the current `best` offer, we update the storage. */
```
lnist, line 157: stitchOffers no longer exists

src/MgvOfferTaking.sol
```diff
@@ -21,55 +21,63 @@ import {MgvHasOffers} from "./MgvHasOffers.sol";
...
 
   /* # Market Orders */
 
   /* ## Market Order */
   //+clear+
 
-  /* A market order specifies a (`outbound_tkn`,`inbound_tkn`,`tickSpacing`) offer list, a desired total amount of `outbound_tkn` (`takerWants`), and an available total amount of `inbound_tkn` (`takerGives`). It returns four `uint`s: the total amount of `outbound_tkn` received, the total amount of `inbound_tkn` spent, the penalty received by msg.sender (in wei), and the fee paid by the taker (in wei).
+  /* A market order specifies a (`outbound`, `inbound`,`tickSpacing`) offer list, a limit price it is ready to pay (in the form of `maxTick`, the log base 1.0001 of the price), and a volume `fillVolume`. If `fillWants` is true, that volume is the amount of `olKey.outbound` the taker wants to buy. If `fillWants` is false, that volume is the amount of `olKey.inbound` the taker wants to sell.
+  
+  It returns four `uint`s: the total amount of `olKey.outbound` received, the total amount of `olKey.inbound` spent, the penalty received by msg.sender (in wei), and the fee paid by the taker (in wei).
```
lnist:
```suggestion
  It returns four `uint`s: the total amount of `olKey.outbound` received, the total amount of `olKey.inbound` spent, the penalty received by msg.sender (in wei), and the fee paid by the taker (in wei of `olKey.outbound`).
```

preprocessing/structs.ts
adhusson: I think this is a omnigraffle file, Jean has the sources. I think we could leave it for now and redo this one and the others with a proper Mangrove style later, what do you think?

preprocessing/structs.ts
adhusson: Agree but will address in a separate commit for clarity

preprocessing/structs.ts
adhusson: Yes, it's a limitation of solcco, do you it's OK to leave as is for now?

src/MgvOfferTaking.sol
```diff
@@ -467,55 +508,60 @@ abstract contract MgvOfferTaking is MgvHasOffers {
...
   /* ## Execute */
   /* Execution of the offer will be attempted with volume limited by the offer's advertised volume.
-     NB: The caller must ensure that the ratio of the offer is low enough; This is not checked here.
+     **Warning**: The caller must ensure that the price of the offer is low enough; This is not checked here.
 
      Summary of the meaning of the return values:
     * `gasused` is the gas consumed by the execution
     * `makerData` is the data returned after executing the offer
-    * `mgvData` is an [internal Mangrove status code](#MgvOfferTaking/statusCodes).
+    * <a id="MgvOfferTaking/internalStatusCodes"></a>`mgvData` is a status code internal to `execute`. It can hold [any value that `mgvData` can hold](#MgvOfferTaking/statusCodes). Within `execute`, it can additionally hold the following values:
```
lnist:
```suggestion
    * <a id="MgvOfferTaking/internalStatusCodes"></a>`internalMgvData` is a status code internal to `execute`. It can hold [any value that `mgvData` can hold](#MgvOfferTaking/statusCodes). Within `execute`, it can additionally hold the following values:
```

src/AUDIT_TRAIL_2.md
adhusson: I think this is addressed by copy/pasting Espen's pasted section, but let me know if not.

preprocessing/structs.ts
lnist: Since we have a whole page (with docs), discord, twitter, etc. it seems like a weird place to send comments. But I don't mind :)

src/MgvOfferTaking.sol
```diff
@@ -524,53 +570,58 @@ abstract contract MgvOfferTaking is MgvHasOffers {
...
 
   In detail:
-  1. Flashloans `takerGives` `inbound_tkn` from the taker to the maker and returns false if the loan fails.
+  1. Flashloans `takerGives` units of `sor.olKey.inbound` from the taker to the maker and returns false if the loan fails.
   2. Runs `offerDetail.maker`'s `execute` function.
-  3. Returns the result of the operations, with optional makerData to help the maker debug.
+  3. Returns the result of the operations, with optional `makerData` to help the maker debug.
 
   Made virtual so tests can instrument the function.
   */
```
lnist, line 629: "/* Note : as a possible future feature, we"... delete that section? Since we actually argued that stopping at gasleft is an issue due to gas estimation methodology.

src/MgvLib.sol
adhusson: 1) Good catch, should rename

2) It refers to mutation inside the code, which is still true given this in `MgvOfferTaking.execute`:

```solidity
        if ((mor.fillWants && offerGives <= fillVolume) || (!mor.fillWants && offerWants <= fillVolume)) {
          sor.takerWants = offerGives;
          sor.takerGives = offerWants;
        } else {
          if (mor.fillWants) {
            /* While a possible `offer.wants()=0` is the maker's responsibility, a small enough partial fill may round to 0, so we round up. It is immaterial but more fair to the maker. */
            sor.takerGives = sor.offer.tick().inboundFromOutboundUp(fillVolume);
            sor.takerWants = fillVolume;
          } else {
            sor.takerWants = sor.offer.tick().outboundFromInbound(fillVolume);
            sor.takerGives = fillVolume;
          }
```

src/MgvOfferTakingWithPermit.sol
```diff
@@ -68,7 +68,7 @@ abstract contract MgvOfferTakingWithPermit is MgvOfferTaking {
 
   /* The delegate version of `marketOrder` is `marketOrderFor`, which takes a `taker` address as additional argument. Penalties incurred by failed offers will still be sent to `msg.sender`, but exchanged amounts will be transferred from and to the `taker`. If the `msg.sender`'s allowance for the given `outbound_tkn`,`inbound_tkn` and `taker` are strictly less than the total amount eventually spent by `taker`, the call will fail. */
```
lnist, line 69: (why can't I comment on non-changed lines far away...)
comment for `deductSenderAllowance` states: "its both checks that"
changes "its" to "it"

src/MgvOfferTaking.sol
adhusson: 😱

src/MgvOfferMaking.sol
adhusson: Ah yes it's because now there is `olKey.outbound` (not `olKey.outbound_tkn`). wdyt?

preprocessing/structs.ts
adhusson: We can also add a link but a direct contact point is always pleasant :) Link to https://docs.mangrove.exchange/ ?

src/MgvLib.sol
```diff
@@ -91,20 +96,22 @@ interface HasMgvEvents {
   /* 
     Events in solidity is a hard thing to do in a optimal way. If you look at it as a purely gas efficient issue, you want to emit as few events as possible and with as few fields as possible. But as events also has to be usable for an off chain user, then doing this is not always the best solution.
 
-    We tried to list the main points that we would like to do with events.
+    We tried to list the main forces that are inflicted upon events.
```
lnist:
```suggestion
    We tried to list the main forces that drive which events and data to emit:
```

lnist: c.f. Espen on other PR :)

src/AUDIT_TRAIL_2.md
adhusson: But it's specifically insertion (whether as part of an update) that used to not take constant time, so I think it's ok to just mention that?

src/MgvLib.sol
```diff
@@ -91,20 +96,22 @@ interface HasMgvEvents {
...
 
-    1. Use as little gas as possible
+    1. Use as little gas as possible.
     2. An indexer should be able to keep track of the state of Mangrove.
     3. Doing RPC calls directly, should be able to find offers and other information based on offerList, maker, taker, offer id, etc.
 
-    These 3 points all have there own direction and it is therefore not possible to find a solution that is optimal for all 3 points.
-    We have therefore tried to find a solution that is a good balance between the all 3 points.
+    These forces all pull in their own direction and it is therefore not possible to find a solution that is optimal for all three.
+    The following events are therefore an attempt to balance the trade-offs.
```
lnist:
```suggestion
    These forces are in conflict and it is impossible to find a solution that is optimal for all three.
    The following events are therefore an attempt to balance the trade-offs.
```

lnist: also c.f. Espen :)

doc/tick_tree_structure.png
adhusson: Hmm I think that's more detail than is good for the reader at this point

src/AUDIT_TRAIL_2.md
lnist: To the casual reader I don't think update/retract~=insertion

src/MgvLib.sol
lnist: > Initially the `wants`/`gives` from the taker's pov

I would say its from the maker's pov and then only adjusted for partial fill in the end to:

>, then actual `wants`/`gives` adjusted by offer's ratio and volume.

preprocessing/structs.ts
lnist: Yes, I think that is a good idea.

src/AUDIT_TRAIL_2.md
espendk: Ok. I think it would be clearer mentioning this, but I don't feel strongly about this and we need to move quickly, so 🤷🏻‍♂️

preprocessing/structs.ts
adhusson: I didn't understand that

preprocessing/structs.ts
espendk: alright. Do we have a backlog somewhere for things to fix post code freeze?

preprocessing/structs.ts
adhusson: Ahh ok I was trying to avoid talking about tickSpacing (too much) here. Will try to rephrase

preprocessing/structs.ts
espendk: I would move the "Tree structure" section up, so it's before the "Bins" section.

preprocessing/structs.ts
adhusson: done in d43902cd8bdaf61d86efe49763d542e80198d2e5

preprocessing/structs.ts
espendk: doh, this should have been deleted, it was superseded by the next comment

preprocessing/structs.ts
espendk: Alright, I yield (haven't seen "nonempty" before)

preprocessing/structs.ts
espendk: sure!

preprocessing/structs.ts
adhusson: No, I'll add it to the ticks PR.

doc/tick_tree_structure.png
lnist: Yeah, and its clear in the code.

preprocessing/structs.ts
adhusson: addressed in 81c60aa1479c44215f2c2ab3b1274d715950b1b3

preprocessing/structs.ts
lnist: I meant it as:

> /* * If `lock` is true, orders may not be added nor executed.

> /* * If `lock` is true, orders may not be added, nor executed, nor the offer list read by external contracts.

preprocessing/structs.ts
adhusson: done in 74a0f6eb91856a812fcd67c6e856ce0b64ba66ea

preprocessing/structs.ts
```diff
@@ -1,15 +1,86 @@
...
+   Offers are stored in a tree we call a "tick tree". Thanks to this tree structure, offer operations (insert, update, and retract) take constant time (the height of the tree is fixed).
+
+
+   ### Bins
+
+   <img src="./bin.png" width="80%"></img>
+
+   Below the bottom of the tree are _bins_. A bin is a doubly linked list of offers. All offers in a bin have the same tick. During a market order, offers in a bin are executed in order, from the first to the last. Inserted offers are always appended at the end of a bin. 
+
+   Bins are laid in sequence. In the context of an offer list, each bin has an associated tick (and a tick determines a price). If a bin has tick `t`, all the offers in the following bin have tick `t+tickSpacing`.
```
espendk, line 43: Bins are laid in sequence. In the context of an offer list, each bin has an associated tick (and a tick determines a price). If a bin has tick `t`, the following bin has tick `t+tickSpacing`.

src/MgvLib.sol
adhusson: I'm confused, what do you mean "from taker's pov" or "from maker's pov"? Here I only refer to what mutates this value in memory.

preprocessing/structs.ts
```suggestion
   Bins are laid in sequence. In the context of an offer list, each bin has an associated tick (and a tick determines a price). If a bin has tick `t`, the following bin has tick `t+tickSpacing`.
```

src/MgvLib.sol
lnist: I'm referring to old line 61 / new line 65 - it mentions "from taker's pov"

src/AUDIT_TRAIL_2.md
adhusson: added ref to update f9af61f1c93c45975590d4cfe47985a8a513dc38

preprocessing/structs.ts
espendk: Could even simpler: https://github.com/mangrovedao/mangrove-core/pull/588/files#r1342558338

preprocessing/structs.ts
adhusson: done in e95b438ecdf2406911cfbef7fe696e9e75e32505

PR #589, opened 2023-09-28
Gasprice in mwei

adhusson:
(also better arg names in `DensityLib.paramsTo96X32`)

To review carefully since there's a broken symmetry: to convert gwei to wei is `* 1e9`, eth to gwei is also `* 1e9`, but now that we have mwei, the conversions are:
* eth -> mwei: `* 1e12`
* mwei -> wei: `* 1e6`

### DIFF
--- a/lib/DensityLib.sol
+++ b/lib/DensityLib.sol
@@ -116,32 +116,33 @@ library DensityLib {
   }}
 
   // Convenience function: get a fixed-point density from the given parameters. Computes the price of gas in outbound base tokens, then multiplies by cover_factor.
-  // Warning: multiply input prices by 100
+  // Warning: you must multiply input usd prices by 100
   // not supposed to be gas optimized
   function paramsTo96X32(
     uint outbound_decimals, 
-    uint gasprice_in_gwei, 
-    uint eth_in_usdx100, 
-    uint outbound_display_in_usdx100, 
+    uint gasprice_in_mwei, 
+    uint eth_in_centiusd, 
+    uint outbound_display_in_centiusd, 
     uint cover_factor
   ) internal pure returns (uint) {
     // Do not use unchecked here
     require(uint8(outbound_decimals) == outbound_decimals,"DensityLib/fixedFromParams1/decimals/wrong");
-    uint num = cover_factor * gasprice_in_gwei * (10**outbound_decimals) * eth_in_usdx100;
+    uint num = cover_factor * gasprice_in_mwei * (10**outbound_decimals) * eth_in_centiusd;
     // use * instead of << to trigger overflow check
-    return (num * (1 << 32)) / (outbound_display_in_usdx100 * 1e9);
+    return (num * (1 << 32)) / (outbound_display_in_centiusd * 1e12);
   }
 
+  // Version with token in mwei instead of usd
   function paramsTo96X32(
     uint outbound_decimals, 
-    uint gasprice_in_gwei, 
-    uint outbound_display_in_gwei, 
+    uint gasprice_in_mwei, 
+    uint outbound_display_in_mwei, 
     uint cover_factor
   ) internal pure returns (uint) {
     // Do not use unchecked here
     require(uint8(outbound_decimals) == outbound_decimals,"DensityLib/fixedFromParams2/decimals/wrong");
-    uint num = cover_factor * gasprice_in_gwei * (10**outbound_decimals);
+    uint num = cover_factor * gasprice_in_mwei * (10**outbound_decimals);
     // use * instead of << to trigger overflow check
-    return (num * (1 << 32)) / outbound_display_in_gwei;
+    return (num * (1 << 32)) / outbound_display_in_mwei;
   }
 }
--- a/script/core/ActivateMarket.s.sol
+++ b/script/core/ActivateMarket.s.sol
@@ -12,8 +12,8 @@ import {ActivateSemibook} from "./ActivateSemibook.s.sol";
  TKN1=USDC \
  TKN2=WETH \
  TICK_SPACING=1 \
- TKN1_IN_GWEI=$(cast --to-wei $(bc -l <<< 1/$NATIVE_IN_USDC) gwei) \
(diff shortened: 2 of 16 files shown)


lib/DensityLib.sol
```diff
@@ -116,32 +116,33 @@ library DensityLib {
   }}
 
   // Convenience function: get a fixed-point density from the given parameters. Computes the price of gas in outbound base tokens, then multiplies by cover_factor.
```
espendk, line 118: I don't think `base` makes sense here?
```suggestion
  // Convenience function: get a fixed-point density from the given parameters. Computes the price of gas in outbound tokens, then multiplies by cover_factor.
```

lib/DensityLib.sol
```diff
@@ -116,32 +116,33 @@ library DensityLib {
...
     uint cover_factor
   ) internal pure returns (uint) {
     // Do not use unchecked here
     require(uint8(outbound_decimals) == outbound_decimals,"DensityLib/fixedFromParams1/decimals/wrong");
-    uint num = cover_factor * gasprice_in_gwei * (10**outbound_decimals) * eth_in_usdx100;
+    uint num = cover_factor * gasprice_in_mwei * (10**outbound_decimals) * eth_in_centiusd;
     // use * instead of << to trigger overflow check
-    return (num * (1 << 32)) / (outbound_display_in_usdx100 * 1e9);
+    return (num * (1 << 32)) / (outbound_display_in_centiusd * 1e12);
   }
```
espendk, line 133: looks correct to me (I ignored `cover_factor` for simplicity)

lib/DensityLib.sol
```diff
@@ -116,32 +116,33 @@ library DensityLib {
...
   ) internal pure returns (uint) {
     // Do not use unchecked here
     require(uint8(outbound_decimals) == outbound_decimals,"DensityLib/fixedFromParams2/decimals/wrong");
-    uint num = cover_factor * gasprice_in_gwei * (10**outbound_decimals);
+    uint num = cover_factor * gasprice_in_mwei * (10**outbound_decimals);
     // use * instead of << to trigger overflow check
-    return (num * (1 << 32)) / outbound_display_in_gwei;
+    return (num * (1 << 32)) / outbound_display_in_mwei;
   }
 }
```
espendk, line 148: Same here 👍

script/core/ActivateMarket.s.sol
```diff
@@ -39,29 +39,29 @@ contract ActivateMarket is Deployer {
...
     fee: fee in per 10_000
   */
 
   /* 
-    tknX_in_gwei should be obtained like this:
+    tknX_in_mwei should be obtained like this:
     1. Get the price of one tknX display unit in native token, in display units.
-       For instance, on ethereum, the price of 1 WETH is 1e9 gwei
-    2. Multiply by 1e9
+       For instance, on ethereum, the price of 1 WETH is 1e12 mwei
```
espendk, line 50: This explanation doesn't make sense to me:
Line 49 says: Get the price __in display units__.
Line 50 says: The price of WETH is 1e12 mwei. - that's not *display units* 🤔 Shouldn't that be "1"?

script/core/ActivateSemibook.s.sol
```diff
@@ -65,27 +65,27 @@ contract ActivateSemibook is Test2, Deployer {
...
-    price of gas in tokens (obtained from the price of gas in gwei and the price
-    of tokens in gwei) by COVER_FACTOR to get the density.
+    price of gas in tokens (obtained from the price of gas in mwei and the price
+    of tokens in mwei) by COVER_FACTOR to get the density.
     
     Units:
-       - outbound_in_gwei is in gwei/display token units
+       - outbound_in_mwei is in mwei/display token units
        - COVER_FACTOR is unitless
        - decN is in (base token units)/(display token units)
```
espendk, line 74: What are `base token units` here? Should it just say `token units`?

espendk (approved):
lgtm!

Added some smaller comments

lib/DensityLib.sol
adhusson: I think the meaning is "base as opposed to display units" (since one argument is in `outbound_display_...`). Maybe `in wei of outbound tokens` or `in base units of outbound tokens`?

lib/DensityLib.sol
adhusson: 🙏❤️ Thank you for checking!

script/core/ActivateSemibook.s.sol
adhusson: see my other comment abbove, I always struggled with how to explicitly reference "non-display units"

script/core/ActivateMarket.s.sol
adhusson: I would say "the price 1 WETH (in display units) (aka 1e18 WETH in wei) is 1e12 mwei", does that make sense?

lib/DensityLib.sol
espendk: Ah, I see. Not sure what the good solution is. IMO the ambiguity of using `wei` et al. for other things than native tokens is problematic and should be avoided. So `base units` is probably better.

script/core/ActivateMarket.s.sol
espendk: First, I would change the example to non-WETH since that makes the text somewhat ambiguous.

Second, I would only state the native token in display units in the example in Step 1. Step 2 deals with converting to mwei, so I'd continue the example there (instead of stuffing it all into Step 1).

lib/DensityLib.sol
adhusson: addressed here a7e62d0d3da31bcc78957f9b825c3361fd9240cd

script/core/ActivateMarket.s.sol
adhusson: Tried to fix in 480c9e9bbe02e02ccf6ed5611170f5760db23de6

PR #590, opened 2023-09-28
Remove inverted mangrove

adhusson:


espendk (approved):
Nice simplification and seems the right priority to me to remove a feature that hasn't yet been used - and we have plenty to do with the normal Mangrove.

PR #591, opened 2023-09-28
return constant if possible + don't compute shouldUpdateBranch unless necessary

adhusson:
Very small optim, more for code clarity than anything else, but still must be checked

### DIFF
--- a/src/MgvHasOffers.sol
+++ b/src/MgvHasOffers.sol
@@ -74,11 +74,6 @@ contract MgvHasOffers is MgvCommon {
         // with
         // offerId == local.best() (but best will maybe go away in the future)
 
-        // If shouldUpdateBranch is false is means we are about to insert anyway, so no need to load the best branch right now
-        // if local.bin < offerBin then a better branch is already cached. note that local.bin >= offerBin implies local.bin = offerTick
-        // no need to check for prevId/nextId == 0: if offer is last of leaf, it will be checked by leaf.isEmpty()
-        shouldUpdateBranch = shouldUpdateBranch && prevId == 0 && !bestBin.strictlyBetter(offerBin);
-
         if (prevId == 0) {
           // offer was tick's first. new first offer is offer.next (may be 0)
           leaf = leaf.setBinFirst(offerBin, nextId);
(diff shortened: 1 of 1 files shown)


src/MgvHasOffers.sol
```diff
@@ -97,8 +92,10 @@ contract MgvHasOffers is MgvCommon {
           nextOfferData.offer = nextOfferData.offer.prev(prevId);
         }
         if (prevId != 0 && nextId != 0) {
-          return (local, shouldUpdateBranch);
+          return (local, false);
         }
+
+        shouldUpdateBranch = shouldUpdateBranch && prevId == 0 && !bestBin.strictlyBetter(offerBin);
```
espendk, line 98: I don't understand why we should update the branch if `bestBin == offerBin`? Doesn't that mean that `local` already has everything cached correctly?

(I know this wasn't changed here, but it struck me when reading the code)

src/MgvHasOffers.sol
```diff
@@ -74,11 +74,6 @@ contract MgvHasOffers is MgvCommon {
         // with
         // offerId == local.best() (but best will maybe go away in the future)
 
-        // If shouldUpdateBranch is false is means we are about to insert anyway, so no need to load the best branch right now
-        // if local.bin < offerBin then a better branch is already cached. note that local.bin >= offerBin implies local.bin = offerTick
-        // no need to check for prevId/nextId == 0: if offer is last of leaf, it will be checked by leaf.isEmpty()
```
espendk, line 79: Should the comment be retained? I see it's a bit outdated, but still nice to have the explanation.

espendk (approved):
Looks good to me! Added a few questions

src/MgvHasOffers.sol
adhusson: in this case, we are deleting the first offer of a bin, so the best bin might have become empty. In that case we need to execute the following if branch at the end of the `dislodgeOffer` function:

```
      if (shouldUpdateBranch) {
        local = local.binPosInLeaf(leaf.bestNonEmptyBinPos());
      }
```

src/MgvHasOffers.sol
adhusson: It's being replaced by a comment in another branch :)

src/MgvHasOffers.sol
espendk: ah, so it handles the case where we're deleting the best offer?

adhusson: yes!

PR #594, opened 2023-09-29
make effects of getNextBest easier to think about

adhusson:
In the process of writing docs, I found that explaining `getNextBest` was easier if it did not leave `local` and `mor.leaf` in an inconsistent state.

At the cost of some gas, this PR makes `getNextBest` always fully update `local` and `mor.leaf` (even though it's only needed at the end of the market order).

Before this PR, the documentation has to explain that calling `getNextBest` is only safe if you know you won't need `local.binPosInLeaf` or a correct `mor.leaf` value and that you'll know how to update both values yourself later.

Now it's possible to just say "it updates `local` and `leaf` to reflect the new best offer".

It still does not update offer pointers, that would be a much higher cost, and that particular optimization is also easier to explain.

Gas changes (classic/200, everything other test shows 0 diff):

image

### DIFF
--- a/src/MgvOfferTaking.sol
+++ b/src/MgvOfferTaking.sol
@@ -89,9 +89,11 @@ abstract contract MgvOfferTaking is MgvHasOffers {
     Bin offerBin = offer.bin(tickSpacing);
     uint nextId = offer.next();
 
+    Leaf leaf = mor.leaf;
+    leaf = leaf.setBinFirst(offerBin, nextId);
+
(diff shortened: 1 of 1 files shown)


espendk:
> At the cost of some gas

There seems to be both savings and costs in the table, that's a bit surprising no?

espendk (approved):
lgtm

adhusson:
The reductions are surprising and even the increase amounts are surprising.

PR #595, opened 2023-09-29
Feat/consistent offer wants

adhusson:
As of now:
- We round down `offer.wants()` when there's a full fill (it's up to the maker to not set a tick+volume that yields a wants of 0)
- We round up `offer.wants()` when there's a partial fill (for fairness, so the taker cannot choose a partial fill value to repeatedly give 0 to the maker while taking a nonzero amount)
- We consider that there's a partial fill when the taker-specified volume exactly matches the offer volume

So there's a possible inconsistency between what `offer.wants()` returns and what a `clean` with the exact right amounts will ask of the taker. Same if the marketOrder has a `fillVolume` that exactly matches the offer.

There are 3 ways to resolve the inconsistency:
- round up `offer.wants()` even at full fill
- round down induced `offer.wants()` at partial fill
- consider exact match to be full fills (instead of partial fills)

For consistency / simplicity, this PR picks option 3: it considers that there's a full fill when the taker-specified volume exactly matches the offer volume. It is negligibly more gas but gains in consistency.

EDIT: now also rouding up in offer.wants(), but leaving the `<=` to make code easier to read

### DIFF
--- a/src/MgvOfferTaking.sol
+++ b/src/MgvOfferTaking.sol
@@ -482,7 +482,7 @@ abstract contract MgvOfferTaking is MgvHasOffers {
         uint offerGives = sor.offer.gives();
         uint offerWants = sor.offer.wants();
         /* <a id="MgvOfferTaking/computeVolume"></a> Volume requested depends on total gives (or wants) by taker. Let `volume = sor.takerWants` if `mor.fillWants` is true, and `volume = sor.takerGives` otherwise; note that `volume <= fillVolume` in all cases. Example with `fillWants=true`: if `offerGives < fillVolume` the first branch of the outer `if` sets `volume = offerGives` and we are done; otherwise the 1st branch of the inner if is taken and sets `volume = fillVolume` and we are done. */
-        if ((mor.fillWants && offerGives < fillVolume) || (!mor.fillWants && offerWants < fillVolume)) {
+        if ((mor.fillWants && offerGives <= fillVolume) || (!mor.fillWants && offerWants <= fillVolume)) {
           sor.takerWants = offerGives;
           sor.takerGives = offerWants;
         } else {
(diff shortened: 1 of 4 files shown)


src/MgvOfferTaking.sol
```diff
@@ -482,7 +482,7 @@ abstract contract MgvOfferTaking is MgvHasOffers {
         uint offerGives = sor.offer.gives();
         uint offerWants = sor.offer.wants();
         /* <a id="MgvOfferTaking/computeVolume"></a> Volume requested depends on total gives (or wants) by taker. Let `volume = sor.takerWants` if `mor.fillWants` is true, and `volume = sor.takerGives` otherwise; note that `volume <= fillVolume` in all cases. Example with `fillWants=true`: if `offerGives < fillVolume` the first branch of the outer `if` sets `volume = offerGives` and we are done; otherwise the 1st branch of the inner if is taken and sets `volume = fillVolume` and we are done. */
```
espendk, line 484:
```suggestion
        /* <a id="MgvOfferTaking/computeVolume"></a> Volume requested depends on total gives (or wants) by taker. Let `volume = sor.takerWants` if `mor.fillWants` is true, and `volume = sor.takerGives` otherwise; note that `volume <= fillVolume` in all cases. Example with `fillWants=true`: if `offerGives <= fillVolume` the first branch of the outer `if` sets `volume = offerGives` and we are done; otherwise the 1st branch of the inner if is taken and sets `volume = fillVolume` and we are done. */
```

espendk (approved):
I think it makes sense to consider an exact match a full fill - not doing so actually seems quite weird.

It's not clear to me why `offer.wants()` doesn't round up in all cases, but that seems somewhat orthogonal to whether an exact match is a full fill.

PR #597, opened 2023-09-30
Feat/align volumes and ticks

adhusson:
- tickrange is now (like uniswap -887272;887272. It's the largest tickrange such that the resulting prices fit in 128 bits.
- volumes (`fillVolume`, `takerWants`, `takerGives`, `offer.gives`) are on 127 bits
- tick math adjusted
- some solidity->assembly optimizations
- error messages adjusted, tests added

### DIFF
--- a/lib/Constants.sol
+++ b/lib/Constants.sol
@@ -46,7 +46,14 @@ uint constant MAX_RATIO_MANTISSA = 340256786836388094050805785052946541084;
 int constant MAX_RATIO_EXP = 0;
 uint constant MANTISSA_BITS = 128;
 uint constant MANTISSA_BITS_MINUS_ONE = 127;
-uint constant MAX_SAFE_VOLUME = 340282366920938463463374607431768211455;
+/* 
+With |tick|<=887272 and normalized mantissas on 128 bits, the maximum possible mantissa is 340282295208261841796968287475569060645, so the maximum safe volume before overflow is actually 340282438633630198193436196978374475856. 
+
+The immediate idea is to set MAX_SAFE_VOLUME to `(1<<max_safe_volume_bits)-1`, where `max_safe_volume_bits = 256-MANTISSA_BITS`, for simplicity. But we also have a `ByVolume` API, so we also need the ratios `inbound_amount/outbound_amount` to not correspond to a tick outside of the tick range. 
+
+Instead of fine-tuning MAX_SAFE_VOLUME to be exactly as big as it can be, we just set `max_safe_volume_bits = 256 - MANTISSA_BITS - 1`.
+*/
+uint constant MAX_SAFE_VOLUME = 170141183460469231731687303715884105727;
 // Without optimizer enabled it fails above 79. With optimizer and 200 runs it fails above 80. Set default a bit lower to be safe.
 uint constant INITIAL_MAX_RECURSION_DEPTH = 75;
 uint constant INITIAL_MAX_GASREQ_FOR_FAILING_OFFERS_MULTIPLIER = 3;
--- a/lib/Constants.sol
+++ b/lib/Constants.sol
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity ^0.8.17;
 
+/* For all the relevant constants here, the non-literal expression that computes them is checked in `Constants.t.sol`. */
+
 uint constant ONE = 1; // useful to name it for drawing attention sometimes
 uint constant ONES = type(uint).max;
 uint constant TOPBIT = 1 << 255;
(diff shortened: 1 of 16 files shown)


lib/Constants.sol
```diff
@@ -37,26 +37,30 @@ int constant NUM_BINS = 2097152;
...
+int constant MIN_TICK = -887272;
+int constant MAX_TICK = 887272;
+uint constant MIN_RATIO_MANTISSA = 170153974464283981435225617938057077692;
+int constant MIN_RATIO_EXP = 255;
+uint constant MAX_RATIO_MANTISSA = 340256786836388094050805785052946541084;
+int constant MAX_RATIO_EXP = 0;
+uint constant MANTISSA_BITS = 128;
+uint constant MANTISSA_BITS_MINUS_ONE = 127;
+/* 
+With |tick|<=887272 and normalized mantissas on 128 bits, the maximum possible mantissa is 340282295208261841796968287475569060645, so the maximum safe volume before overflow is actually 340282438633630198193436196978374475856.
```
espendk, line 54: How are these numbers calculated? A comment on this would make this easier to understand and check :-)

lib/Constants.sol
```diff
@@ -37,26 +37,30 @@ int constant NUM_BINS = 2097152;
...
+uint constant MIN_RATIO_MANTISSA = 170153974464283981435225617938057077692;
+int constant MIN_RATIO_EXP = 255;
+uint constant MAX_RATIO_MANTISSA = 340256786836388094050805785052946541084;
+int constant MAX_RATIO_EXP = 0;
+uint constant MANTISSA_BITS = 128;
+uint constant MANTISSA_BITS_MINUS_ONE = 127;
+/* 
+With |tick|<=887272 and normalized mantissas on 128 bits, the maximum possible mantissa is 340282295208261841796968287475569060645, so the maximum safe volume before overflow is actually 340282438633630198193436196978374475856. 
 
+The immediate idea is to set MAX_SAFE_VOLUME to `(1<<max_safe_volume_bits)-1`, where `max_safe_volume_bits = 256-MANTISSA_BITS`, for simplicity. But we also have a `ByVolume` API, so we also need the ratios `inbound_amount/outbound_amount` to not correspond to a tick outside of the tick range.
```
espendk: The implication in the last sentence is that `1/MAX_SAFE_VOLUME` and `MAX_SAFE_VOLUME/1` must correspond to a tick, right?

Which means
```
MAX_SAFE_VOLUME/1  <  1.0001^887272 = 340256786836388094050805785052946505084
```

which isn't satisfied for `MAX_SAFE_VOLUME = (1<<max_safe_volume_bits)-1 = 340282366920938463463374607431768211460`.

lib/Constants.sol
```diff
@@ -37,26 +37,30 @@ int constant NUM_BINS = 2097152;
...
-int constant MAX_TICK = 1048575;
-uint constant MIN_RATIO_MANTISSA = 4735129379934731672174804159539094721182826496;
-int constant MIN_RATIO_EXP = 303;
-uint constant MAX_RATIO_MANTISSA = 3441571814221581909035848501253497354125574144;
-int constant MAX_RATIO_EXP = 0;
-uint constant MANTISSA_BITS = 152;
-uint constant MANTISSA_BITS_MINUS_ONE = 151;
-// Maximum volume that can be multiplied by a ratio mantissa
-uint constant MAX_SAFE_VOLUME = 20282409603651670423947251286015;
+Instead of fine-tuning MAX_SAFE_VOLUME to be exactly as big as it can be, we just set `max_safe_volume_bits = 256 - MANTISSA_BITS - 1`.
```
espendk: Do we gain anything by making `MAX_SAFE_VOLUME` fit within 127 bits?

lib/TickLib.sol
```diff
@@ -285,24 +287,32 @@ library TickLib {
...
+    int diff = log2ratio+int(exp)-int(MANTISSA_BITS_MINUS_ONE);
+    if (diff > 0) {
+      // For |tick| <= 887272, this drops at most 5 bits of precision
+      man = man >> uint(diff);
+    } else {
+      man = man << uint(-diff);
+    }
+    // For |tick| << 887272, log2ratio <= 127
+    exp = uint(int(MANTISSA_BITS_MINUS_ONE)-log2ratio);
+    }
```
espendk: indentation is weird here

test/core/MakerOperations.t.sol
```diff
@@ -1066,4 +1066,16 @@ contract MakerOperationsTest is MangroveTest, IMaker {
...
+  function test_new_offer_extremas() public {
+    mgv.setDensity96X32(olKey, 0);
+    vm.expectRevert("mgv/writeOffer/gives/tooBig");
+    uint ofr = mkr.newOfferByTick(Tick.wrap(0), MAX_SAFE_VOLUME + 1, 100_000, 0);
+    vm.expectRevert("mgv/writeOffer/gives/tooBig");
+    mkr.updateOfferByTick(Tick.wrap(0), MAX_SAFE_VOLUME + 1, 100_000, 0, ofr);
+    vm.expectRevert("mgv/writeOffer/tick/outOfRange");
+    mkr.updateOfferByTick(Tick.wrap(MAX_TICK + 1), 1, 100_000, 0, ofr);
+    vm.expectRevert("mgv/writeOffer/tick/outOfRange");
+    mkr.newOfferByTick(Tick.wrap(MAX_TICK + 1), 1, 100_000, 0);
```
espendk, line 1088: Should this also check that all of these succeed if `+ 1` is removed?

test/core/TakerOperations.t.sol
```diff
@@ -658,15 +658,65 @@ contract TakerOperationsTest is MangroveTest {
...
+    mgv.marketOrderByTick(olKey, Tick.wrap(MAX_TICK + 1), 100, true);
+    vm.expectRevert("mgv/mOrder/tick/outOfRange");
+    mgv.marketOrderByTick(olKey, Tick.wrap(MIN_TICK - 1), 100, true);
+  }
+
+  function test_marketOrderForByTick_extrema_ko(address taker) public {
+    vm.expectRevert("mgv/mOrder/tick/outOfRange");
+    mgv.marketOrderForByTick(olKey, Tick.wrap(MAX_TICK + 1), 100, true, taker);
+    vm.expectRevert("mgv/mOrder/tick/outOfRange");
+    mgv.marketOrderForByTick(olKey, Tick.wrap(MIN_TICK - 1), 100, true, taker);
```
espendk: Nit: Why are these separate functions, when the ByVolume tests are combined?

test/core/TickAndBin.t.sol
```diff
@@ -48,33 +48,40 @@ contract TickAndBinTest is MangroveTest {
...
     });
 
     inner_test_ratioFromTick({
       tick: Tick.wrap(1),
-      expected_sig: 2854780834950460954092783596092880171791548416,
-      expected_exp: 151
+      expected_sig: 170158197578815278654860472446255694138,
+      expected_exp: 127
     });
   }
```
espendk: Notes on how to interpret & check these numbers would be great documentation.

espendk (approved):
I haven't checked the math in any detail, that will require more time (and pointers) than I have right now. Let me know if I should.

lib/Constants.sol
adhusson: Will add comments on this (also see Constant.t.sol)

lib/Constants.sol
adhusson: I'll rephrase, the implication is just that `1.0001^MAX_TICK` should be more than `MAX_SAFE_VOLUME`

lib/Constants.sol
adhusson: Do you mean as opposed to 128 bits ? We gain the fact that the bounds on offer volume, fillVolume, and takerWants/takerGives are all the same. If it was on 128 bits, we'd need to document the fact that the `ByVolume` functions can sometimes revert.

lib/TickLib.sol
adhusson: yup will fix

lib/Constants.sol
espendk: The comment makes is sound like we're lazy and can't be bothered to fine-tune ;-)
But if I understand correctly, that's not the case. Instead, I think you're saying "the actual max is somewhere between 2^127 and 2^128 - 1, but choosing 2^127 - 1 means we can simply restrict volumes to 127 bits and be done with it".

test/core/MakerOperations.t.sol
adhusson: yes, will add them

test/core/TickAndBin.t.sol
adhusson: will add comments

test/core/TakerOperations.t.sol
adhusson: yes, will make similar

lib/Constants.sol
adhusson: yes that's exactly it

PR #599, opened 2023-10-02
fix various typos

lnist:


src/MgvLib.sol
```diff
@@ -421,7 +421,7 @@ interface HasMgvEvents {
 interface IMaker {
   /* Called upon offer execution. 
   - If the call throws, Mangrove will not try to transfer funds and the first 32 bytes of revert reason are passed to `makerPosthook` as `makerData`
-  - If the call returns normally, returndata is passed to `makerPosthook` as `makerData` and Mangrove will attempt to transfer the funds.
+  - If the call returns normally, returnData is passed to `makerPosthook` as `makerData` and Mangrove will attempt to transfer the funds.
```
espendk: Maybe there should be backticks around `returnData` here? The casing seems to indicate it's a code concept.

espendk (approved):
lgtm

src/MgvLib.sol
lnist: added

adhusson (approved):

PR #600, opened 2023-10-02
minor fixees to event doc

lnist:


src/MgvLib.sol
```diff
@@ -96,14 +96,14 @@ interface HasMgvEvents {
   /* 
     Events in solidity is a hard thing to do in a optimal way. If you look at it as a purely gas efficient issue, you want to emit as few events as possible and with as few fields as possible. But as events also has to be usable for an off chain user, then doing this is not always the best solution.
 
-    We tried to list the main points that we would like to do with events.
+    We tried to list the main forces that are inflicted upon events.
```
espendk, line 99: "inflict" sounds somewhat ominous 😅

Maybe this is better?

```suggestion
    We tried to list the main forces that drive which events and data to emit:
```

src/MgvLib.sol
```diff
@@ -96,14 +96,14 @@ interface HasMgvEvents {
...
 
-    1. Use as little gas as possible
+    1. Use as little gas as possible.
     2. An indexer should be able to keep track of the state of Mangrove.
     3. Doing RPC calls directly, should be able to find offers and other information based on offerList, maker, taker, offer id, etc.
 
-    These 3 points all have there own direction and it is therefore not possible to find a solution that is optimal for all 3 points.
-    We have therefore tried to find a solution that is a good balance between the all 3 points.
+    These forces all pull in their own direction and it is therefore not possible to find a solution that is optimal for all three.
+    The following events are therefore an attempt to balance the trade-offs.
```
espendk, line 106:
```suggestion
    These forces are in conflict and it is impossible to find a solution that is optimal for all three.
    The following events are therefore an attempt to balance the trade-offs.
```

adhusson (approved):

src/MgvLib.sol
lnist: I added your suggestions to feat/doc pr since you got front-run

src/MgvLib.sol
lnist: I added your suggestions to feat/doc pr since you got front-run

src/MgvLib.sol
espendk: Thanks and darn

PR #603, opened 2023-10-02
feat(doc): Add natspec for IMangrove

lnist:
It is quite verbose.
It is repetitious.
It is better than nothing, but can definitely be improved.

Closes https://github.com/mangrovedao/mangrove-core/issues/476


src/IMangrove.sol
```diff
@@ -1,27 +1,52 @@
 // SPDX-License-Identifier: Unlicense
 // This file was manually adapted from a file generated by abi-to-sol. It must
-// be kept up-to-date with the actual Mangrove interface. Fully automatic
-// generation is not yet possible due to user-generated types in the external
-// interface lost in the abi generation.
+// be kept up-to-date with the actual Mangrove interface.
```
espendk: > // This file was manually adapted from a file generated by abi-to-sol.

This seems irrelevant, can't we remove it?

src/IMangrove.sol
```diff
@@ -1,27 +1,52 @@
...
 
+  ///@notice See {IERC20Permit-DOMAIN_SEPARATOR}.
+  ///@return the domain separator.
   function DOMAIN_SEPARATOR() external view returns (bytes32);
 
+  ///@notice See {IERC20Permit-PERMIT_TYPEHASH}.
+  ///@return The permit type hash.
   function PERMIT_TYPEHASH() external pure returns (bytes32);
 
+  ///@notice approves the spender to spend the amount of tokens on behalf of the caller.
```
espendk, line 20: Should we put backticks on references to the argument names?

"amount" = "value" - can we rename of them so they match?

src/IMangrove.sol
```diff
@@ -1,27 +1,52 @@
...
+  ///@param outbound_tkn The address of the outbound token.
+  ///@param inbound_tkn The address of the inbound token.
+  ///@param spender The address of the spender.
+  ///@param value The amount of tokens to approve.
+  ///@return true if the approval succeeded; always true.
```
espendk, line 25: > always true.

Does the function revert if approval fails? Or can't it fail at all?

src/IMangrove.sol
```diff
@@ -1,27 +1,52 @@
...
   function approve(address outbound_tkn, address inbound_tkn, address spender, uint value) external returns (bool);
 
+  ///@notice returns the allowance of the spender to spend tokens on behalf of the owner.
+  ///@param outbound_tkn The address of the (maker) outbound token.
+  ///@param inbound_tkn The address of the (maker) inbound token.
+  ///@param owner The address of the owner.
+  ///@param spender The address of the spender.
+  ///@return allowance The amount of tokens the spender is allowed to spend on behalf of the owner.
   function allowances(address outbound_tkn, address inbound_tkn, address owner, address spender)
```
espendk: Isn't it weird that the function is in plural? (not natspec-related, just noticed it now)

src/IMangrove.sol
```diff
@@ -1,27 +1,52 @@
...
+  ///@param outbound_tkn The address of the outbound token.
+  ///@param inbound_tkn The address of the inbound token.
+  ///@param spender The address of the spender.
+  ///@param value The amount of tokens to approve.
+  ///@return true if the approval succeeded; always true.
   function approve(address outbound_tkn, address inbound_tkn, address spender, uint value) external returns (bool);
 
+  ///@notice returns the allowance of the spender to spend tokens on behalf of the owner.
+  ///@param outbound_tkn The address of the (maker) outbound token.
+  ///@param inbound_tkn The address of the (maker) inbound token.
```
espendk, line 30: Why "(maker)" here? Why not on `approve`?

It looks weird to me.

src/IMangrove.sol
```diff
@@ -34,26 +59,40 @@ interface IMangrove is HasMgvEvents {
...
   function nonces(address owner) external view returns (uint nonce);
 
   // # Taker functions
 
-  function marketOrderByVolume(OLKey memory olKey, uint takerWants, uint takerGives, bool fillWants)
-    external
-    returns (uint takerGot, uint takerGave, uint bounty, uint fee);
-
+  ///@notice Performs a market order on a specified offer list taking offers up to a limit price.
+  ///@param olKey The offer list key given by (maker) `outbound`, (maker) `inbound`, and `tickSpacing`.
```
espendk: Above we say "outbound _token_", maybe we should do the same here?

src/IMangrove.sol
```diff
@@ -34,26 +59,40 @@ interface IMangrove is HasMgvEvents {
...
+  ///@param maxTick The limit price the taker is ready to pay (the log base 1.0001 of the price).
```
espendk: I think we should use "ratio" here. I know the other docs ended up using "price" but I really think it's unfortunate and inconsistent - we chose "ratio" to avoid people getting confused about what "price" means for an offer list in the context of a market.

(@adhusson and @jkrivine can battle this one out ;-) )

src/IMangrove.sol
```diff
@@ -34,26 +59,40 @@ interface IMangrove is HasMgvEvents {
...
+  ///@param fillVolume if `fillWants` is true, the amount of `olKey.outbound` the taker wants to buy; otherwise, the amount of `olKey.inbound` the taker wants to sell.
+  ///@param fillWants if true, the matching engine tries to get the taker all they want; otherwise, the matching engine tries to sell all that the taker gives (subject to price).
```
espendk: Both are subject to price. Maybe move that sentence to after the last period?

src/IMangrove.sol
```diff
@@ -34,26 +59,40 @@ interface IMangrove is HasMgvEvents {
...
+  ///@return takerGot The amount of `olKey.outbound` the taker got.
+  ///@return takerGave The amount of `olKey.inbound` the taker gave.
+  ///@return bounty The amount of native token the taker got as a bounty due to failing offers (in wei)
+  ///@return fee The amount of native token the taker paid as a fee (in wei of `olKey.outbound`)
```
espendk: No native tokens are involved. Maybe this instead:

```suggestion
  ///@return fee The amount of `olKey.outbound` token the taker paid as a fee to Mangrove.
```

src/IMangrove.sol
```diff
@@ -34,26 +59,40 @@ interface IMangrove is HasMgvEvents {
...
+  ///@dev The market order stops when the price exceeds (an approximation of) 1.0001^`maxTick`, or when the end of the book has been reached, or:
```
espendk: I think the approximation part is confusing and I'm not sure it adds anything?

```suggestion
  ///@dev The market order stops when there are no more offers at or below `maxTick`, when the end of the book has been reached, or:
```

src/IMangrove.sol
```diff
@@ -34,26 +59,40 @@ interface IMangrove is HasMgvEvents {
...
+  ///@dev * If `fillWants` is true, the market order stops when `fillVolume` units of `olKey.outbound` have been obtained. To buy a specific volume of `olKey.outbound` at any price, set `fillWants` to true, set `fillVolume` to volume you want to buy, and set `maxTick` to the `MAX_TICK` constant.
```
espendk:
```suggestion
  ///@dev * If `fillWants` is true, the market order stops when `fillVolume` units of `olKey.outbound` have been obtained. To buy a specific volume of `olKey.outbound` at any price, set `fillWants` to true, set `fillVolume` to the volume you want to buy, and set `maxTick` to the `MAX_TICK` constant.
```

src/IMangrove.sol
```diff
@@ -34,26 +59,40 @@ interface IMangrove is HasMgvEvents {
...
+  ///@dev * If `fillWants` is false, the market order stops when `fillVolume` units of `olKey.inbound` have been sold. To sell a specific volume of `olKey.inbound` at any price, set `fillWants` to false, set `fillVolume` to the volume you want to sell, and set `maxTick` to the `MAX_TICK` constant.
```
espendk: Maybe "paid" instead of "sold"; I think that better matches "obtained" above.

```suggestion
  ///@dev * If `fillWants` is false, the market order stops when `fillVolume` units of `olKey.inbound` have been paid. To sell a specific volume of `olKey.inbound` at any price, set `fillWants` to false, set `fillVolume` to the volume you want to sell, and set `maxTick` to the `MAX_TICK` constant.
```

src/IMangrove.sol
```diff
@@ -34,26 +59,40 @@ interface IMangrove is HasMgvEvents {
...
+  ///@param maxTick The limit price the taker is ready to pay (the log base 1.0001 of the price).
+  ///@param fillVolume if `fillWants` is true, the amount of `olKey.outbound` the taker wants to buy; otherwise, the amount of `olKey.inbound` the taker wants to sell.
+  ///@param fillWants if true, the matching engine tries to get the taker all they want; otherwise, the matching engine tries to sell all that the taker gives (subject to price).
+  ///@param maxGasreqForFailingOffers The maximum allowed gas required for failing offers (in wei).
+  ///@return takerGot The amount of `olKey.outbound` the taker got.
+  ///@return takerGave The amount of `olKey.inbound` the taker gave.
+  ///@return bounty The amount of native token the taker got as a bounty due to failing offers (in wei)
+  ///@return fee The amount of native token the taker paid as a fee (in wei of `olKey.outbound`)
+  ///@dev Mangrove stops a market order after it has gone through failing offers such that their cumulative `gasreq` is greater than the global `maxGasreqForFailingOffers` parameter. This function can be used for the taker to override that parameter.
```
espendk: same comments as above :-)

src/IMangrove.sol
```diff
@@ -62,118 +101,334 @@ interface IMangrove is HasMgvEvents {
     uint maxGasreqForFailingOffers
   ) external returns (uint takerGot, uint takerGave, uint bounty, uint fee);
 
+  ///@notice Performs a market order on a specified offer list taking offers up to a limit price.
```
espendk: Maybe specify that the limit price is defined by a ratio of inbound/outbound volumes?

Eg
```suggestion
  ///@notice Performs a market order on a specified offer list taking offers up to a limit price defined by a ratio `inbound/outound` of volumes.
```

src/IMangrove.sol
```diff
@@ -62,118 +101,334 @@ interface IMangrove is HasMgvEvents {
...
+  ///@param olKey The offer list key given by (maker) `outbound`, (maker) `inbound`, and `tickSpacing`.
+  ///@param takerWants The amount the taker wants. This is used along with `takerGives` to derive a max price (`maxTick`) note that some precision is lost to rounding.
```
espendk: Are there rounding errors happening here? I was under the impression that we would always derive a `maxTick` such that `log_1.0001(takerGives/takerWants) <= maxTick`.

```suggestion
  ///@param takerWants The amount the taker wants. This is used along with `takerGives` to derive a max price (`maxTick`) which is the lowest allowed tick in the offer list such that `log_1.0001(takerGives/takerWants) <= maxTick`.
```

src/IMangrove.sol
```diff
@@ -62,118 +101,334 @@ interface IMangrove is HasMgvEvents {
...
+  function marketOrderByVolume(OLKey memory olKey, uint takerWants, uint takerGives, bool fillWants)
+    external
+    returns (uint takerGot, uint takerGave, uint bounty, uint fee);
+
+  ///@notice Performs a market order on a specified offer list taking offers up to a limit price for a specified taker.
+  ///@param olKey The offer list key given by (maker) `outbound`, (maker) `inbound`, and `tickSpacing`.
+  ///@param maxTick The limit price the taker is ready to pay (the log base 1.0001 of the price).
+  ///@param fillVolume if `fillWants` is true, the amount of `olKey.outbound` the taker wants to buy; otherwise, the amount of `olKey.inbound` the taker wants to sell.
+  ///@param fillWants if true, the matching engine tries to get the taker all they want; otherwise, the matching engine tries to sell all that the taker gives (subject to price).
+  ///@param taker The taker from which amounts will be transferred from and to the. If the `msg.sender`'s allowance for the given `olKey.outbound`,`olKey.inbound` are strictly less than the total amount eventually spent by `taker`, the call will fail.
```
espendk:
```suggestion
  ///@param taker The taker from which amounts will be transferred from and to. If the `msg.sender`'s allowance for the given `olKey.outbound`,`olKey.inbound` is strictly less than the total amount eventually spent by `taker`, the call will fail.
```

src/IMangrove.sol
```diff
@@ -62,118 +101,334 @@ interface IMangrove is HasMgvEvents {
...
   function fund(address maker) external payable;
 
+  ///@notice Withdraws the caller's (the maker's) native token (funds for provisioning offers) by transferring them to the caller.
+  ///@param amount the amount to withdraw.
+  ///@return noRevert whether the transfer succeeded.
   function withdraw(uint amount) external returns (bool noRevert);
 
+  ///@notice Gets the maker's balance of native tokens (funds for provisioning offers).
+  ///@param maker The maker to get the balance for.
+  ///@return balance The maker's balance of native tokens (funds for provisioning offers).
```
espendk: Maybe we should clarify that the returned value is the "free" balance, i.e., not including any provision tied up by live offers.

src/IMangrove.sol
```diff
@@ -62,118 +101,334 @@ interface IMangrove is HasMgvEvents {
...
   function balanceOf(address maker) external view returns (uint balance);
 
-  function newOfferByVolume(OLKey memory olKey, uint wants, uint gives, uint gasreq, uint gasprice)
+  ///@notice Creates a new offer on Mangrove, where the caller is the maker. The maker can implement the `IMaker` interface to be called during offer execution.
+  ///@param olKey The offer list key given by (maker) `outbound`, (maker) `inbound`, and `tickSpacing`.
+  ///@param tick The tick (which is a power of 1.0001 and induces a price). Note that if `tickSpacing > 1`, the actual tick of the offer may differ from the `tick` argument by up to `tickSpacing-1`.
```
espendk: > which is a power of 1.0001

?

espendk: > the actual tick of the offer may differ from the `tick` argument by up to `tickSpacing-1`.

Maybe this instead:

> the actual tick of the offer will be the smallest tick `offerTick > tick` that satisfies `offerTick % tickSpacing == 0`.

src/IMangrove.sol
```diff
@@ -62,118 +101,334 @@ interface IMangrove is HasMgvEvents {
...
+  ///@param gives The amount of `olKey.outbound` the maker gives. Must be less than MAX_SAFE_VOLUME.
+  ///@param gasreq The amount of gas required to execute the offer logic in the maker's `IMaker` implementation. This will limit the gas available, and the offer will fail if it spends more.
+  ///@param gasprice The maximum gas price the maker is willing to pay a penalty for due to failing execution.
+  ///@return offerId the id of the offer on Mangrove. Can be used to retract or update the offer (even to reuse a taken offer).
+  ///@dev The gasreq and gasprice is used to derive the provision which will be used to pay a bounty to taker if the offer fails.
```
espendk:
```suggestion
  ///@dev The gasreq and gasprice are used to derive the provision which will be used to pay a penalty if the offer fails.
```

src/IMangrove.sol
```diff
@@ -62,118 +101,334 @@ interface IMangrove is HasMgvEvents {
...
+  ///@dev This function is payable to enable delivery of the provision along with the offer creation.
+  function newOfferByTick(OLKey memory olKey, Tick tick, uint gives, uint gasreq, uint gasprice)
     external
     payable
     returns (uint offerId);
 
-  function newOfferByTick(OLKey memory olKey, Tick tick, uint gives, uint gasreq, uint gasprice)
+  ///@notice Creates a new offer on Mangrove, where the caller is the maker. The maker can implement the `IMaker` interface to be called during offer execution.
+  ///@param olKey The offer list key given by (maker) `outbound`, (maker) `inbound`, and `tickSpacing`.
+  ///@param wants The amount of `olKey.inbound` the maker wants. This is used along with `gives` to derive a tick (price). Note that some precision is lost to rounding.
```
espendk: Same note re. loss of precision as further up.

src/IMangrove.sol
```diff
@@ -62,118 +101,334 @@ interface IMangrove is HasMgvEvents {
...
+  ///@param gasprice The maximum gas price the maker is willing to pay a penalty for due to failing execution.
+  ///@return offerId the id of the offer on Mangrove. Can be used to retract or update the offer (even to reuse a taken offer).
+  ///@dev This function is just a wrapper for `newOfferByTick`, see that function for details.
+  function newOfferByVolume(OLKey memory olKey, uint wants, uint gives, uint gasreq, uint gasprice)
     external
     payable
     returns (uint offerId);
 
-  function updateOfferByVolume(OLKey memory olKey, uint wants, uint gives, uint gasreq, uint gasprice, uint offerId)
+  ///@notice Updates an existing, owned offer on Mangrove, where the caller is the maker.
```
espendk:
```suggestion
  ///@notice Updates an existing offer on Mangrove, where the caller is the maker.
```

src/IMangrove.sol
```diff
@@ -62,118 +101,334 @@ interface IMangrove is HasMgvEvents {
...
+  ///@param gives The amount of `olKey.outbound` the maker gives. This is used along with `wants` to derive a tick (price). Note that some precision is lost to rounding. Must be less than MAX_SAFE_VOLUME.
+  ///@param gasreq The amount of gas required to execute the offer logic in the maker's `IMaker` implementation.
+  ///@param gasprice The maximum gas price the maker is willing to pay a penalty for due to failing execution.
+  ///@param offerId The id of the offer on Mangrove.
+  ///@dev This function is just a wrapper for `updateOfferByTick`, see that function for details.
+  function updateOfferByVolume(OLKey memory olKey, uint wants, uint gives, uint gasreq, uint gasprice, uint offerId)
     external
     payable;
 
+  ///@notice Retracts an owned offer from Mangrove, where the caller is the maker.
```
espendk:
```suggestion
  ///@notice Retracts an offer from Mangrove, where the caller is the maker.
```

src/IMangrove.sol
```diff
@@ -62,118 +101,334 @@ interface IMangrove is HasMgvEvents {
...
+  ///@param olKey The offer list key given by (maker) `outbound`, (maker) `inbound`, and `tickSpacing`.
+  ///@param offerId The id of the offer on Mangrove.
+  ///@param deprovision Whether to deprovision the offer (i.e. return the provision to the maker's balance on Mangrove).
```
espendk:
```suggestion
  ///@param deprovision Whether to deprovision the offer, i.e, return the provision to the maker's balance on Mangrove.
```

src/IMangrove.sol
```diff
@@ -62,118 +101,334 @@ interface IMangrove is HasMgvEvents {
...
+  ///@return true if locked; otherwise, false.
+  ///@dev The lock protects modifying or inspecting the offer list while an order is in progress.
   function locked(OLKey memory olKey) external view returns (bool);
 
+  ///@notice The `offerId` of the best offer on the offer list.
+  ///@param olKey The offer list key given by (maker) `outbound`, (maker) `inbound`, and `tickSpacing`.
+  ///@return offerId The `offerId` of the best offer on the offer list.
   function best(OLKey memory olKey) external view returns (uint offerId);
 
+  ///@notice the offer list key given by the hash (if the offer list key has been activated at least once).
```
espendk:
```suggestion
  ///@notice Gets the offer list key with the given hash (if the offer list key has been activated at least once).
```

src/IMangrove.sol
```diff
@@ -62,118 +101,334 @@ interface IMangrove is HasMgvEvents {
...
+  ///@return _local The local configuration for the offer list.
   function config(OLKey memory olKey) external view returns (Global _global, Local _local);
 
+  ///@notice Determines whether the reentrancy lock is in effect for the offer list.
+  ///@param olKey The offer list key given by (maker) `outbound`, (maker) `inbound`, and `tickSpacing`.
+  ///@return true if locked; otherwise, false.
+  ///@dev The lock protects modifying or inspecting the offer list while an order is in progress.
   function locked(OLKey memory olKey) external view returns (bool);
 
+  ///@notice The `offerId` of the best offer on the offer list.
```
espendk:
```suggestion
  ///@notice Gets the `offerId` of the best offer in the offer list.
```

src/IMangrove.sol
```diff
@@ -62,118 +101,334 @@ interface IMangrove is HasMgvEvents {
...
 
   // # Offer view functions
 
+  ///@notice Gets an offer in packed format.
+  ///@param olKey The offer list key given by (maker) `outbound`, (maker) `inbound`, and `tickSpacing`.
+  ///@param offerId The `offerId` of the offer on the offer list.
+  ///@return offer The offer in packed format.
   function offers(OLKey memory olKey, uint offerId) external view returns (Offer offer);
 
+  ///@notice Gets an offer detail in packed format.
```
espendk:
```suggestion
  ///@notice Gets an offer's details in packed format.
```

src/IMangrove.sol
```diff
@@ -62,118 +101,334 @@ interface IMangrove is HasMgvEvents {
...
+  ///@param olKey The offer list key given by (maker) `outbound`, (maker) `inbound`, and `tickSpacing`.
+  ///@param offerId The `offerId` of the offer on the offer list.
+  ///@return offerDetail The offer detail in packed format.
```
espendk:
```suggestion
  ///@return offerDetail The offer details in packed format.
```

src/IMangrove.sol
```diff
@@ -62,118 +101,334 @@ interface IMangrove is HasMgvEvents {
...
   function offerDetails(OLKey memory olKey, uint offerId) external view returns (OfferDetail offerDetail);
 
+  ///@notice Gets both an offer and an offer detail in packed format.
```
espendk:
```suggestion
  ///@notice Gets both an offer and its details in packed format.
```

src/IMangrove.sol
```diff
@@ -62,118 +101,334 @@ interface IMangrove is HasMgvEvents {
...
+  ///@param olKey The offer list key given by (maker) `outbound`, (maker) `inbound`, and `tickSpacing`.
+  ///@param offerId The `offerId` of the offer on the offer list.
+  ///@return offer The offer in packed format.
+  ///@return offerDetail The offer detail in packed format.
```
espendk:
```suggestion
  ///@return offerDetail The offer details in packed format.
```

src/IMangrove.sol
```diff
@@ -62,118 +101,334 @@ interface IMangrove is HasMgvEvents {
...
+  ///@notice Gets the governance address.
+  ///@return the governance address.
   function governance() external view returns (address);
 
+  ///@notice Activates an offer list.
+  ///@param olKey The offer list key given by (maker) `outbound`, (maker) `inbound`, and `tickSpacing`.
+  ///@param fee in basis points, of `olKey.outbound` given to the taker. This fee is sent to Mangrove. Fee is capped to ~2.5%.
+  ///@param density96X32 The density of the offer list used to define a minimum offer volume. See `setDensity96X32`.
+  ///@param offer_gasbase The gasbase of the offer list used to define a minimum provision necessary for offers. See `setGasbase`.
+  ///@dev The flipped offer list is expected to have the same `tickSpacing`.
```
espendk: Not sure why this is here? To say that it's expected that the flipped offer list is also activated? (won't always be true)

src/IMangrove.sol
```diff
@@ -62,118 +101,334 @@ interface IMangrove is HasMgvEvents {
...
   function activate(OLKey memory olKey, uint fee, uint density96X32, uint offer_gasbase) external;
 
+  ///@notice Deactivates an offer list.
+  ///@param olKey The offer list key given by (maker) `outbound`, (maker) `inbound`, and `tickSpacing`.
   function deactivate(OLKey memory olKey) external;
 
+  ///@notice Kills the Mangrove instance. A dead instance cannot have offers executed or funds received.
```
espendk:
```suggestion
  ///@notice Kills the Mangrove instance. A dead instance cannot have offers executed or funds received, but offers can be retracted and funds can be withdrawn.
```

src/IMangrove.sol
```diff
@@ -62,118 +101,334 @@ interface IMangrove is HasMgvEvents {
...
+  ///@param offerId The `offerId` of the offer on the offer list.
+  ///@return offer The offer in packed format.
+  ///@return offerDetail The offer detail in packed format.
   function offerData(OLKey memory olKey, uint offerId) external view returns (Offer offer, OfferDetail offerDetail);
 
   // # Governance functions
 
+  ///@notice Gets the governance address.
+  ///@return the governance address.
   function governance() external view returns (address);
```
espendk, line 311: Should we annotate the functions that can only be called by governance?

src/IMangrove.sol
```diff
@@ -62,118 +101,334 @@ interface IMangrove is HasMgvEvents {
...
+  ///@param olKey The offer list key given by (maker) `outbound`, (maker) `inbound`, and `tickSpacing`.
+  ///@param fee in basis points, of `olKey.outbound` given to the taker. This fee is sent to Mangrove. Fee is capped to ~2.5%.
   function setFee(OLKey memory olKey, uint fee) external;
 
+  ///@notice Sets the gasbase for the offer list.
+  ///@param olKey The offer list key given by (maker) `outbound`, (maker) `inbound`, and `tickSpacing`.
+  ///@param offer_gasbase The gasbase of the offer list used to define a minimum provision necessary for offers. Represents the gas overhead used by processing the offer inside Mangrove + the overhead of initiating an entire order. Stored in thousands in a maximum of 9 bits.
   function setGasbase(OLKey memory olKey, uint offer_gasbase) external;
 
+  ///@notice Sets the gasmax for Mangrove, the maximum amount of gas required to execute an offer.
```
espendk: Maybe this is more precise?

```suggestion
  ///@notice Sets the gasmax for Mangrove, the maximum amount of gas an offer can require to execute.
```

src/IMangrove.sol
```diff
@@ -62,118 +101,334 @@ interface IMangrove is HasMgvEvents {
...
+  ///@notice Sets a new governance address.
+  ///@param governanceAddress The new governance address.
   function setGovernance(address governanceAddress) external;
 
+  ///@notice Sets the monitor/oracle. The `monitor/oracle` can provide real-time values for `gasprice` and `density` to Mangrove. It can also receive liquidity event notifications.
+  ///@param monitor The new monitor/oracle address.
   function setMonitor(address monitor) external;
 
+  ///@notice Sets whether Mangrove notifies the Monitor/Oracle when and offer is taken
+  ///@param notify Whether Mangrove notifies the Monitor/Oracle when and offer is taken
```
espendk:
```suggestion
  ///@notice Sets whether Mangrove notifies the Monitor when and offer is taken
  ///@param notify Whether Mangrove notifies the Monitor when and offer is taken
```

src/IMangrove.sol
```diff
@@ -62,118 +101,334 @@ interface IMangrove is HasMgvEvents {
...
   function setNotify(bool notify) external;
 
+  ///@notice Sets whether Mangrove uses the monitor/oracle for `gasprice` and `density` values.
+  ///@param useOracle Whether Mangrove uses the monitor/oracle for `gasprice` and `density` values.
```
espendk: Maybe this is easier to read?

```suggestion
  ///@notice Sets whether Mangrove uses the monitor as oracle for `gasprice` and `density` values.
  ///@param useOracle Whether Mangrove uses the monitor as oracle for `gasprice` and `density` values.
```

espendk (commented):
Great!

Added a bunch of comments/suggestions.

src/IMangrove.sol
lnist: The checker enforces all functions to have natspec. So easier to just add "dummy" natspec than improving the tool now :D

src/IMangrove.sol
lnist: agree

src/IMangrove.sol
lnist: Lack of cross referencing and terminology definitions make it a bit of a mess. Because is it known here that outbound is the maker's perspective?
I have now added (maker) to the other.

src/IMangrove.sol
lnist: backticks: In general I am not a big fan unless its important. Reading prose is easier on the eyes.

amount=value: I am unsure about this permit api since it is borrowed from somewhere, maybe we should keep it as is?

The intend of this exercise was not to revisit APIs :D if we do, then I'm not the one to answer. @adhusson ?

src/IMangrove.sol
lnist: @adhusson ?

src/IMangrove.sol
lnist: Changed to outbound_tkn so leaving as is.

src/IMangrove.sol
lnist: lol, wut. Agree :)

src/IMangrove.sol
lnist:
```
 function approve(address outbound_tkn, address inbound_tkn, address spender, uint value) external returns (bool) {
    unchecked {
      _allowances[outbound_tkn][inbound_tkn][msg.sender][spender] = value;
      emit Approval(outbound_tkn, inbound_tkn, msg.sender, spender, value);
      return true;
    }
  }

```

src/IMangrove.sol
lnist: I cannot come up with a good suggestion right now. Can you spell it out? :)

src/IMangrove.sol
lnist: done

src/IMangrove.sol
espendk: I'll try

src/IMangrove.sol
lnist: done. Also fixed source.

src/IMangrove.sol
```diff
@@ -34,26 +58,40 @@ interface IMangrove is HasMgvEvents {
...
 
-  function marketOrderByVolume(OLKey memory olKey, uint takerWants, uint takerGives, bool fillWants)
-    external
-    returns (uint takerGot, uint takerGave, uint bounty, uint fee);
-
+  ///@notice Performs a market order on a specified offer list taking offers up to a limit price.
+  ///@param olKey The offer list key given by (maker) `outbound_tkn`, (maker) `inbound_tkn`, and `tickSpacing`.
+  ///@param maxTick Must be `>= MIN_TICK` and `<= MAX_TICK`. The limit price the taker is ready to pay (the log base 1.0001 of the price).
+  ///@param fillVolume Must be `<= MAX_SAFE_VOLUME`. If `fillWants` is true, the amount of `olKey.outbound_tkn` the taker wants to buy; otherwise, the amount of `olKey.inbound_tkn` the taker wants to sell.
+  ///@param fillWants if true, the matching engine tries to get the taker all they want; otherwise, the matching engine tries to sell all that the taker gives (subject to price).
```
espendk, line 72:
```suggestion
  ///@param fillWants if true, the matching engine tries to get the taker all they want; otherwise, the matching engine tries to sell all that the taker gives. In both cases subject to the price limit.
```

src/IMangrove.sol
espendk:
```suggestion
  ///@param fillWants if true, the matching engine tries to get the taker all they want; otherwise, the matching engine tries to sell all that the taker gives. In both cases subject to the price limit.
```

src/IMangrove.sol
lnist: resolved manually.

src/IMangrove.sol
lnist: c.f. MgvOfferMaking. But yes the formulation is a bit weird.

lnist: c.f. MgvOfferMaking.

src/IMangrove.sol
lnist: thx :)

src/IMangrove.sol
espendk: Not just weird, wrong no?

@adhusson

src/IMangrove.sol
adhusson: value is the standard name for this `approve` argument I would say let's stick with it https://eips.ethereum.org/EIPS/eip-20#approve

src/IMangrove.sol
espendk: @adhusson wdyt?

src/IMangrove.sol
adhusson: yeah let's update it!

src/IMangrove.sol
adhusson: wait @jkrivine is team ratio or team price? anyway that makes sense in this context I agree, maybe something like "The log of the limit price..." (so that it starts off straightforward and plain) and then "where price is the ratio of blah blah blah"?

src/IMangrove.sol
lnist: Changed to
>  ///@dev If the flipped offer list is active then the offer lists are expected to have the same `tickSpacing`.

src/IMangrove.sol
espendk: @jkrivine is the champion of Team Ratio! ♘

src/IMangrove.sol
adhusson: it is supposed to be used as a power of 1.0001 😄

adhusson: agreed will change

espendk: lol, that's not how I read it at all 😂

espendk: Let me propose somthing. If GitHub will allow me to make suggestions

espendk (approved):
lgtm!

(especially the reverts of @lnist's changes re. `feePaid` :-P)

adhusson:
I merged but I did not realized I was reverting something 🤣

lnist:
> I merged but I did not realized I was reverting something 🤣

You just consolidated it the opposite way :)

PR #612, opened 2023-10-05
Configuration scripts for new deployments

jkrivine:
This PR does the following:
* adds ERC20 addresses for arbitrum context 
* adds a generic script that infers and logs density and gasbase for a specific market. 
The script required some minor refactoring of existing scripts (only in use for testnet)

### DIFF
--- a/script/deployers/EvalMarketParameters.s.sol
+++ b/script/deployers/EvalMarketParameters.s.sol
@@ -25,10 +25,8 @@ import {console} from "forge-std/console.sol";
  */
 
 contract EvalMarketParameters is Deployer, ActivateUtils {
-  uint nativePrice;
-
   function run() public {
-    nativePrice = vm.envUint("NATIVE_IN_USD");
+    uint nativePrice = vm.envUint("NATIVE_IN_USD");
     uint gaspriceOverride = vm.envUint("GASPRICE_OVERRIDE");
     IERC20 token0 = IERC20(envAddressOrName("TOKEN0"));
     IERC20 token1 = IERC20(envAddressOrName("TOKEN1"));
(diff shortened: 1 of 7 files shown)


script/deployers/EvalMarketParameters.s.sol
```diff
@@ -0,0 +1,56 @@
...
+  function run() public {
+    nativePrice = vm.envUint("NATIVE_IN_USD");
+    uint gaspriceOverride = vm.envUint("GASPRICE_OVERRIDE");
+    IERC20 token0 = IERC20(envAddressOrName("TOKEN0"));
+    IERC20 token1 = IERC20(envAddressOrName("TOKEN1"));
+    uint price0 = vm.envUint("TOKEN0_IN_USD");
+    uint price1 = vm.envUint("TOKEN1_IN_USD");
+    uint coverFactor = vm.envUint("COVER_FACTOR");
+
+    innerRun(token0, token1, price0, price1, gaspriceOverride, coverFactor);
```
espendk: nit: The named parameters pattern as in the other scripts would be nice & concise here.

script/deployers/EvalMarketParameters.s.sol
```diff
@@ -0,0 +1,56 @@
...
+ * GASPRICE_OVERRIDE is the gasprice (in gwei) to consider to compute density
+ * COVER_FACTOR to compute density, i.e density =
+ *
+ *
+ * usage with Native token being AETH (with n=1):
+ * GASPRICE_OVERRIDE=1 NATIVE_IN_USD=1600 TOKEN0=USDC TOKEN1=USDT TOKEN0_IN_USD=1 TOKEN1_IN_USD=1 COVER_FACTOR=100 forge script EvalMarketParameters --fork-url arbitrum
+ */
+
+contract EvalMarketParameters is Deployer, ActivateUtils {
+  uint nativePrice;
```
espendk: This won't be set if `innerRun` is called directly from another script.

Would IMO be cleaner to make it a parameter of `toGweiOfNative` (which would then be pure and easy to reason about).

espendk (changes requested):
lgtm.

There's a minor issue in the new EvalMarketScript that I think should be addressed.

PR #613, opened 2023-10-05
use remappings internally & rename `mgv_*` to `@mgv/*`

### LINKED ISSUE #611: Used mgv_* prefix for imported forge libs
Right now we only use `mgv_lib` on our own libraries but we should use it when importing `forge-std` so that there is no clash when another project imports `mangrove` and uses a different forge-std version

For instance if project A imports `mangrove`, and project A uses another version of `forge-std`, then the remapping that wins will be `forge-std -> path/to/projectA/lib/forge-std/src`, so mangrove will use projectA's forge-std

adhusson:
Closes #611

also changes remappings from `mgv_X -> X/` to `@mgv/X -> X/` (more clear this is a remapping thanks to the `@`, and `/` makes more sense than `_`).

note that when `mangrove` is imported, the remapping will become `@mgv/X -> lib/mangrove-core/X/`

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,11 +13,13 @@ Just `forge install mangrovedao/mangrove-core`.
 ⚠️ You will not get the usual remapping `mangrove-core/=lib/mangrove-core/src/` (because forge's remapping generation heuristic sees the `preprocessing/lib/` directory and decides to remap to the parent dir). Instead, you will get:
 
 ```
-mgv_src/=lib/mangrove-core/src/
-mgv_lib/=lib/mangrove-core/lib/
-mgv_test/=lib/mangrove-core/test/
-mgv_script/=lib/mangrove-core/script/
+@mgv/src/=lib/mangrove-core/src/
+@mgv/lib/=lib/mangrove-core/lib/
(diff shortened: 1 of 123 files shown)


README.md
```diff
@@ -13,12 +13,14 @@ Just `forge install mangrovedao/mangrove-core`.
...
-mgv_lib/=lib/mangrove-core/lib/
-mgv_test/=lib/mangrove-core/test/
-mgv_script/=lib/mangrove-core/script/
+@mgv/src/=lib/mangrove-core/src/
+@mgv/lib/=lib/mangrove-core/lib/
+@mgv/test/=lib/mangrove-core/test/
+@mgv/script/=lib/mangrove-core/script/
 ```
 
+Use this likely-unique prefix even internally so projects that depend on mangrove don't mess with mangrove's internal dependencies
```
espendk:
```suggestion
Use this likely-unique prefix even internally so projects that depend on Mangrove don't mess with Mangrove's internal dependencies.
```

espendk (dismissed):
lgtm

espendk (approved):

PR #614, opened 2023-10-06
feat(gastest): Add printDescription(postfix) and expose gas-measurement

lnist:


gas-measurement.sh
```diff
@@ -6,7 +6,7 @@ default_json_file="${out_dir}/gas-measurement.json"
 function run_gas_measurement() {
   local json_file=$1
   local tmp_file="$1.tmp"
-  forge test --mp "test/core/gas/*" -vv --silent --json > "${json_file}"
+  forge test --mp "$GAS_MATCH_PATH" -vv --silent --json > "${json_file}"
```
espendk, line 9: Would be nice if `GAS_MATCH_PATH` was a parameter instead and/or had a default value.

What happens if I run `gas-measurement.sh` without this env var set? That's what I normally do and what the instructions in the README.md say :-)

package.json
```diff
@@ -36,7 +37,8 @@
     "/lib",
     "/LICENSE",
     "/README.md",
-    "/doc"
+    "/doc",
+    "./gas-measurement.sh"
```
espendk: Why the `.` here, when it's not there for the other files?

test/lib/gas/GasTestBase.t.sol
```diff
@@ -40,6 +40,11 @@ abstract contract GasTestBaseStored {
     console.log("Description: %s", description);
   }
 
+  function printDescription(string memory postfix) public virtual {
+    description = string.concat(description, postfix);
```
espendk: I'm surprised that this changes the description. If this is called multiple times, then the description grows each time?

Perhaps it would be more predictable, if the postfix was only printed this once? Eg by having this function do the actual printing and `printDescription()` forward to this with an empty postfix?

espendk (commented):
Nice. Added some smaller proposals.

gas-measurement.sh
lnist: Changed readme. Yes this could be improved, but I don't want to spend time on it right now. The argument passing is extremely rudimentary right now.

espendk: ok 😔

test/lib/gas/GasTestBase.t.sol
lnist: agree :)

espendk (approved):

PR #617, opened 2023-10-10
feat(gasreq): Adds virtual to makerPosthook and enables gasreq measurement

lnist:
In order to measure gasreq, we add virtual to makerPosthook, so that a special Mangrove can be crafted which captures the gas used for a maker.

This is then used in the strat library to measure gasreq for Kandel and MangroveOrder.

Additionally, a more expensive gasbase case was found - we missed an oracle, and leaving behind offers is more expensive.

gas-measurement is made resilient to test-changes when using diff.


espendk (approved):
lgtm

PR #648, opened 2023-10-31
docs: Update contact map and sequence chart to MGV v2

espendk:
Minimal update of the two first diagrams in the code doc:

- migrated diagrams from OmniGraffle to draw.io, so it's easier to update in the future (source is embedded in the PNG files).
- tried to match the old style and layout
- updated the Mangrove logo
- updated the methods mentioned in the contact map to MGV v2.

Resolution should be better, but we should have a graphic person update this, so deeming this good enough for now.


lnist (commented):
You should be able to up the resolution by exporting with zoom=200% or similar. Other than that, looks good.

lnist (approved):

PR #651, opened 2023-11-07
feat: Use addresses from mangrove-deployments and context-addresses packages

espendk:
This PR is a first improvement to the deployment process: It moves context and deployment addresses out of this repo to these new repos/npm packages:

- Deployments: https://github.com/mangrovedao/mangrove-deployments/
- Context addresses: https://github.com/mangrovedao/context-addresses/ 

This enables better management of deployments, incl. matching deployments to contract versions. Furthermore, it removes the need to publish new versions of this package due to deployments; This in turn allow us to use the package SemVer for contract versioning.

Similar considerations apply to extracting the context addresses.

Later PR's will address other remaining aspects re. the deployment process such as:

- Separate deployment and governance scripts
- Update README with the new process.

## Brief solution overview

The existing file-based address infrastructure is retained: Addresses are still read from files in the `addresses` folder using the same code and directory&file structures as before.

The change is that, instead of having the addresses files in git, they are generated in a new build step by copying addresses from
- https://github.com/mangrovedao/mangrove-deployments/ and writes them to `addresses/deployed`
- https://github.com/mangrovedao/context-addresses/ and writes them to `addresses/context`

This is handled by two new scripts:
- `copyDeploymentAddresses.js`
- `copyContextAddresses.js`.

These scripts have a few configuration parameters in `config.js`. See that file for details and documentation.

Also, the CI file guard that guarded against unintended changes to addresses has been removed, since addresses are no longer in this repo.

### Versioning and deployments
When querying `mangrove-deployments` for deployment addresses, one must specify the contract version and whether the deployment is of a "released" version.

By default, `copyDeploymentAddresses.js` queries for contract versions matching patches of the current package version, regardless of whether it's released. Both defaults can be changed in `config.js`; In that case, the user is responsible for specifying a version that is compatible with the current version of the governance scripts.


copyDeploymentAddresses.js
```diff
@@ -0,0 +1,83 @@
...
+    const networkAddresses = getOrCreateNetworkAddresses(networkId);
+    networkAddresses.push({
+      name:
+        contractDeployments.deploymentName ?? contractDeployments.contractName,
+      address: networkDeployments.primaryAddress,
+    });
+  }
+}
+
+// Replace the addresses files with the loaded deployment addresses
```
dontrolle: The comment says "Replace [...]". This is actually "Create [...]", right? It's a minor thing, but maybe worth fixing, since it supports the understanding that there shouldn't be anything in `/addresses/deployed/` before this script runs?

copyContextAddresses.js
```diff
@@ -0,0 +1,87 @@
...
+        networkAddresses.push({
+          name: erc20.symbol,
+          address: networkInstance.address,
+        });
+      }
+    }
+  }
+}
+
+// Replace the addresses files with the loaded context addresses
```
dontrolle: (Same as for `copyDeployementAddresses.js`) 


The comment says "Replace [...]". This is actually "Create [...]", right? It's a minor thing, but maybe worth fixing, since it supports the understanding that there shouldn't be anything in `/addresses/context/` before this script runs?

dontrolle (changes requested):

copyContextAddresses.js
espendk: good catch. Originally, I read the existing files and merged with the imported addresses, so "replace" was correct. That is no longer the case, so I'll fix it.

espendk: Fixed in 70917d5bde336873c24c5418c9510b92ee540ebb

copyDeploymentAddresses.js
espendk: Fixed in 70917d5bde336873c24c5418c9510b92ee540ebb

dontrolle (approved):
🌴

PR #682, opened 2023-12-14
docs(license): Update licenses

espendk:
- Change license of Contants.sol and TickLib.sol from BUSL-1.1 to MIT
- Replace Unlicense and BSD-2-Clause with MIT
- Replace AGPL-3.0 with AGPL-3.0-or-later
- Consistently use AGPL-3.0-or-later for tools, scripts, and tests
- Restore MIT license for OpenZeppelin code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Next version
 
+- Update licenses:
+  - Change license of Contants.sol and TickLib.sol from BUSL-1.1 to MIT
+  - Replace Unlicense and BSD-2-Clause with MIT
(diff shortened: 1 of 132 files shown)


dontrolle (changes requested):
We need something in the CHANGELOG for this.

espendk:
> We need something in the CHANGELOG for this.

Indeed. I just forgot to push that commit - it's here: 8388696432a9c6fc04dd8b1ea5795a6706a8087c

dontrolle (approved):
I've checked for missing instances; and since the CHANGELOG has been updated I'm approving.

PR #684, opened 2023-12-15
docs(license): Replace AGPL with MIT

espendk:


dontrolle (approved):
Checked and looks good.

PR #729, opened 2024-02-16
feat: support blast yield

Louis-Amas:
---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1206633588405271


maxencerb (dismissed):

script/core/deployers/BlastMangroveDeployer.s.sol
```diff
@@ -0,0 +1,60 @@
...
+
+import {BlastMangrove} from "@mgv/src/chains/blast/core/BlastMangrove.sol";
+import {MgvReader} from "@mgv/src/periphery/MgvReader.sol";
+import {MgvOracle} from "@mgv/src/periphery/MgvOracle.sol";
+import {IBlastMangrove} from "@mgv/src/chains/blast/IBlastMangrove.sol";
+import {Deployer} from "@mgv/script/lib/Deployer.sol";
+import {MgvReaderDeployer} from "@mgv/script/periphery/deployers/MgvReaderDeployer.s.sol";
+import {BaseBlast} from "@mgv/src/chains/blast/utils/BaseBlast.sol";
+
+contract MangroveDeployer is Deployer, BaseBlast {
```
espendk: Name should be updated :-)

src/chains/blast/utils/BaseBlast.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.10;
+
+import {IBlast} from "../interfaces/IBlast.sol";
+
+/// @title BaseBlast
+/// @author Mangrove
+/// @notice A contract that sets the BLAST address
+contract BaseBlast {
```
espendk: Since this just holds a constant I think a library would be more appropriate.

src/chains/blast/core/BlastMgvView.sol
```diff
@@ -0,0 +1,14 @@
...
+import {MgvView} from "../../../core/MgvView.sol";
+import {IBlastPoints} from "../interfaces/IBlastPoints.sol";
+
+/// @title BlastMgvView
+/// @notice A contract that inherits MgvView
+/// @dev This contract is used to get the points admin for blast
+contract BlastMgvView is MgvView, IBlastPoints {
+  function blastPointsAdmin() external view override returns (address) {
+    return _governance;
+  }
```
espendk: Could we put this function in `BlastMangrove` or will this make the contract too big?

If we could, I think we could avoid this file and `BlastMgvAppendix` + we wouldn't need to modify `Mangrove.sol`.

src/chains/blast/utils/BlastGasAndYieldClaimable.sol
espendk, line 1: Since this is only used in `BlastMangrove` maybe we could simply call `BLAST.configure` there?
That would be simpler and I don't think that one call warrants its own file :-)

espendk (commented):
Mostly fine, but I think we should consider simplifying it - I added some suggestions on how we can do that.

Also, tests should be added. You can eg be inspired by `MangroveDeployer.t.sol`.

script/core/deployers/BlastMangroveDeployer.s.sol
```diff
@@ -0,0 +1,57 @@
...
+import {BlastMangrove} from "@mgv/src/chains/blast/core/BlastMangrove.sol";
+import {MgvReader} from "@mgv/src/periphery/MgvReader.sol";
+import {MgvOracle} from "@mgv/src/periphery/MgvOracle.sol";
+import {IBlastMangrove} from "@mgv/src/chains/blast/IBlastMangrove.sol";
+import {Deployer} from "@mgv/script/lib/Deployer.sol";
+import {MgvReaderDeployer} from "@mgv/script/periphery/deployers/MgvReaderDeployer.s.sol";
+import {BlastLib} from "@mgv/src/chains/blast/lib/BlastLib.sol";
+import {MangroveDeployer} from "./MangroveDeployer.s.sol";
+
+contract BlastMangroveDeployer is MangroveDeployer {
```
espendk, line 13: This pattern is new and I find it a bit weird, since there's no actual inheritance. The only purpose is to be able to reuse the tests. And since the deployment scripts are totally independent, I’d say the tests should be as well.

Alternatively, you could introduce proper inheritance such that `BlastMangroveDeployer` only implements the few required changes to the `MangroveDeployer` script: Deploying `BlastMangrove` instead of `Mangrove` and setting the Blast governor.

I think the second option would be easier to maintain, as it maximizes reuse + makes it clear that those are the only two differences for that deployment.

src/toy/blast/Blast.sol
```diff
@@ -0,0 +1,265 @@
...
+}
+
+interface IYield {
+  function configure(address contractAddress, uint8 flags) external returns (uint);
+  function claim(address contractAddress, address recipientOfYield, uint desiredAmount) external returns (uint);
+  function getClaimableAmount(address contractAddress) external view returns (uint);
+  function getConfiguration(address contractAddress) external view returns (uint8);
+}
+
+contract Blast {
```
espendk: Maybe this should this implement the `IBlast` interface to ensure they're in sync?

espendk (commented):
Great improvements :-) Added some smaller suggestions.

espendk (approved):
lgtm

